PR #1, opened 2019-04-07
Feature/csi-12: add basic controller

olgashtivelman:
Added some basic infrastructure ,
-  grpc server for CSI controller and basic skeleton.
- storage array action interface + with just xiv implementation of getvolume as starter.
- storage array connection basics with limiting that number of open connections per storage type\array.
- basic folder layout 

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1)


ranhrl:
___
*[controller/array_action/array_connection_manager.py, line 19 at r4](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1#-LchBYWTAW3QAGINlPxD:-LchBYWTAW3QAGINlPxE:b3d30qk) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/365f07d3347645fa31c2ffbb7e925b4f0de736eb/controller/array_action/array_connection_manager.py#L19)):*
> ```Python
> class ArrayConnectionManager(object):
> 
>     def __init__(self, array_type, user, password, endpoint):  # TODO return the params back. 
> ```

should be endpoints (list) to support A9K multiple IP Addresses

ranhrl:
___
*[controller/array_action/array_connection_manager.py, line 51 at r4](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1#-LchCb_b6ofACRrAWzD4:-LchCb_b6ofACRrAWzD5:b-pnhtzk) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/365f07d3347645fa31c2ffbb7e925b4f0de736eb/controller/array_action/array_connection_manager.py#L51)):*
> ```Python
> 
>             logger.debug("got connection lock. array connection dict is: {}".format(array_connections_dict))
>             self.med_class = med_class(self.user, self.password, self.endpoint)
> ```

You should verify that you did not reach the max number of connections before you instantiate a new connection and not after.

ranhrl:
___
*[controller/array_action/array_mediator_interface.py, line 44 at r4](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1#-LchDAiYA6l8aKNbBQHF:-LchDAiYA6l8aKNbBQHG:b7tyk75) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/365f07d3347645fa31c2ffbb7e925b4f0de736eb/controller/array_action/array_mediator_interface.py#L44)):*
> ```Python
>             PoolDoesNotMatchCapabilities
>             CapabilityNotSupported (user is passing capability value that does not exist, or wrong spelling..) 
> ```

More Errors: OutOfResources (could be capacity or other resources (no more volumes in the domain etc); InvalidArgument (volume name is not valid); PermissionDenied;

ranhrl:
___
*[controller/array_action/array_mediator_interface.py, line 114 at r4](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1#-LchEMaR2pq0O66W0Lfv:-LchEMaR2pq0O66W0Lfw:bub9o4v) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/365f07d3347645fa31c2ffbb7e925b4f0de736eb/controller/array_action/array_mediator_interface.py#L114)):*
> ```Python
>             noAvailableLun
>             volumeNotFound
>             hostNotFound
> ```

PermissionDenied

ranhrl:
___
*[controller/array_action/array_mediator_xiv.py, line 24 at r4](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1#-LchFP-I2AUZ02Eo6ncc:-LchFP-I2AUZ02Eo6ncd:bxy4v77) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/365f07d3347645fa31c2ffbb7e925b4f0de736eb/controller/array_action/array_mediator_xiv.py#L24)):*
> ```Python
>     MAX_CONNECTION_RETRY = 3
> 
>     def __init__(self, user, password, endpoint):
> ```

you could use the same concept as in XCLIClient, means use a "connect" function to instantiate the class instead of __init__, just a suggestion.

ranhrl:
___
*[controller/array_action/array_mediator_xiv.py, line 47 at r4](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1#-LchFDiMBzeA8frMHz8k:-LchFDiMBzeA8frMHz8l:b-bb0x4h) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/365f07d3347645fa31c2ffbb7e925b4f0de736eb/controller/array_action/array_mediator_xiv.py#L47)):*
> ```Python
>             raise CredentailsError(self.endpoint)
>      
>     def close(self):
> ```

if you use connect, then this should be disconnect

ranhrl:
___
*[controller/array_action/array_mediator_xiv.py, line 51 at r4](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1#-LchG4CO76UVZ7Pt6-Cl:-LchG4CO76UVZ7Pt6-Cm:bat1f0e) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/365f07d3347645fa31c2ffbb7e925b4f0de736eb/controller/array_action/array_mediator_xiv.py#L51)):*
> ```Python
>             self.client.close()
>         
>     def _convert_volume_size(self, size):
> ```

Is this convert_to_bytes? (then name it properly).

ranhrl:
___
*[controller/array_action/array_mediator_xiv.py, line 58 at r4](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1#-LchGSVJErUEm5DtAnw7:-LchGSVJErUEm5DtAnw8:b-7xq626) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/365f07d3347645fa31c2ffbb7e925b4f0de736eb/controller/array_action/array_mediator_xiv.py#L58)):*
> ```Python
>         cli_volume = self.client.cmd.vol_list(vol=vol_name).as_single_element
>         logger.debug(cli_volume)
>         array_vol = Volume(self._convert_volume_size(int(cli_volume.size)))
> ```

cli_volume.size is not accurate enough, if you need to convert the size to bytes you should use cli_volume.size_MiB * 1024 * 1024

ranhrl:
___
*[controller/controller_server/csi_controller_server.py, line 30 at r4](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1#-LchHikz9-We2vlNlb2j:-LchHikz9-We2vlNlb2k:b-6l3be4) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/365f07d3347645fa31c2ffbb7e925b4f0de736eb/controller/controller_server/csi_controller_server.py#L30)):*
> ```Python
>         try:
>             with ArrayConnectionManager(xiv_type, user, password, array) as array_mediator:
>                 logger.debug(array_mediator)
> ```

Since all debug messages get to the log (and potentially to the users), any log statement that stays has to have a meaning.
I still suggest to put more messages under info (have info level very detailed) but keep things that are not for customers in debug, and set the default level to info eventually.

ranhrl:
___
*[controller/controller_server/csi_controller_server.py, line 93 at r4](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1#-LchIkUr5Zgr8smBsr58:-LchIkUr5Zgr8smBsr59:bouk2ef) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/365f07d3347645fa31c2ffbb7e925b4f0de736eb/controller/controller_server/csi_controller_server.py#L93)):*
> ```Python
>         # controller_server.add_insecure_port('[::]:{}'.format(self.server_port))
>         # controller_server.add_insecure_port('unix://{}'.format(self.server_port))
>         controller_server.add_insecure_port(self.endpoint)
> ```

is it insecured by definition? (spec)

ranhrl:
___
*[controller/controller_server/csi_controller_server.py, line 104 at r4](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1#-LchItykEGPD8t9ydBak:-LchItykEGPD8t9ydBal:bxy0ql3) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/365f07d3347645fa31c2ffbb7e925b4f0de736eb/controller/controller_server/csi_controller_server.py#L104)):*
> ```Python
>         except KeyboardInterrupt:
>             controller_server.stop(0)
>             print('Controller Server Stopped ...')
> ```

Print?

ranhrl:
___
*[controller/controller_server/csi_controller_server.py, line 107 at r4](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1#-LchJ06VCRk3tzj-G3Wq:-LchJ06VCRk3tzj-G3Wr:b-18oitq) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/365f07d3347645fa31c2ffbb7e925b4f0de736eb/controller/controller_server/csi_controller_server.py#L107)):*
> ```Python
> 
> 
> if __name__ == '__main__':
> ```

main should be in a function as well...

ranhrl:
___
*[controller/tests/array_action/array_mediator_xiv_tests.py, line 32 at r4](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1#-LchJhBH3FM5rhd0Azzn:-LchJhBH3FM5rhd0Azzo:b-stiila) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/365f07d3347645fa31c2ffbb7e925b4f0de736eb/controller/tests/array_action/array_mediator_xiv_tests.py#L32)):*
> ```Python
>         res = self.mediator.get_volume("some name")            
> 
>         self.assertTrue(res.size == self.mediator._convert_volume_size(vol.size))
> ```

The test should actually verify the res.size against a number you expect to get for vol.size == 10)

ranhrl:
___
*[controller/array_action/array_mediator_interface.py, line 44 at r4](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1#-LchDAiYA6l8aKNbBQHF:-LcjtlyFCT8ftE7InAjR:b-ki9llo) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/365f07d3347645fa31c2ffbb7e925b4f0de736eb/controller/array_action/array_mediator_interface.py#L44)):*

why needed?
if there issues will happened, then specific error of the storage system will raised, and it will raised up to the controler which will raise it as error.

I think we should keep formal errors in the interface only for the one that going to be used in the csi controller flow.   Please check architecture wiki as well.   And provide your feedback please.
</blockquote></details>

There is at list InvalidArgument in the spec, you only verify it for empty string, but I think it's just an example and you can return it for every volume name not supported.

ranhrl:
___
*[controller/array_action/array_mediator_svc.py, line 1 at r4](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1#-LchEnrzC3SbVVnOsHm4:-LchEnrzC3SbVVnOsHm5:byzx7xc) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/365f07d3347645fa31c2ffbb7e925b4f0de736eb/controller/array_action/array_mediator_svc.py#L1)):*
> ```Python
> class SVCArrayMediator():
> ```

Should Inherit ArrayMediator?

ranhrl:
___
*[controller/controller_server/csi_controller_server.py, line 93 at r4](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1#-LchIkUr5Zgr8smBsr58:-LcjuxCUBnQ6FRHibJDd:b9ll1bb) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/365f07d3347645fa31c2ffbb7e925b4f0de736eb/controller/controller_server/csi_controller_server.py#L93)):*

since its grpc unix socket and its inside the pod, then i don;t see any security needed here.
The spec doesn't state anything about it.
</blockquote></details>

What others are doing? I would consult with the community anyway.

ranhrl:
___
*[controller/array_action/array_mediator_interface.py, line 44 at r4](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1#-LchDAiYA6l8aKNbBQHF:-Lckd5Ud2UZlSwGQCpAl:b-pl0lr3) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/365f07d3347645fa31c2ffbb7e925b4f0de736eb/controller/array_action/array_mediator_interface.py#L44)):*
<details><summary><i>Previously, olgashtivelman wrote…</i></summary><blockquote>

I tried to create a vol name with invalid characters and got an "illiegal name" (pyxcli.errors.IllegalNameForObjectError) error on the storage system, so do you mean we should add this one? 
because invalid argument can be a nything and i am not sure when this returns... 
</blockquote></details>

I believe that it should be translated to the closest Error the API expects in that case, in addition to logging out the error text of course.

ranhrl:
___
*[controller/array_action/array_mediator_interface.py, line 114 at r4](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1#-LchEMaR2pq0O66W0Lfv:-LckdKF79_L_64mns_VM:babwjfr) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/365f07d3347645fa31c2ffbb7e925b4f0de736eb/controller/array_action/array_mediator_interface.py#L114)):*
<details><summary><i>Previously, olgashtivelman wrote…</i></summary><blockquote>

permissions for what?
for the user/passwrod from the sercret? this then should b added to all the APIs since they all cna have a wrong username\password and i think we mentioned this in the architecutre and i can add this here, is that what you meant? 
</blockquote></details>

Yes, you can be given a user that can login but not create volume in the requested pool for example (read-only user for instance).

ranhrl:
___
*[controller/array_action/array_mediator_xiv.py, line 51 at r4](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1#-LchG4CO76UVZ7Pt6-Cl:-Lckddp51zK6QohrDz1E:b2nhysk) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/365f07d3347645fa31c2ffbb7e925b4f0de736eb/controller/array_action/array_mediator_xiv.py#L51)):*
<details><summary><i>Previously, olgashtivelman wrote…</i></summary><blockquote>

i am not sure what exactly we need to convert it to so this was just my testing and thats why i left this as it is - and did not name it "bytes" or aything else . i think when we will actually use the get function and know exactly was is passing and what is needed that will be a good place to change this accordingly. 
</blockquote></details>

ack

ranhrl:
___
*[controller/array_action/array_mediator_xiv.py, line 58 at r4](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1#-LchGSVJErUEm5DtAnw7:-LckdhdT4iefkUt6CTbo:b3977yl) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/365f07d3347645fa31c2ffbb7e925b4f0de736eb/controller/array_action/array_mediator_xiv.py#L58)):*
<details><summary><i>Previously, olgashtivelman wrote…</i></summary><blockquote>

why? isn't the size always in GiB? (its what it looks like the response)
</blockquote></details>

it is rounded to the closest GB (and not GiB), this is one of XIV/A9K confusions. 
Actually, size_MiB is also not the right value to use in this case, but capacity is the right accurate one and it is in blocks:

From SC:

logical_capacity=self.convert_size_blocks_to_bytes(int(cli_volume.capacity))

ranhrl:
___
*[controller/controller_server/csi_controller_server.py, line 30 at r4](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1#-LchHikz9-We2vlNlb2j:-LckfW7QDLruSdXUpPYH:b2nhysk) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/365f07d3347645fa31c2ffbb7e925b4f0de736eb/controller/controller_server/csi_controller_server.py#L30)):*
<details><summary><i>Previously, olgashtivelman wrote…</i></summary><blockquote>

i removed this.
 but as i mentioned before this is not actually final code the create function is not written yet this is just so that there will be some logic there. 
</blockquote></details>

ack

ranhrl:
___
*[controller/controller_server/csi_controller_server.py, line 93 at r4](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/1#-LchIkUr5Zgr8smBsr58:-LckfbFK0PFdLGU-DSJt:b-wynlyb) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/365f07d3347645fa31c2ffbb7e925b4f0de736eb/controller/controller_server/csi_controller_server.py#L93)):*
<details><summary><i>Previously, olgashtivelman wrote…</i></summary><blockquote>

ember CSI is using this as well seems : https://github.com/embercsi/ember-csi/blob/master/ember_csi/ember_csi.py#L56
i dont know who else is doing python? 
</blockquote></details>

And the Go implementations, what do they use?

ranhrl:
LGTM

PR #4, opened 2019-04-15
CSI-33: added identity functionality for controller

olgashtivelman:
implement identity functionality for controller service.

(dev note: the csi tests for identity passed with this implementation)

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/4)

### DIFF
--- a/controller/controller_server/csi_controller_server.py
+++ b/controller/controller_server/csi_controller_server.py
@@ -10,6 +10,7 @@ from test_settings import  user, password, array, vol_name
 
 logger = get_stdout_logger()
 
+
 class ControllerServicer(csi_pb2_grpc.ControllerServicer):
     """
     gRPC server for Digestor Service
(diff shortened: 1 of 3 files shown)


controller/controller_server/csi_controller_server.py
```diff
@@ -75,17 +76,41 @@ def GetCapacity(self, request, context):
...
+        context.set_code(grpc.StatusCode.OK)
         return csi_pb2.ControllerGetCapabilitiesResponse(
            capabilities=[csi_pb2.ControllerServiceCapability(
                             rpc=csi_pb2.ControllerServiceCapability.RPC(type=types.Value("CREATE_DELETE_VOLUME"))),
                          csi_pb2.ControllerServiceCapability(
                              rpc=csi_pb2.ControllerServiceCapability.RPC(type=types.Value("PUBLISH_UNPUBLISH_VOLUME")))  ])
  
+    def GetPluginInfo(self, request, context):
+        logger.debug("GetPluginInfo")
+        context.set_code(grpc.StatusCode.OK)
```
shay-berman: Not sure you need to do that, because by default its OK.
I think only for error you should set the grpc status code.  (relevant for other places in the code here). 

if in the code we should raise error(e.g  For example if no name in conf file) then u can use context.abort(grpc.StatusCode.<FAIL_CONST>) which raises the relevant grpc error.  and then also use it in the UT to check if exception raised.

controller/controller_server/csi_controller_server.py
```diff
@@ -75,17 +76,41 @@ def GetCapacity(self, request, context):
...
+        return csi_pb2.GetPluginInfoResponse(name="ibm-block-csi-driver", vendor_version="1.0.0")
+    
+    def GetPluginCapabilities(self, request, context):
+        logger.debug("GetPluginCapabilities")
+        context.set_code(grpc.StatusCode.OK)
```
shay-berman: don't think its needed as mentioned above.

controller/controller_server/csi_controller_server.py
```diff
@@ -75,17 +76,41 @@ def GetCapacity(self, request, context):
...
+        types = csi_pb2.PluginCapability.Service.Type
+        return csi_pb2.GetPluginCapabilitiesResponse(
+            capabilities=[csi_pb2.PluginCapability(
+                            service=csi_pb2.PluginCapability.Service(type=types.Value("CONTROLLER_SERVICE")))
```
shay-berman: since identity should be done also in python for node, then I think its better also here to put the CONTROLLER_SERVER as one of the consts inside the plugin.conf (maybe like a list IdentityCapabilities=CONTROLLER_SERVER), and just use it from here.

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -50,3 +51,27 @@ def test_create_volume_exception(self, a_enter, get_volume):
         res = self.servicer.CreateVolume(request, context)
         context.set_code.assert_called_once_with(grpc.StatusCode.INTERNAL)
         self.mediator.get_volume.assert_called_once_with(vol_name)
+
+    def test_controller_get_capabilities(self):
+        request = Mock()
+        context = Mock()
+        res = self.servicer.ControllerGetCapabilities(request, context)
+        context.set_code.assert_called_once_with(grpc.StatusCode.OK)
```
shay-berman: instead u can check if exception was not raised.

controller/controller_server/csi_controller_server.py
olgashtivelman: why should identity be done in python also for the node?

olgashtivelman:
implement identity functionality for controller service.

PR #27, opened 2019-07-02
Feature/csi node docker build in CI pipeline

shay-berman:
# Describe
Build csi node as docker image and trigger it as part of the build pipeline in jenkins. So from now on there is ready csi node image in the internal artifactory after every push in git (together with csi controller image which already there).

# Full detail
## new  `Dockerfile-csi-node` file
- Building csi node based on centos7 (maybe we will change to UBI later) and locate the exec inside /root and also copy the config.yaml file.
- Uses multistage build for compilation:
   - Uses go.1.12.6 image
   - Make sure go mod in dedicated docker cache layer to avoid downloading every time.
   - Using `make ibm-block-csi-driver` inside the dockerfile to build the csi node (so dockerfile use the make but also develop can use it locally to compile)
   - Remove .git from .dockerignore in order to have the .git inside the docker image build stage to get the commit hash during the make build.

## new  `Dockerfile-csi-node.test` file
- Almost the same as above but just for running go unit testing inside container during CI pipeline. For that we need parser for xunit and also trigger make test inside the test container.

## Makefile new targets 
- `test-xunit` is for run tests locally and with xunit output
- `test-xunit-in-container` is for run the above inside container in CI pipeline. Which eventually save the xunit of the testing inside $(CURDIR)/build/reports/ so the pipeline collect it to the job.
- `csi-build-images-and-push-artifactory` to run the build and push to registry as a Makefile target (just nice convention instead of script). Adding this target to the CI pipeline.

## scripts/ci/build_push_images.sh updated 
- Now this script build and push csi node image (not only csi controller)

## jenkins_pipeline_csi new phase
1. Adding `CSI-node: go Unit testing` phase in CI pipeline.
2. Update the name of the build phase to `CSI-controller & node: Build and push images` since now it also build the csi node image.
So now the build pipeline look like this(every phase runs in its own container) ->

image

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/27)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,2 @@
-# ignore .git and .cache folders
-.git
+# ignore .cache folder
 .cache
--- a/Dockerfile-csi-node
+++ b/Dockerfile-csi-node
@@ -24,7 +24,7 @@ COPY go.sum .
 RUN go mod download
 
 COPY . .
-RUN make 
+RUN make ibm-block-csi-driver
 
 # Final stage
 FROM centos:7
--- a/Dockerfile-csi-node
+++ b/Dockerfile-csi-node
@@ -1,0 +1,35 @@
+# Copyright IBM Corporation 2019.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Build stage
+FROM golang:1.12.6 as builder
+
+WORKDIR /go/src/github.com/ibm/ibm-block-csi-driver
+ENV GO111MODULE=on
+
+# Populate the module cache based on the go.{mod,sum} files.
+COPY go.mod .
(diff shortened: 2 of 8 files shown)


ranhrl:
___
*[go.mod, line 8 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/27#-LixdJHS2y5SU2KaKDYn:-LixdJHS2y5SU2KaKDYo:bgi9n96) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/05db94d88bbe69cdde7f4fbf78462c66ab762b64/go.mod#L8)):*
> ```XML
> 	github.com/container-storage-interface/spec v1.1.0
> 	github.com/golang/mock v1.1.1
> 	github.com/tebeka/go2xunit v1.4.10
> ```

Do we ship it? Is it in the list of open source requirements for legal?

ranhrl:
___
*[Makefile, line 42 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/27#-LixdyHs5wIlfQJMhgxl:-LixdyHs5wIlfQJMhgxm:b-o4dleu) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/05db94d88bbe69cdde7f4fbf78462c66ab762b64/Makefile#L42)):*
> ```Makefile
> 	mkdir -p ./build/reports
> 	go test -v -race ./node/... | go2xunit -output build/reports/csi-node-unitests.xml
> 	go test -v -race ./node/...	# run again so the makefile will fail in case tests failing
> ```

Is this some kind of a work around?

ranhrl:
:lgtm:

PR #30, opened 2019-07-03
csi-node daemonset in to the plugin manifest

shay-berman:
Adding daemonset for csi-node container and its side cars containers(liveness-probe, node-driver-registrar).

The csi-node container (ibm-block-csi-node):
- privileged container
- Pass the spec.nodeName to the csi node (Used for GetNodeInfo)
- The container has the following host path : 
        path: /var/lib/kubelet/pods         ## This volume is where the driver mounts volumes
        path: /var/lib/kubelet/plugins/ibm-block-csi-driver/   ## This volume is where the socket for kubelet->driver communication is done
        path: /var/lib/kubelet/plugins
       /var/lib/kubelet/plugins_registry/    ## This volume is where the node-driver-registrar registers the plugin with kubelet
       /dev           ## To discover the new devices
       /etc/iscsi/   ## To retrieve the IQN for the GetNodeInfo API
       /sys            ## To discover the new devices

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/30)

### DIFF
--- a/deploy/kubernetes/v1.13/ibm-block-csi-driver.yaml
+++ b/deploy/kubernetes/v1.13/ibm-block-csi-driver.yaml
@@ -217,7 +217,7 @@ spec:
           imagePullPolicy: "IfNotPresent"
           args :
             - --csi-endpoint=$(CSI_ENDPOINT)
-#            - --loglevel=$(CSI_LOGLEVEL)
+            - --loglevel=$(CSI_LOGLEVEL)
           env:
             - name: CSI_ENDPOINT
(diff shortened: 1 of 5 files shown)


ranhrl:
:lgtm:

shay-berman:
I see that @olgashtivelman  approved and Ran put LGTM so I will merge to dev. Then Olga will merge the PR of the nodegetinfo which may have a one line conflict with this one, so olga notice about the conflict and just resolve it in your PR.

PR #32, opened 2019-07-07
Feature/csi 44 node get info implementation

olgashtivelman:
In this PR the following was done:
1. adding flags to pass the config_file + the hostname to the node. 
2. This PR implements the nodeGetInfo -> will use the /etc/iscsi/initiatorname.iscsi file to read the iqn, and build the nodeId using : "hostname:nodeid". 
3. this PR is doing unittests using gomock. for this there is a line added to the makefile : "go generate" which will create the mocks. and there are some changes in the tests to allow using mocks (the mock is for node_utils)

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/32)


shay-berman:
@olgashtivelman  i just resolved the conflict from dev, so it will take your update.

please continue review my comments and lets merge it

PR #42, opened 2019-08-04
Minor formatting proposals

kant:
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/42)


shay-berman:
Unfortunately currently we cannot get outside contribution at this point in time.
But thanks for the feedback.

kant:
Thanks for the feedback!

PR #43, opened 2019-08-08
Refactor - NodePublishVolume will do the mount instead of NodeStage

shay-berman:
Suggestion for moving the mount operation from the nodestage to the nodepublishvolume.

1. nodestage will do the discovery and also mkfs (and testing mount to nodestage, but after testing it will unmount it). 
2. nodestage creates the nodeinfo file to the stagingTarget directory (which is not mounted!).
3. Nodepublish volume will get the dm device from the nodestage info file (on the stagingPath) and will mount it to the targetPath(the final path of the pod)
4. Nodeunpublishvolume will do the same as before, just unmount the targetPath.
5. nodeunstage will read the stage info file and then delete the dm devices and the sdX devices. (no need to unmount the stagingPath because its never mounted to the staging path with this proposal)

Pros:
1. stageinfo file stored on the node filesystem and not inside the PV mount point. So we don't mass with customer volume data and that way we can also support RO mount (because the staging info file is not storing on the PV it self).
2. Only one mount point on the host, which is simpler.
3. Less idempotancy issues because nodeunstage node will delete the stage info file at the end of the nodeunstage method (without risking of loosing this info file).

Cons:
- Added mount for checking inside the nodestage because i wanted to use k8s mounter tool FormatAndMount, so after success we unmount it.

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/43)

### DIFF
--- a/node/pkg/driver/device_connectivity/errors.go
+++ b/node/pkg/driver/device_connectivity/errors.go
@@ -35,7 +35,7 @@ type ConnectivityIscsiStorageTargetNotFoundError struct {
 }
 
 func (e *ConnectivityIscsiStorageTargetNotFoundError) Error() string {
-	return fmt.Sprintf("Connectivity Error: Storage target name [%s] was not found on the host, under directory %s", e.StorageTargetName, e.DirectoryPath)
+	return fmt.Sprintf("Connectivity Error: Storage target name [%s] was not found on the host, under directory %s. Please check the host connectivity to the storage.", e.StorageTargetName, e.DirectoryPath)
 }
 
 type MultipleDeviceNotFoundError struct {
@@ -44,7 +44,7 @@ type MultipleDeviceNotFoundError struct {
 }
 
 func (e *MultipleDeviceNotFoundError) Error() string {
-	return fmt.Sprintf("Couldn't find dm-* of the physical device path [%s -> %s] ", e.DiskByPathDevice, e.LinkToPhysicalDevice)
+	return fmt.Sprintf("Couldn't find dm-* of the physical device path [%s -> %s]. Please check the host connectivity to the storage.", e.DiskByPathDevice, e.LinkToPhysicalDevice)
 }
 
 type ErrorNothingWasWrittenToScanFileError struct {
--- a/node/pkg/driver/node.go
+++ b/node/pkg/driver/node.go
@@ -140,6 +140,23 @@ func (d *NodeService) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 	if _, err := os.Stat(stagingPath); os.IsNotExist(err) {
 		klog.V(4).Infof("Target path directory does not exist. creating : {%v}", stagingPath)
 		d.mounter.MakeDir(stagingPath) // TODO consider to chmod the directory
+	} else {
+		// checking idempotent case 
+		_, err := os.Stat(stageInfoPath)
+		if err == nil {
+			klog.V(4).Infof("Idempotent case: node stage was already called before on this path. checking stage info")
+			// lets read the file and comprae the stageInfo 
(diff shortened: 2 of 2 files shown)


ranhrl:
___
*[node/pkg/driver/node.go, line 167 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/43#-LlzbM0gACifk3qfkZ87:-LlzbM0gACifk3qfkZ88:b-6sqz2k) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/72f6e234c49714d92566501679878109b947c946/node/pkg/driver/node.go#L167)):*
> ```Go
> 	}
> 
> 	err = d.mounter.FormatAndMount(device, stagingPath, fsType, nil) // TODO: pass mount options
> ```

Are you resilient to a previous error in this new way? If you try to mount and it is already mounted? (could happen if the function failed between the "mount test" and "unmount".

ranhrl:
___
*[node/pkg/driver/node.go, line 176 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/43#-Llzc18L8r6noI9pdzqF:-Llzc18L8r6noI9pdzqG:b-z5q5fm) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/72f6e234c49714d92566501679878109b947c946/node/pkg/driver/node.go#L176)):*
> ```Go
> 	if err != nil {
> 		return nil, status.Errorf(codes.Internal, "Could not unmount target %q: %v during NodeStage mount check.", stagingPath, err)
> 	}
> ```

Consider to put the "preparation phase" (makedir, mkfs, mount) in a separate function.

ranhrl:
___
*[node/pkg/driver/node.go, line 361 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/43#-LlzcxFgDGJhuJKEuOA7:-LlzcxFgDGJhuJKEuOA8:briyml1) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/72f6e234c49714d92566501679878109b947c946/node/pkg/driver/node.go#L361)):*
> ```Go
> 	mountList, err := d.mounter.List()
> 	for _, mount := range mountList {
> 		klog.V(5).Infof("mount device : {%v}, path : {%v} is equel to targetPath {%s}", mount.Device, mount.Path, targetPath)
> ```

The info message is misleading, you mean that you are checking if mount device is equal to target path.

ranhrl:
___
*[node/pkg/driver/node.go, line 361 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/43#-LlzcxFgDGJhuJKEuOA7:-LlzdEVg7MRCZNBE2Bw0:b-pj3112) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/72f6e234c49714d92566501679878109b947c946/node/pkg/driver/node.go#L361)):*
<details><summary><i>Previously, ranhrl (Ran Harel) wrote…</i></summary><blockquote>

The info message is misleading, you mean that you are checking if mount device is equal to target path.
</blockquote></details>

Means that you should write to the log: Checking if target path is already mounted on mount point.

ranhrl:
___
*[node/pkg/driver/node.go, line 369 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/43#-LlzdThn0kFVzUWBIy6K:-LlzdThn0kFVzUWBIy6L:b4h1wlt) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/72f6e234c49714d92566501679878109b947c946/node/pkg/driver/node.go#L369)):*
> ```Go
> 	}
> 
> 	// Read staging info file in order to find the mpath device for mounting.
> ```

"reading from file", consider to log it instead of comment it

PR #46, opened 2019-08-13
Fix/csi-sanity ready for node APIs and new flat config keys

shay-berman:
## Description
This PR is to return back to green the CI pipeline of the CSI community testing. First by fixing the keys in the storage class and second by adding node container as part of the testing.

## Full detail

1. Since the storage class param where changed the csi community testing should be aligned.
    - Fix the csi_params so pool will be in the main level.
2. Make the tests ready for adding new CSI node tests by setting up node docker container as part of `run_community_csi_test.sh` script. 
   - The csi community testing has a feature to allow running node and controller if different sockets.
   - The node should run and listen to dedicated socket `ENV ENDPOINT="/tmp/k8s_dir/nodecsi" ` and the csi test container should use it for node operations. Its done by new script `run_node_server_for_csi_test.sh `.
      
   - The controller container will continue to use `ENV ENDPOINT_CONTROLLER="/tmp/k8s_dir/f"`
   -  Running the node container with redirection to ${CONTROLLER_LOGS}_node file that the pipeline should collect.
   - The script kill the csi-controller and csi-node containers after done.
   - The csi tests container must run with root user (if not he cannot connect to the socket for some reason, anyway it just for internal UT).

3. Enable basic node APIs NodeGetCapabilities, 	NodeGetInfo to make sure the node works ok.

This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/46)


shay-berman:
Tested in the CI, and its now green. 
Merging to develop after @igorgonibm  review, and I will open a new PR to add more CSI node testing.

PR #49, opened 2019-08-15
CSI-383 get_array_one_iscsi_name_for_svc

xyanyang:
currently only return one array iqn, need to return all later

### DIFF
--- a/controller/array_action/array_mediator_svc.py
+++ b/controller/array_action/array_mediator_svc.py
@@ -367,3 +367,18 @@ class SVCArrayMediator(ArrayMediator):
         except Exception as ex:
             logger.exception(ex)
             raise ex
+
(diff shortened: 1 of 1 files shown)


controller/array_action/array_mediator_svc.py
```diff
@@ -367,3 +367,19 @@ def unmap_volume(self, volume_id, host_name):
         except Exception as ex:
             logger.exception(ex)
             raise ex
+
+
+    def get_array_iscsi_name(self):
```
ranhrl: seems that get_array_iscsi_name for XIV returns a list, see https://github.com/IBM/ibm-block-csi-driver/blob/develop/controller/array_action/array_mediator_xiv.py

PR #50, opened 2019-08-18
CSI-383 implement get_array_iscsi_name for svc and update multiple ar…

xyanyang:
add UT for get_array_iscsi_name
update UT for device_connectivity_iscsi.go to test one more array iqns
move split arrayIdentifier string action to GetInfoFromPublishContext

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/50)


node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
```diff
@@ -48,10 +48,26 @@ var (
...
+		e := &ErrorNotFoundArrayIdentifiers{lunId}
+		klog.Errorf(e.Error())
+		return e
+	}
+
+	for _, iqn := range arrayIdentifiers{
+		h, e := r.Helper.GetIscsiSessionHostsForArrayIQN(iqn)
+		if e != nil {
+			klog.Errorf(e.Error())
+			err = e
```
ranhrl: What kind of errors are you expecting here?
There is an option that you will get multiple errors, or some disconnected sessions but you will still go forward to the next identifier, and you will still append the session. The logic is not clear to me, can you explain what you mean to do?
What is 'h'? can you use a more self explanatory variable name?

node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
```diff
@@ -97,22 +113,32 @@ func (r OsDeviceConnectivityIscsi) GetMpathDevice(volumeId string, lunId int, ar
...
+		dps, exists, e := r.Helper.WaitForPathToExist(dp, 5, 1)
+		if e != nil {
+			klog.Errorf("GetMpathDevice: No device found error : %v ", e.Error())
+			err = e
+		} else if !exists {
+			e := &MultipleDeviceNotFoundForLunError{volumeId, lunId, []string{iqn}}
+			klog.Errorf(e.Error())
+			err = e
+		}
+		devicePaths = append(devicePaths, dps...)
```
ranhrl: If there is an error above, what happens here? does it still append the dps to the devicePaths?

node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
```diff
@@ -133,12 +159,12 @@ func (r OsDeviceConnectivityIscsi) GetMpathDevice(volumeId string, lunId int, ar
...
+	klog.V(4).Infof("GetMpathDevice: Found multipath devices: [%s] that relats to lunId=%d and arrayIdentifiers=%s", mps, lunId, arrayIdentifiers)
 
 	if len(devicePathTosysfs) > 1 {
-		return "", &MultipleDmDevicesError{volumeId, lunId, arrayIdentifier, devicePathTosysfs}
+		return "", &MultipleDmDevicesError{volumeId, lunId, arrayIdentifiers, devicePathTosysfs}
 	} else if len(devicePathTosysfs) == 0 {
-		return "", &MultipleDeviceNotFoundForLunError{volumeId, lunId, arrayIdentifier}
+		return "", &MultipleDeviceNotFoundForLunError{volumeId, lunId, arrayIdentifiers}
 	}
 	var md string
```
ranhrl: Do we now ever get to this lines of code or are they "dead"?

node/pkg/driver/device_connectivity/device_connectivity_iscsi_test.go
```diff
@@ -199,13 +199,15 @@ func TestGetMpathDevice(t *testing.T) {
...
 			fake_helper := mocks.NewMockOsDeviceConnectivityHelperIscsiInterface(mockCtrl)
 			lunId := 0
-			arrayIdentifier := "X"
-			path := strings.Join([]string{"/dev/disk/by-path/ip*", "iscsi", arrayIdentifier, "lun", strconv.Itoa(lunId)}, "-")
-			fake_helper.EXPECT().WaitForPathToExist(path, 5, 1).Return(
-				tc.waitForPathToExistReturn.devicePaths,
-				tc.waitForPathToExistReturn.exists,
-				tc.waitForPathToExistReturn.err,
-			)
+			arrayIdentifiers := []string{"X"}
```
ranhrl: I think that what we need is one test for one iSCSI name and a separate test to having 2 or more iSCSI names.

node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
xyanyang: I think we can ignore the errors when the length of sessionHosts is not 0.
for different scsi hosts(/sys/class/scsi_host/hostX/scan), they are just the different paths of the same physical device, 
so if get iscsi session hosts for one array iqn failed, we just need to continues with the next one.

if it's 0, I only returned the last one error here, it's not very resonable, it better to add all the error messages together.

for the h, I will update it, Thanks

node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
xyanyang: if there is an error, we will get nil as the dps here, devicePaths has not any change when appends nil

node/pkg/driver/device_connectivity/device_connectivity_iscsi_test.go
xyanyang: yes, I will add test cases about one more iscsi names

node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
xyanyang: Yes, I will remove this else if block

controller/array_action/array_mediator_svc.py
```diff
@@ -367,19 +367,10 @@ def unmap_volume(self, volume_id, host_name):
         except Exception as ex:
             logger.exception(ex)
             raise ex
-
-
+    
     def get_array_iscsi_name(self):
         logger.debug("Getting array nodes iscsi name")
         nodes_list = self.client.svcinfo.lsnode().as_list
```
27149chen: you don't need .as_list, all the commands return an iterator, you can use it in for...in... directly

node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
```diff
@@ -48,10 +48,26 @@ var (
 	TimeOutMultipathFlashCmd = 4 * 1000
 )
 
-func (r OsDeviceConnectivityIscsi) RescanDevices(lunId int, arrayIdentifier string) error {
-	klog.V(5).Infof("Rescan : Start rescan on specific lun, on lun : {%v}, with array iqn : {%v}", lunId, arrayIdentifier)
-	sessionHosts, err := r.Helper.GetIscsiSessionHostsForArrayIQN(arrayIdentifier)
-	if err != nil {
+func (r OsDeviceConnectivityIscsi) RescanDevices(lunId int, arrayIdentifiers []string) error {
```
27149chen: arrayIdentifier is not a reasonable name, consider to rename it.

node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
```diff
@@ -48,10 +48,26 @@ var (
...
+	klog.V(5).Infof("Rescan : Start rescan on specific lun, on lun : {%v}, with array iqn : {%v}", lunId, arrayIdentifiers)
+	var sessionHosts []int
+	var err error
+
+	if len(arrayIdentifiers) == 0 {
+		e := &ErrorNotFoundArrayIdentifiers{lunId}
```
27149chen: also this error name

node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
27149chen: If first time e != nil and second time exists == false, final err will be deviceNotFound, I don't think it is an expected err

controller/array_action/array_mediator_svc.py
xyanyang: ok, I will update it

node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
xyanyang: I will append all the error messages together

xyanyang (commented):
> *[Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/50)* status: 0 of 13 files reviewed, 8 unresolved discussions (waiting on @27149chen, @ranhrl, @shay-berman, and @wangpww)

___
*[controller/array_action/array_mediator_svc.py, line 375 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/50#-Ln6BGRXAzUlrFUDouvf:-Ln7bi2c5g7w2xkyiLs6:b84yjf0) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/0864887dfd54893130f2f39ea266d4f3dea77249/controller/array_action/array_mediator_svc.py#L375)):*

second though
its a bit strange that you return here a string of iqn,iqn,iqn...

sounds like you need to return a list of iqn and not string with , delimiter.
In the controller where you need to use it to add to the publish context(generate_csi_publish_volume_response) you need to replace the list to string. In addition when you do that don't forget to return list also on xiv mediator (so last line return [array_iqn], so you will not break A9000)


Also don't forget to update the interface docstring (array_mediator_interface) get_array_iscsi_name() that now it return list. (BTW you forgot to update it also in the other approch where you return string but wit iqn,iqn,iqn)

Do you agree of this change?
</blockquote></details>

done, Thanks for your suggestion
___
*[node/pkg/driver/device_connectivity/device_connectivity_iscsi.go, line 124 at r2](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/50#-LmiQ1BcCFu4vthhouW8:-Ln77__yEiogyDl2bRiX:b82uws5) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/8bab397c4dca2c951c5efabf673847174b40c97f/node/pkg/driver/device_connectivity/device_connectivity_iscsi.go#L124)):*

sure
I also prefer your PR will go first.

which ticket you open?
</blockquote></details>

it is CSI-442

shay-berman:
@igorgonibm  - please add into the description of the PR more detail about the change to the new logger and how other should use it from now on (I want to share China team for that important change as well)

In addition, I added a short fix to the logger (-loglevel fix) inside https://github.com/IBM/ibm-block-csi-driver/pull/50, please review it and approve.

PR #56, opened 2019-08-27
FC support: CSI controller (SVC) - get_array_fc_wwns, get_host_by_host_identifiers

wangpww:
1. add new method get_array_fc_wwns() to get the wwn of the connected FC port from the storage array
2. adjust exist method get_host_by_host_identifiers to support get the host by the wwn
    a. if found more than 1 host by iqn adn wwns, return MultipleHostsFoundError
    b. if find the same host by the iqn and wwns, return 'host_name, ['iscsi', 'fc']'
    c. if find the host by iqn, return  "host_name, ['iscsi']"
    d. if find the host by wwn, return  "host_name, ['fc']"
    e. otherwise, return hostNotFound

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/56)


shay-berman:
please add reviewable link to the description (you can see how it was done in older PR. BTW it should be automatically but not sure why its stopped)

wangpww:
I have added it, please help to review it

igorgonibm:
___
*[controller/array_action/array_mediator_svc.py, line 258 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/56#-LoEp2x2-LZQ9uLcvfHl:-LoEp2x37AKaU07LD6GA:b-hm47zc) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/10e80404e4cf0bbafbc270533113da1e5fb9ea9c/controller/array_action/array_mediator_svc.py#L258)):*
> ```Python
>             else:
>                 raise controller_errors.MultipleHostsFoundError(fc_wwns,
>                                                                 fc_host)
> ```

I think error should include bot fc_wwns and iscsi_iqn

PR #57, opened 2019-08-28
FC support: CSI controller(SVC): ControllerPublishVolume alignment

wangpww:
ControllerPublishVolume adjustment to support FC
some utils method adjustment to support FC
UT add and update

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/57)


igorgonibm:
___
*[controller/array_action/array_mediator_svc.py, line 405 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/57#-LnmCmhFEHPexelYLJz9:-LnmCmhFEHPexelYLJzA:b-as94vm) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/5dddbd07260f7717458783b00a8652b9fb578cea/controller/array_action/array_mediator_svc.py#L405)):*
> ```Python
>                      "related to host : {}.".format(host_name))
>         fc_port_wwns = []
>         try:
> ```

I don't see usage of host_name argument.
Was it intended to be used in lsFabric(host=host_name) ?

igorgonibm:
___
*[controller/array_action/array_mediator_svc.py, line 252 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/57#-LnmDhdxC46YP71Ks19R:-LnmDhdxC46YP71Ks19S:bqsmvbc) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/5dddbd07260f7717458783b00a8652b9fb578cea/controller/array_action/array_mediator_svc.py#L252)):*
> ```Python
>                              "{1}".format(iscsi_iqn, iscsi_host))
>             if set(fc_wwns) == set(host_initiator_wwns):
>                 fc_host = host_detail.get('name', '')
> ```

should fc_wwns and host_initiator_wwns have the exact same set of ports?

igorgonibm:
___
*[controller/array_action/array_mediator_svc.py, line 252 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/57#-LnmDhdxC46YP71Ks19R:-LnqAaZ9-gbsUlMh5l3F:b-vi4cfg) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/5dddbd07260f7717458783b00a8652b9fb578cea/controller/array_action/array_mediator_svc.py#L252)):*
<details><summary><i>Previously, wangpww wrote…</i></summary><blockquote>

They are all list, so I use the set to ignore the order of elements
</blockquote></details>

Yes but you sure the sets should be exactly the same?
For example if fc_wwns have all the ports of host_iitiators_wwns but also have others as well? Or vice versa?

igorgonibm:
___
*[controller/array_action/array_mediator_svc.py, line 252 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/57#-LnmDhdxC46YP71Ks19R:-Lnr5VmS-QTVvOJU-U_J:b-dh4650) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/5dddbd07260f7717458783b00a8652b9fb578cea/controller/array_action/array_mediator_svc.py#L252)):*
<details><summary><i>Previously, wangpww wrote…</i></summary><blockquote>

I understood your question now. For example, node has two fc wwn, wwn1 and wwn2, but the user defined the host on storage but only add wwn1 as the port. And if the node has three fc wwns, the case is more complex. I think I need to open a ticket about it, so I can refactor it after the beta, do you agree with me?
</blockquote></details>

Please advise Shay Berman
There is the same issue with xiv (which I should handle)

PR #65, opened 2019-09-08
feature/A9000 FC support

igorgonibm:
Support fc in a9000.
- Updated array_mediator_xiv.py
    - added method get_array_fc_wwns  (get wwns targets of the array)
    - update method get_host_by_host_identifiers 

Note: internal ticket csi-446 

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/65)


igorgonibm:
___
*[controller/array_action/utils.py, line 24 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/65#-LoK0WkB9Rf4R4JFjaKv:-LoKLMT10_88D1rWBhOd:bjzz3tu) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/fef4ab1ca5481df01d3943dff3cdc0d1359fa9e1/controller/array_action/utils.py#L24)):*

sound overwhelming

I think we should make it simpler and refactor the initiators to be not as param, 1 for iscs 1 for wwn and later one for NVME.
Instead we should define it as dict of {'iscsi':[iqn1], 'fc':[wwn1,wwn2,wwn3...]}
This will make it more flexible and make it much ready for NVME.

what do you say?

then u don't really need get_all_port or many other boiler plates in the code.
</blockquote></details>

Considered it too
But there are only 3 types and there will not be new ones in the close future

igorgonibm:
___
*[controller/array_action/array_mediator_xiv.py, line 182 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/65#-LoK7Zon5cBP4xDo9jWt:-LoKLtZgDs9PUZoioSfP:b-3u9hc2) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/fef4ab1ca5481df01d3943dff3cdc0d1359fa9e1/controller/array_action/array_mediator_xiv.py#L182)):*

it could be a bug
because you may have host1 with iqn and also wwn so u will find 2 host which are the same.
if fc found you should return it, if no fc but iscs then you return iscsi.
</blockquote></details>

right. now I remember why I added boolean flag before!

igorgonibm:
___
*[controller/array_action/array_mediator_xiv.py, line 175 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/65#-LoK556A5LoKTazQYrGu:-LoKM5q8DwwlevtfcMGh:b-405wwo) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/fef4ab1ca5481df01d3943dff3cdc0d1359fa9e1/controller/array_action/array_mediator_xiv.py#L175)):*

lets talk about it and consider how to simplify it.
</blockquote></details>

couldn't find a simpler way
This why I told you that the case of host with more than one iqn made the code ugly.

igorgonibm:
___
*[controller/array_action/array_mediator_xiv.py, line 182 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/65#-LoK7Zon5cBP4xDo9jWt:-LoKOp763ziK5TDelHbq:b-7ql1td) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/fef4ab1ca5481df01d3943dff3cdc0d1359fa9e1/controller/array_action/array_mediator_xiv.py#L182)):*
<details><summary><i>Previously, igorgonibm wrote…</i></summary><blockquote>

right. now I remember why I added boolean flag before!
</blockquote></details>

fixed

igorgonibm:
___
*[controller/common/node_info.py, line 40 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/65#-LoPurGS9iOuIo3X3CNI:-LoQBYxQC6qwVcCdE-rr:b-trbvj5) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/23088dfdff0884f251ec9519f5a30c92e4cfa49d/controller/common/node_info.py#L40)):*
<details><summary><i>Previously, 27149chen (Guang Jiong Lou) wrote…</i></summary><blockquote>

Igor, please consider the problems I raised in the wiki: https://wiki.xiv.ibm.com/pages/viewpage.action?pageId=253826817
What if you find a match, but the port is offline? what if a node has several ports, but the host definition on storage side only has one port (not a good multipath)? and so on...
In short, I don't think it means that you find the right (or best) host when the intersection of the two sets is not empty.
</blockquote></details>

There is corresponding ticket for that after Beta CSI-483

PR #68, opened 2019-09-11
Fix/CSI-485 works without /etc/iscsi directory

shay-berman:
Ran tested the plugin on FC environemnt without any iscsi service on the worker nodes (which is the common use case in Fc env). And he find an issue that the csi node is failing due to missing /etc/iscsi directory.

So this PR fix 
Before this PR /etc/iscsi hostpath was specified in the yaml. So I removed this specific hostpath from the yaml and fixed the code to take /etc/iscsi from the /host/etc/iscsi which is already the hostpath of the worker node. So there is no really need to specify also /etc/iscsi in the yaml if we already have hostpath / inside csi node in /host directory (reminder - csi node requires the / of the worker node because it discover multipath devices inside the csi node container).

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/68)

### DIFF
--- a/deploy/kubernetes/v1.13/ibm-block-csi-driver.yaml
+++ b/deploy/kubernetes/v1.13/ibm-block-csi-driver.yaml
@@ -407,8 +407,6 @@ spec:
 
             - name: device-dir
               mountPath: /dev
-            - name: iscsi-dir
-              mountPath: /etc/iscsi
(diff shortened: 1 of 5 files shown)


shay-berman:
I tested it manually on my dev k8s cluster.

PR #73, opened 2019-09-16
helper script for pushing images to dockerhub

shay-berman:
- Helper script that pull an source image [registry1/image:tag] and then push to different registry target[ registry1/image:tag].
- Script also tag latest on the image is requires.
- The script do validation to make sure the image on the target is not exist already, and also validate the target image is downloadable at the end.
- It uses in the CI to publish the CSI driver new images from internal registry to DockerHub:
   https://hub.docker.com/r/ibmcom/ibm-block-csi-driver-node
   https://hub.docker.com/r/ibmcom/ibm-block-csi-driver-controller

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/73)

### DIFF
--- a/scripts/ci/helper_to_push_docker_image.sh
+++ b/scripts/ci/helper_to_push_docker_image.sh
@@ -1,0 +1,75 @@
+#!/bin/bash -xe
+
+# This script is for internal use in CI
+# Assume docker login to the external registry was done in advance.
+
+function usage()
+{
+    echo $0 [internal-image-path]  [external-image-path] [latest-tag-optional]
+    exit 1
+}
+
+[ $# -eq 2 -o $# -eq 3 ] || { usage; }
+
+internal_image=$1
+external_image=$2
+latest="$3"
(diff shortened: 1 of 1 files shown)


igorgonibm:
___
*[scripts/ci/helper_to_push_docker_image.sh, line 8 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/73#-LotByok1YpRzRdsIGqC:-LotByol0svZaakY6chO:b-2b7hz5) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/c988f03e59090ca203998ed93e9ec74e1fc87b2c/scripts/ci/helper_to_push_docker_image.sh#L8)):*
> ```Shell
> function usage()
> {
>     echo $0 [internal-image-path]  [external-image-path] [latest-tag-optional]
> ```

seems there are two spaces  between [internal ..]  [external..]

PR #79, opened 2019-09-19
BUG fix - avoid logging the syncMap content + loadorstore (not thread safe)

27149chen:
This PR tries to fix issue [CSI-505](https://jira.xiv.ibm.com/browse/CSI-505).

The problem is that we should not print a sync.Map!!

Although sync.Map is goroutine safe, you must use the methods it exposes. If not, it is not safe. So you can not print a whole sync.Map because it doesn't have such a method to return its conent. If you print the sync.Map directly, all the fields will be printed. It has 4 fields: mu, read, dirty and misses. "dirty" is a map, and it is not safe, that's the root cause we faced: during printing "dirty", some items are removed, and we enconter a "index out of range" error.

We can use sync.Map.Range to get all kvs, but I don't think it is necessary.

In addition this PR contains fix with https://github.com/IBM/ibm-block-csi-driver/pull/80 (see detail in PR #80 )

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/79)

### DIFF
--- a/node/pkg/driver/sync_lock.go
+++ b/node/pkg/driver/sync_lock.go
@@ -49,11 +49,11 @@ func (s SyncLock) GetSyncMap() *sync.Map {
 
 func (s SyncLock) AddVolumeLock(id string, msg string) error {
 	goid := util.GetGoID()
-	logger.Debugf("Lock for action %s, Try to acquire lock for volume ID=%s (syncMap=%v) goid=%d", msg, id, s.SyncMap, goid)
+	logger.Debugf("Lock for action %s, Try to acquire lock for volume ID=%s goid=%d", msg, id, goid)
 	_, exists := s.SyncMap.Load(id)
 	if !exists {
 		s.SyncMap.Store(id, 0)
-		logger.Debugf("Lock for action %s, Succeed to acquire lock for volume ID=%s (syncMap=%v) goid=%d", msg, id, s.SyncMap, goid)
+		logger.Debugf("Lock for action %s, Succeed to acquire lock for volume ID=%s goid=%d", msg, id, goid)
 		return nil
 	} else {
 		logger.Debugf("Lock for action %s, Lock for volume ID=%s is already in use by other thread. goid=%d", msg, id, goid)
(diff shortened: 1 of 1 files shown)


igorgonibm:
___
*[node/pkg/driver/sync_lock.go, line 51 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/79#-LpNB0g39g7ABmaHZJjs:-LpNB0g39g7ABmaHZJjt:brw57cu) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/bd08c1f97bc5853dee692cae834f23bb3a28a30b/node/pkg/driver/sync_lock.go#L51)):*
> ```Go
> 
> func (s SyncLock) AddVolumeLock(id string, msg string) error {
> 	goid := util.GetGoID()
> ```

Please remove volume id and goid from all log messages
These two things are added automatically to all log messages.

igorgonibm:
___
*[node/pkg/driver/sync_lock.go, line 51 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/79#-LpNB0g39g7ABmaHZJjs:-LpNKePZ87axulWtGtiN:b653eqb) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/bd08c1f97bc5853dee692cae834f23bb3a28a30b/node/pkg/driver/sync_lock.go#L51)):*
<details><summary><i>Previously, igorgonibm wrote…</i></summary><blockquote>

Please remove volume id and goid from all log messages
These two things are added automatically to all log messages.
</blockquote></details>

I'll fix it

igorgonibm:
___
*[node/pkg/driver/sync_lock.go, line 51 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/79#-LpNB0g39g7ABmaHZJjs:-LpNbk2_De6H9qDXsBaf:bppv8hm) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/bd08c1f97bc5853dee692cae834f23bb3a28a30b/node/pkg/driver/sync_lock.go#L51)):*
<details><summary><i>Previously, igorgonibm wrote…</i></summary><blockquote>

I'll fix it
</blockquote></details>

Fixed in this branch

shay-berman:
@igorgonibm  thanks for running staging on the last commit you added.
@27149chen please go ahead and merge it to develop tomorrow morning.

shay-berman:
CI a9000 and svc passed OK.
Since @27149chen  is on vacation, I merged #80 into this PR and now I will merge it all into develop with @igorgonibm .

PR #80, opened 2019-09-22
Fix/csi 507 syncloadorstore

shay-berman:
This PR fix thread save issue Lou found in ticket CSI-507.
Using LoadOrStorage instead of Load and then Store. And also just use Delete instead of Load and then Delete.

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/80)

### DIFF
--- a/node/pkg/driver/sync_lock.go
+++ b/node/pkg/driver/sync_lock.go
@@ -48,9 +48,8 @@ func (s SyncLock) GetSyncMap() *sync.Map {
 
 func (s SyncLock) AddVolumeLock(id string, msg string) error {
 	logger.Debugf("Lock for action %s, Try to acquire lock for volume", msg)
-	_, exists := s.SyncMap.Load(id)
+	_, exists := s.SyncMap.LoadOrStore(id, 0)
 	if !exists {
-		s.SyncMap.Store(id, 0)
 		logger.Debugf("Lock for action %s, Succeed to acquire lock for volume", msg)
 		return nil
 	} else {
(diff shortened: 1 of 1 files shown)


shay-berman:
Run CI sanity of this PR to make sure it works ok.

igorgonibm:
___
*[node/pkg/driver/sync_lock.go, line 64 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/80#-LpS-_eG09ZwuyYvfD3f:-LpS-_eG09ZwuyYvfD3g:b-ax4ifu) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/69f25791e3f43634ea1877d0b37a6c9629f80703/node/pkg/driver/sync_lock.go#L64)):*
> ```Go
> 	logger.Debugf("Lock for action %s, release lock for volume", msg)
> 
> 	s.SyncMap.Delete(id)
> ```

Will it not fail if id is not in map?

igorgonibm:
___
*[node/pkg/driver/sync_lock.go, line 60 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/80#-LpS0P6E701AqJqXuZM-:-LpS0P6E701AqJqXuZM0:by5kh01) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/bd08c1f97bc5853dee692cae834f23bb3a28a30b/node/pkg/driver/sync_lock.go#L60)):*
> ```Go
> 	} else {
> 		logger.Debugf("Lock for action %s, Lock for volume ID=%s is already in use by other thread. goid=%d", msg, id, goid)
> 		return &VolumeAlreadyProcessingError{id}
> ```

I already removed log goid/volid from log messages and line goid := util.GetGoID()

PR #86, opened 2019-09-29
fix CSI-528

wangpww:
Bug fix for CSI-528
The root cause is that our code will converts the iqn to a list regardless of whether it is a list or not

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/86)

### DIFF
--- a/controller/array_action/array_mediator_svc.py
+++ b/controller/array_action/array_mediator_svc.py
@@ -239,7 +239,9 @@ class SVCArrayMediator(ArrayMediator):
             wwns_value = host_detail.get('WWPN', [])
             if not isinstance(wwns_value, list):
                 wwns_value = [wwns_value, ]
-            if initiators.is_array_iscsi_iqns_match([iscsi_names]):
+            if not isinstance(iscsi_names, list):
+                iscsi_names = [iscsi_names, ]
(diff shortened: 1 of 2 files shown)


ranhrl:
___
*[controller/tests/array_action/svc/array_mediator_svc_test.py, line 272 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/86#-LpvanMiE9qaXYOquy2h:-LpvanMiE9qaXYOquy2i:bcq8nvf) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/48ae346769010483aceba0ddf8ad32ad9815129f/controller/tests/array_action/svc/array_mediator_svc_test.py#L272)):*
> ```Python
>                                   'WWPN': ['abc1']})
>         host_munch_ret_2 = Munch({'id': 'host_id_2', 'name': 'test_host_3',
>                                   'iscsi_name': ['iqn.test.2'],
> ```

you don't have a test case that actually returns a list with more than one item. I would expect you to add one to test the code.

shay-berman:
@wangpww  please when u open PR assign it to you and also put milestone. and if its a bug add BUG label to it.
Thanks

PR #88, opened 2019-09-29
update svc volume name max length

wangpww:
from 64->63

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/88)


shay-berman:
@wangpww  @xyanyang 
Before approve make sure unit testing passing OK.
Because now the build job is red due to unit tests that were not aligned with this change.

PR #90, opened 2019-10-06
Feature/csi 500 yaml safe load

igorgonibm:
[CSI security code scan]: Use of unsafe yaml load
Use safe_load instead of load to load yaml file in controller

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/90)


ranhrl:
:lgtm:

PR #92, opened 2019-10-12
fix performance issue csi-536 slow svc map operation

xyanyang:
fix the performance issue "CSI-536 slow svc map operation" for ControllerPublishVolume/ControllerUnpublishVolume

since we only have 2 storage ssh connections to handle grpc request, so it's easily to encounter the Exception NoConnectionAvailableException.
when external-attacher receives grpc error, it will put this VA(VolumeAttachment) back to vaQueue with exponential backoff retry time(that is delay to put it back to queue).
so the request interval will reach 5 minutes after several times failure(retryIntervalMax is 5*time.Minute)

solution
set a timeout 11s(since csiTimeout for kubelet is 15 second), during this time, retry to map volume if encounter the NoConnectionAvailableException, response to external-attacher after the timeout

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/92)

### DIFF
--- a/controller/array_action/array_mediator_action.py
+++ b/controller/array_action/array_mediator_action.py
@@ -33,7 +33,7 @@ def map_volume(user, password, array_addresses, array_type, vol_id, initiators):
                 if mapping == host_name:
                     logger.debug("idempotent case - volume is already mapped to host.")
                     return mappings[mapping], connectivity_type, array_initiators
-            raise controller_errors.VolumeMappedToMultipleHostsError(mappings, vol_id)
+            raise controller_errors.VolumeMappedToMultipleHostsError(mappings)
 
         logger.debug(
             "no mappings were found for volume. mapping vol : {0} to host : {1}".format(
--- a/controller/array_action/array_mediator_action.py
+++ b/controller/array_action/array_mediator_action.py
@@ -1,0 +1,69 @@
+from controller.array_action.array_connection_manager import ArrayConnectionManager
+from controller.common.csi_logger import get_stdout_logger
+from controller.array_action.errors import NoConnectionAvailableException
+from controller.controller_server import utils
+from controller.controller_server import messages
+import controller.array_action.errors as controller_errors
+from controller.array_action.config import FC_CONNECTIVITY_TYPE
+
+from retry import retry
+
+logger = get_stdout_logger()
+
+@retry(NoConnectionAvailableException, tries=11, delay=1)
+def map_volume(user, password, array_addresses, array_type, vol_id, initiators):
+    with ArrayConnectionManager(user, password, array_addresses, array_type) as array_mediator:
+        host_name, connectivity_types = array_mediator.get_host_by_host_identifiers(initiators)
+
+        logger.debug(
+            "hostname : {}, connectivity_types  : {}".format(host_name,
+                                                              connectivity_types))
+
+        connectivity_type = utils.choose_connectivity_type(connectivity_types)
+
+        if FC_CONNECTIVITY_TYPE == connectivity_type:
+            array_initiators = array_mediator.get_array_fc_wwns(host_name)
+        else:
+            array_initiators = array_mediator.get_array_iqns()
+        mappings = array_mediator.get_volume_mappings(vol_id)
+        if len(mappings) >= 1:
+            logger.debug(
+                "{0} mappings have been found for volume. the mappings are: {1}".format(len(mappings), mappings))
+            if len(mappings) == 1:
+                mapping = list(mappings)[0]
+                if mapping == host_name:
+                    logger.debug("idempotent case - volume is already mapped to host.")
+                    return mappings[mapping], connectivity_type, array_initiators
+            err = messages.more_then_one_mapping_message.format(mappings)
+            raise controller_errors.MappingError(vol_id, host_name, err)
+
+        logger.debug(
+            "no mappings were found for volume. mapping vol : {0} to host : {1}".format(
+                vol_id, host_name))
+
+        try:
+            lun = array_mediator.map_volume(vol_id, host_name)
+            logger.debug("lun : {}".format(lun))
+        except controller_errors.LunAlreadyInUseError as ex:
+            logger.warning(
+                "Lun was already in use. re-trying the operation. {0}".format(
+                    ex))
+            for i in range(array_mediator.max_lun_retries - 1):
+                try:
+                    lun = array_mediator.map_volume(vol_id, host_name)
+                    break
+                except controller_errors.LunAlreadyInUseError as inner_ex:
+                    logger.warning(
+                        "re-trying map volume. try #{0}. {1}".format(i,
+                                                                     inner_ex))
+            else:  # will get here only if the for statement is false.
+                raise ex
+
+        return lun, connectivity_type, array_initiators
+
+@retry(NoConnectionAvailableException, tries=11, delay=1)
+def unmap_volume(user, password, array_addresses, array_type, vol_id, initiators):
+    with ArrayConnectionManager(user, password, array_addresses, array_type) as array_mediator:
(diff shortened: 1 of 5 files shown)


xyanyang:
> > _[Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/92)_ status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @27149chen, @shay-berman, and @xyanyang)
> 
> _[controller/controller_server/csi_controller_server.py, line 184 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/92#-Lr0dfOL-at8gCdPSrIU:-Lr0dfOM2TyTJqi2PPTI:b-fs7nok) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/e2ebf3ad415fad22450b4a6a03e187631f459992/controller/controller_server/csi_controller_server.py#L184)):_
> 
> > ```python
> >         logger.info("ControllerPublishVolume")
> >         retry = 0
> >         while retry <11:
> > ```
> 
> I think you should move the while rety<11 and try: one line above the with ArrayConnectionManager, since the exceptions related to this area only. and all the validate\get_vol_id... are static and will not be changed by each iteration.
> 
> And I would even take it one more step and instead of change a lot of code like below, I would just add decorator with retry on this specific exaption then you don't need to do any code change just 1 decorator named @retries like this -> https://www.saltycrane.com/blog/2009/11/trying-out-retry-decorator-python/
> 
> again , decorator will help to simplify the code, so the retry logic will be done in the decorator and not part of the function code, and it will be easier to maintain.
> thoughts?
> 
> _[controller/controller_server/csi_controller_server.py, line 295 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/92#-Lr0fT8q7JgLbK-loTEU:-Lr0fT8rCO8H39ljig1J:b-kqybo0) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/e2ebf3ad415fad22450b4a6a03e187631f459992/controller/controller_server/csi_controller_server.py#L295)):_
> 
> > ```python
> >         set_current_thread_name(request.volume_id)
> >         logger.info("ControllerUnpublishVolume")
> >         retry = 0
> > ```
> 
> same here
> no need to change the code, just use the retry decorator on this specific exception

I have move while retry above with the with ArrayConnectionManager
and I agree the decorator is simple for retry.
But for us, we still need to add additional code 
for current func ControllerUnpublishVolume, if catch an Exception, it will return, so we need to catch the NoConnectionAvailableException first then to raise it, let it retry.
and if it's still failed after 11 times retry, we need to return the Exception, so it seems not save code for us.
is it right?

xyanyang:
> > _[Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/92)_ status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @27149chen and @xyanyang)
> 
> _[controller/requirements.txt, line 11 at r2](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/92#-LrDPZUdEfvEvl7idaze:-LrDPZUdEfvEvl7idazf:b-amfto1) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/3f00e49061c138a5565f81021e6ebe3cf7363973/controller/requirements.txt#L11)):_
> 
> > ```
> > # SVC python client
> > pysvc==1.1.1
> > retry==0.9.2
> > ```
> 
> its new dep?
> if so then check its license, and consult with Nina about adding it. if so then add it to the dependancy list.
> 
> If possible to avoid adding new dep, it will be easier.
> 
> BTW i don't see u use retry at all in the code, do i miss something?
> 
> _[controller/controller_server/csi_controller_server.py, line 184 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/92#-Lr0dfOL-at8gCdPSrIU:-LrDQOj10ZEj3-3GFT4v:b3kre) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/e2ebf3ad415fad22450b4a6a03e187631f459992/controller/controller_server/csi_controller_server.py#L184)):_
> Previously, shay-berman wrote…
> 
> ?
> 
> _[controller/controller_server/csi_controller_server.py, line 295 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/92#-Lr0fT8q7JgLbK-loTEU:-LrDQCLREaaWNSztS7Np:b-v9bt5k) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/e2ebf3ad415fad22450b4a6a03e187631f459992/controller/controller_server/csi_controller_server.py#L295)):_
> Previously, xyanyang wrote…
> 
> I still see while and not decorator, so I am not sure what is the change you did here.
> 
> you can also consult with Lou on this

xyanyang:
1. I have use the retry as decorator in my new commit
2. and for this package license, I have talked with Nina, she will help with it.

shay-berman:
also please make sure the new dependency is OK - talk with Ni Na to get approval and update the wiki with the new dependency.

xyanyang:
> please check my comments.
> I think its LGTM, but consider to avoid using unmap in the common file just for retries.
> 
> Make sure you run staging a9000 and SVC before merging (make sure its green, since this is a major change).
> 
> > Reviewed 3 of 3 files at r3.
> > _[Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/92#-:-LrInfoS7IJANFGq1q2F:b-47lvvl)_ status: all files reviewed, 4 unresolved discussions (waiting on @27149chen and @xyanyang)
> 
> _[controller/array_action/array_mediator_action.py, line 14 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/92#-LrInMLcEz3I4kLcWUab:-LrIiYgW6uMNq195RJh6:b2fngst) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/236993d1c5d3cb52cc13db9b0748b5e402fc6fba/controller/array_action/array_mediator_action.py#L14)):_
> 
> > ```python
> > @retry(NoConnectionAvailableException, tries=11, delay=1)
> > def map_volume(user, password, array_addresses, array_type, vol_id, initiators):
> > ```
> 
> looks ok
> 
> _[controller/array_action/array_mediator_action.py, line 65 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/92#-LrInQ4aB76HjoBFl2LK:-LrInQ4aB76HjoBFl2LL:bs1qdw4) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/236993d1c5d3cb52cc13db9b0748b5e402fc6fba/controller/array_action/array_mediator_action.py#L65)):_
> 
> > ```python
> > @retry(NoConnectionAvailableException, tries=11, delay=1)
> > def unmap_volume(user, password, array_addresses, array_type, vol_id, initiators):
> > ```
> 
> looks like no code here, so maybe just add the retry directly on the xiv\svc mediator function.

this decorator is used on the code "with ArrayConnectionManager(user, password, array_addresses, array_type) as array_mediator", catch its ex NoConnectionAvailableException to retry.
so can't move it to xiv/svc function

controller/array_action/array_mediator_action.py
```diff
@@ -0,0 +1,67 @@
...
+                    ex))
+            for i in range(array_mediator.max_lun_retries - 1):
+                try:
+                    lun = array_mediator.map_volume(vol_id, host_name)
+                    break
+                except controller_errors.LunAlreadyInUseError as inner_ex:
+                    logger.warning(
+                        "re-trying map volume. try #{0}. {1}".format(i,
+                                                                     inner_ex))
+            else:  # will get here only if the for statement is false.
```
oriyarde, line 57: I think "if the for loop doesn't break" is more accurate

PR #95, opened 2019-10-23
add defer to close multipath disk in FlushMultipathDevice

xyanyang:
in the func FlushMultipathDevice, we use os.Open to open the multipath disk to check whether the disk exists or not, but we forget to close it after that.
this causes: 
if we run the "multipath -f /dev/dm-x" failed at the first time, we try to run this command again, we will get an error, it says "the disk map in use".

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/95)


xyanyang:
update the os.Open to os.Stat and the xavi jekins job is passed on the new commit, please approve it

PR #97, opened 2019-10-23
improvment/csi-477 XFS as the default fstype instead of ext4

igorgonibm:
Use default file system xfs instead of ext4
Changed default value in node.go and few relevant comments and text files (e..g. readme).

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/97)


shay-berman:
hi @igorgonibm , since we are at code freeze we are pushing only critical issue.
And this ticket was moved from the release (u can see its not on the board).
So please close this PR and open it again after the GA.

PR #100, opened 2019-10-24
add hostIPC to avoid multipath -f stuck on csi node

shay-berman:
same as https://github.com/IBM/ibm-block-csi-operator/pulls but for the old deployment method - to validate in on CI (since Operator still not run in CI)

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/100)

### DIFF
--- a/deploy/kubernetes/v1.13/ibm-block-csi-driver.yaml
+++ b/deploy/kubernetes/v1.13/ibm-block-csi-driver.yaml
@@ -391,6 +391,7 @@ spec:
             - --hostname=$(KUBE_NODE_NAME)
             - --config-file-path=./config.yaml
             - --loglevel=$(CSI_LOGLEVEL)
+          hostIPC: true
           env:
             - name: CSI_ENDPOINT
               value: unix:///csi/csi.sock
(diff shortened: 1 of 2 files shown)


deploy/kubernetes/v1.13/ibm-block-csi-driver.yaml
```diff
@@ -391,6 +391,7 @@ spec:
             - --hostname=$(KUBE_NODE_NAME)
             - --config-file-path=./config.yaml
             - --loglevel=$(CSI_LOGLEVEL)
+          hostIPC: true
```
xyanyang: this param should be added under DaemonSet.spec.template.spec.
added under spec.conatiners like here will encounter the error 
`error validating data: ValidationError(DaemonSet.spec.template.spec.containers[0]): unknown field "hostIPC" in io.k8s.api.core.v1.Container; if you choose to ignore these errors, turn validation off with --validate=false`

deploy/kubernetes/v1.14/ibm-block-csi-driver.yaml
```diff
@@ -387,6 +387,7 @@ spec:
             - --hostname=$(KUBE_NODE_NAME)
             - --config-file-path=./config.yaml
             - --loglevel=$(CSI_LOGLEVEL)
+          hostIPC: true
```
xyanyang: same as above

shay-berman:
@xyanyang  does the staging CI is green on this PR?   please confirm then i will merge

PR #102, opened 2019-11-01
update FlushMultipathDevice to fix nil pointer dereference issue

xyanyang:
fix the node crash issue
when "multipath -f dm-x" is timeout, the os.Stat(fullDevice) will return nil to the err, then call err.Error() will get a panic

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/102)


node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
```diff
@@ -206,7 +206,7 @@ func (r OsDeviceConnectivityHelperScsiGeneric) FlushMultipathDevice(mpathDevice
 	r.MutexMultipathF.Unlock()
 
 	if err != nil {
-		if _, err := os.Stat(fullDevice); os.IsNotExist(err) {
+		if _, e := os.Stat(fullDevice); os.IsNotExist(e) {
```
27149chen: good catch

PR #107, opened 2019-11-15
fix ibm-block-csi-driver-node version

wangpww:
ibm-block-csi-driver-node:9.0.0 -> ibm-block-csi-driver-node:1.0.0

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/107)


shay-berman:
@wangpww  don't see a real reason to change it.  since no one need to use these files. its deprecated -> all should be done via operator.  do u agree?

wangpww:
Yes, I agree with you, close it

oriyarde:
@wangpww can we delete this branch?

PR #114, opened 2020-01-14
Csi 752 delete snapshot a9000

oriyarde:
add DeleteSnapshot implementation


controller/array_action/array_mediator_xiv.py
```diff
@@ -211,6 +211,25 @@ def delete_volume(self, volume_id):
 
         logger.info("Finished volume deletion. id : {0}".format(volume_id))
 
+    def delete_snapshot(self, snapshot_id):
+        logger.info("Deleting snapshot with id : {0}".format(snapshot_id))
+        try:
+            snapshot_name = self._get_vol_by_wwn(snapshot_id)
```
igorgonibm: add log with snapshot name to log

controller/controller_server/csi_controller_server.py
```diff
@@ -343,9 +343,53 @@ def CreateSnapshot(self, request, context):
             return csi_pb2.CreateSnapshotResponse()
 
     def DeleteSnapshot(self, request, context):
-        # TODO
-        logger.info("Delete snapshot")
-        return csi_pb2.DeleteSnapshotResponse()
+        set_current_thread_name(request.snapshot_id)
+        logger.info("DeleteSnapshot")
```
igorgonibm: better "DelateSnapshot start"

controller/controller_server/csi_controller_server.py
```diff
@@ -343,9 +343,53 @@ def CreateSnapshot(self, request, context):
...
+        secrets = request.secrets
+
+        try:
+            utils.validate_delete_snapshot_request(request)
+
+            user, password, array_addresses = utils.get_array_connection_info_from_secret(secrets)
+
+            try:
+                array_type, snapshot_id = utils.get_volume_id_info(request.snapshot_id)
```
igorgonibm: add to utils function get_snapshot_id_info(...) and use it

controller/controller_server/csi_controller_server.py
```diff
@@ -343,9 +343,53 @@ def CreateSnapshot(self, request, context):
...
+                return csi_pb2.DeleteSnapshotResponse()
+
+            with ArrayConnectionManager(user, password, array_addresses, array_type) as array_mediator:
+
+                logger.debug(array_mediator)
+
+                try:
+                    array_mediator.delete_snapshot(snapshot_id)
+
+                except controller_errors.SnapshotNotFoundError as ex:
```
igorgonibm: why different try/catch?
Remove try and add catches to external catches

controller/controller_server/csi_controller_server.py
oriyarde: should I add a "start" to the rest of functions as well?

controller/array_action/array_mediator_xiv.py
oriyarde: there is one in "_get_vol_by_wwn" function. should we have two consecutive and similar logs?

controller/controller_server/csi_controller_server.py
oriyarde: should I make a generic function to be used by both functions?
and if so, should it:
1. take the object type as a second parameter (like "_get_snapshot_name")? - this will require a refactor to make object type a const + map it to exceptions, etc...
or 
2. copy paste the code (like the first half of "get_snapshot")?

controller/array_action/array_mediator_xiv.py
igorgonibm: didn't see it. Ok ignore

controller/controller_server/csi_controller_server.py
igorgonibm: If other methods use this way ignore it

controller/controller_server/csi_controller_server.py

oriyarde:
moved to https://github.com/IBM/ibm-block-csi-driver/pull/159

PR #118, opened 2020-01-29
CSI-1102 change version to 1.1.0

oriyarde:
update from 1.0.0 to 1.1.0

### DIFF
--- a/NOTICES
+++ b/NOTICES
@@ -5,7 +5,7 @@ This file details additional Third Party Software License Agreements
 and third party notices and information that are required 
 to be reproduced for the following programs:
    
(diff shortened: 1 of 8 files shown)


NOTICES
```diff
@@ -5,7 +5,7 @@ This file details additional Third Party Software License Agreements
 and third party notices and information that are required 
 to be reproduced for the following programs:
    
-IBM Block Storage CSI Driver version 1.0.0
+IBM Block Storage CSI Driver version 1.1.0
```
27149chen: Please don't edit this file, we should get it from legal team before release.

PR #119, opened 2020-02-04
Story/csi 1135 vol cap validation fix

igorgonibm:
Fix volume block/fs validation in con troller

### DIFF
--- a/controller/controller_server/config.py
+++ b/controller/controller_server/config.py
@@ -4,6 +4,7 @@ SUPPORTED_FS_TYPES = ["ext4", "xfs"]
 
 access_types = csi_pb2.VolumeCapability.AccessMode
 SUPPORTED_ACCESS_MODE = [access_types.SINGLE_NODE_WRITER]
+VOLUME_CAPABILITIES_FIELD_ACCESS_TYPE_MOUNT = 'mount'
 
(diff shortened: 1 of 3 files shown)


controller/tests/controller_server/utils_test.py
```diff
@@ -61,12 +62,6 @@ def test_validate_volume_capabilities(self):
         with self.assertRaises(ValidationException):
             utils.validate_csi_volume_capabilties([caps])
 
-        caps = Mock()
-        caps.mount = None
-        caps.access_mode.mode = access_types.SINGLE_NODE_READER_ONLY
-        with self.assertRaises(ValidationException):
```
oriyarde: why is this removed?

igorgonibm: mount can never be none
The parallel test is added to check block. Which makes HasField return false instead of setting mount to None

controller/controller_server/config.py
```diff
@@ -4,6 +4,7 @@
 
 access_types = csi_pb2.VolumeCapability.AccessMode
```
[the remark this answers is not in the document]
igorgonibm: renamed

oriyarde:

PR #121, opened 2020-02-10
Story/csi 1040 change mount check

igorgonibm:
Change the way we check if mount exists


oriyarde:

PR #124, opened 2020-02-23
Feature/csi 1088 compile on z systems

oriyarde:
Driver was looking for this:
/dev/disk/by-path/pci*-fc-0x5001234567891011-lun-3

But Z machines actually have this:
/dev/disk/by-path/ccw-0.0.a200-fc-0x5001234567891011-lun-3

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/124)


node/pkg/driver/device_connectivity/device_path.go
```diff
@@ -0,0 +1,6 @@
+// +build !s390x
```
27149chen: what about Power? I think you'd better remove "!"

node/pkg/driver/device_connectivity/device_path_s390x.go
```diff
@@ -0,0 +1,5 @@
+package device_connectivity
+
+// https://cgit.freedesktop.org/systemd/systemd/tree/src/udev/udev-builtin-path_id.c#n613
+// on Z-Systems, the udev subsystem is a Channel Command Word
+var FcSubsystem = "ccw"
```
27149chen: please add an ut to verify it.

Dockerfile-csi-controller
```diff
@@ -28,6 +28,8 @@ LABEL name="IBM block storage CSI driver controller" \
 
 COPY controller/requirements.txt /driver/controller/
 RUN pip3 install --upgrade pip==19.1.1
+# avoid default boringssl lib, since it does not support z systems
+ENV GRPC_PYTHON_BUILD_SYSTEM_OPENSSL=True
```
27149chen: where will this env be used?

node/pkg/driver/device_connectivity/device_path_s390x.go
oriyarde: do you mean that the ut will have two consts with the same values, and it will choose which to compare to `FcSubsystem` according to the runtime architecture?

node/pkg/driver/device_connectivity/device_path.go
oriyarde: I saw "pci" also in Power.
If I ever see anything different, I will update accordingly

Dockerfile-csi-controller
oriyarde: Actually, this line is already in develop, sorry.
This env is used by 3rd party in the next line, when pip installs (compiles) grpcio

node/pkg/driver/device_connectivity/device_path_s390x.go
```diff
@@ -0,0 +1,5 @@
+package device_connectivity
+
+// https://cgit.freedesktop.org/systemd/systemd/tree/src/udev/udev-builtin-path_id.c#n613
```
igorgonibm: Remove the link to c file.
It doesn;t help to understand and also can change or be different depending on os version

node/pkg/driver/device_connectivity/device_path.go
```diff
@@ -0,0 +1,6 @@
...
+
+package device_connectivity
+
+// For non Z-systems, the FC subsystem is pci
+var FcSubsystem = "pci"
```
igorgonibm: use const

node/pkg/driver/device_connectivity/device_path_s390x.go
igorgonibm: use const.
I think it should be private per package (use first letter in lowercase)

node/pkg/driver/device_connectivity/device_path_s390x.go
oriyarde: image

node/pkg/driver/device_connectivity/device_path.go
oriyarde: image

node/pkg/driver/device_connectivity/device_path_s390x.go
oriyarde: image

oriyarde: done. hope it's OK

node/pkg/driver/device_connectivity/device_path.go
oriyarde: turns out the pci in power is not relevant to us.
updated: https://github.com/IBM/ibm-block-csi-driver/pull/178

PR #131, opened 2020-03-10
Feature/csi 1165 iscsi login svc

oriyarde:
Add automatic iSCSI login

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131)


igorgonibm:
___
*[common/config.yaml, line 12 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M23qsr_0HkkqsvUOit7:-M23qsr_0HkkqsvUOit8:b9nfei0) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/b96cf310baeb1c33408d21cc79069dcf6578c559/common/config.yaml#L12)):*
> ```YAML
>    publish_context_array_iqn : "PUBLISH_CONTEXT_ARRAY_IQN"
>    publish_context_fc_initiators : "PUBLISH_CONTEXT_ARRAY_FC_INITIATORS"
> #  "<array_iqn>" keys (with value of iscsi target IPs)
> ```

I'm not sure we need this here but if so better to do it this way:
#<array_iqn_1>  : comma-separated list of  iqn_1 iscsi target ips
#<array_iqn_2>  : comma-separated list of  iqn_2 iscsi target ips
#...
#<array_iqn_k>  : comma-separated list of  iqn_k iscsi target ips

oriyarde:
> Are all changes specific to this svc iscsi login feature? If not, please don't introduce any code refactor in this pr.

Then please review [this small PR first](https://github.com/IBM/ibm-block-csi-driver/pull/134)

node/pkg/driver/node.go
```diff
@@ -132,6 +133,10 @@ func (d *NodeService) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("Wrong connectivity type %s", connectivityType))
 	}
 
+	if err := osDeviceConnectivity.EnsureLogin(ipsByArrayInitiator); err != nil {
```
27149chen: You should do it only once, it is a bad practice to do iscsi login in every nodeStage request.

node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
```diff
@@ -32,6 +34,44 @@ func NewOsDeviceConnectivityIscsi(executer executer.ExecuterInterface) OsDeviceC
...
 }
 
+func (r OsDeviceConnectivityIscsi) iscsiCmd(args ...string) (string, error) {
+	out, err := exec.Command("iscsiadm", args...).CombinedOutput()
+	return string(out), err
+}
+
+func (r OsDeviceConnectivityIscsi) iscsiDiscoverAndLogin(iqn, ip string) error {
+	logger.Debugf("iscsiDiscoverAndLogin: iqn {%s}, ip {%s}", iqn, ip)
+	output, err := r.iscsiCmd("-m", "discovery", "-t", "sendtargets", "-p", ip)
```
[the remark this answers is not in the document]
oriyarde: if we use `discoverydb`, does this mean we will also have to add (`-o new`) discovery records?

node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
```diff
@@ -32,6 +34,44 @@ func NewOsDeviceConnectivityIscsi(executer executer.ExecuterInterface) OsDeviceC
...
+	var err error
+	for arrayIdentifier, ips := range ipsByArrayIdentifier {
+		for _, ip := range ips {
+			err = r.iscsiDiscoverAndLogin(arrayIdentifier, ip)
+			if err == nil {
+				isAnyLoginSucceeded = true
+			}
+		}
+	}
+	if !isAnyLoginSucceeded {
```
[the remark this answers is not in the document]
oriyarde: True, but I was thinking - if a user decides to manually configure their multipath to allow a single session (per iqn), should we really prevent them?
We already have an indicative message in `MultipleDeviceNotFoundError` later in the code

node/pkg/driver/node.go
oriyarde: is it better to parse all sessions (`iscsiadm -m session`) in every nodeStage request?

igorgonibm:
___
*[controller/array_action/array_mediator_ds8k.py, line 294 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M29l9UX1_PL6jBvVLKD:-M29l9UY-5REO83DNztP:bjsc7yi) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/f254e9837494d67511250f80502916447023627e/controller/array_action/array_mediator_ds8k.py#L294)):*
> ```Python
> 
>     def get_iscsi_targets_by_iqn(self):
>         return {}
> ```

Needs to raise some sort of UnsupportedException. Is there some parallel exception in python?

igorgonibm:
___
*[controller/array_action/array_mediator_svc.py, line 403 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M29mkSo3QvRhDT3HNeA:-M29mkSo3QvRhDT3HNeB:bfw510y) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/f254e9837494d67511250f80502916447023627e/controller/array_action/array_mediator_svc.py#L403)):*
> ```Python
>     def _list_ip_ports(self):
>         try:
>             return self.client.svcinfo.lsportip(filtervalue='state=configured:failover=no')
> ```

what is failover? Are you sure this filter is must?

igorgonibm:
___
*[controller/array_action/array_mediator_svc.py, line 409 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M29nIT0CXTUy0UscbP0:-M29nIT0CXTUy0UscbP1:b-blknqg) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/f254e9837494d67511250f80502916447023627e/controller/array_action/array_mediator_svc.py#L409)):*
> ```Python
> 
>     @staticmethod
>     def _extract_ips_by_node_id(ports):
> ```

I don't like the word exract.
maybe  _create_ips_by_node_id_map ?

igorgonibm:
___
*[controller/array_action/array_mediator_svc.py, line 437 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M29oalSCVIhoOFiE7fR:-M29oalSCVIhoOFiE7fS:btzzmrl) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/f254e9837494d67511250f80502916447023627e/controller/array_action/array_mediator_svc.py#L437)):*
> ```Python
>         ips_by_iqn = self._unify_ips_by_iqn(iqns_by_node_id, ips_by_node_id)
> 
>         if ips_by_iqn and any(ips_by_iqn.values()):
> ```

Can there be a situation of iqn with ports and an other iqn without?
If so better to write warning to log for each iqn without ports.

igorgonibm:
___
*[node/pkg/driver/driver.go, line 115 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M29qjJ03Fb0_MsN7RLG:-M29qjJ03Fb0_MsN7RLH:bmtlqin) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/f254e9837494d67511250f80502916447023627e/node/pkg/driver/driver.go#L115)):*
> ```Go
> 		Publish_context_array_iqn              string
> 		Publish_context_fc_initiators          string
> 		// "<array_iqn>" keys (with value of iscsi target IPs)
> ```

same comment as for yaml
#iqn1 : .;.;
#iqn2 : ..
...

igorgonibm:
___
*[node/pkg/driver/node.go, line 137 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M29sKr01MK-5XkbnPlB:-M29sKr01MK-5XkbnPlC:b-ryqlee) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/f254e9837494d67511250f80502916447023627e/node/pkg/driver/node.go#L137)):*
> ```Go
> 
> 	if err := osDeviceConnectivity.EnsureLogin(ipsByArrayInitiator); err != nil {
> 		return nil, status.Error(codes.Internal, err.Error())
> ```

Do we wan t to return error if login fails?
User will still be able to do login manually (except RHCOS). Lets discuss this.

igorgonibm:
___
*[node/pkg/driver/node.go, line 244 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M29sxq76CYB5JEenbFS:-M29sxq76CYB5JEenbFT:br7szan) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/f254e9837494d67511250f80502916447023627e/node/pkg/driver/node.go#L244)):*
> ```Go
> 	}
> 
> 	if connectivityType == "iscsi" {
> ```

If I'll push before you use constant I added in device_connectivity.

igorgonibm:
___
*[node/pkg/driver/node_utils.go, line 99 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M29v42S6E-8_OKUGjjs:-M29v42S6E-8_OKUGjjt:b-o4e46r) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/f254e9837494d67511250f80502916447023627e/node/pkg/driver/node_utils.go#L99)):*
> ```Go
> func (n NodeUtils) GetInfoFromPublishContext(publishContext map[string]string, configYaml ConfigFile) (string, int, map[string][]string, error) {
> 	// this will return :  connectivityType, lun, ipsByArrayInitiator, error
> 	ipsByArrayInitiator := make(map[string][]string)
> ```

isn't it ips by iqn?

igorgonibm:
___
*[node/pkg/driver/device_connectivity/device_connectivity_fc.go, line 35 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M29vbw9E4dxFSRlVO7J:-M29vbw9E4dxFSRlVO7K:b-qyow9g) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/f254e9837494d67511250f80502916447023627e/node/pkg/driver/device_connectivity/device_connectivity_fc.go#L35)):*
> ```Go
> }
> 
> func (r OsDeviceConnectivityFc) EnsureLogin(_ map[string][]string) error {
> ```

add comment: // FC doesn't require login

igorgonibm:
___
*[node/pkg/driver/device_connectivity/device_connectivity_iscsi.go, line 52 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M29wJDO4VOkc0thE-om:-M29wJDO4VOkc0thE-on:b41fcw8) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/f254e9837494d67511250f80502916447023627e/node/pkg/driver/device_connectivity/device_connectivity_iscsi.go#L52)):*
> ```Go
> 	}
> 
> 	output, err = r.iscsiCmd("-m", "node", "-p", targetPortal+":3260", "-T", targetName, "--login")
> ```

create constant with value 3260 and use it here

igorgonibm:
___
*[node/pkg/driver/device_connectivity/device_connectivity_iscsi.go, line 69 at r2](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M23qbhhyxsL24qIv2c0-r2-69:-M29wj-WEiUQDt_pzfGV:bpcyt6n) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/95694494236eee869aa842d5ecde1c048850742f/node/pkg/driver/device_connectivity/device_connectivity_iscsi.go#L69)):*
<details><summary><i>Previously, oriyarde wrote…</i></summary><blockquote>

True, but I was thinking - if a user decides to manually configure their multipath to allow a single session (per iqn), should we really prevent them?
We already have an indicative message in `MultipleDeviceNotFoundError` later in the code
</blockquote></details>

I would anyway log warning if there are less than  2 successful logins.

controller/controller_server/utils.py
```diff
@@ -192,15 +192,26 @@ def generate_csi_publish_volume_response(lun, connectivity_type, config, array_i
...
+        for iqn, ips in array_initiators.items():
+            publish_context[iqn] = separator.join(ips)
+
+        array_initiators_param = config["controller"]["publish_context_array_iqn"]
+        publish_context[array_initiators_param] = separator.join(array_initiators.keys())
+    elif connectivity_type == FC_CONNECTIVITY_TYPE:
+        array_initiators_param = config["controller"]["publish_context_fc_initiators"]
+        publish_context[array_initiators_param] = separator.join(array_initiators)
+    else:
+        raise UnsupportedConnectivityTypeError(connectivity_type)
```
27149chen: Again,  this error should be raise before calling generate_xxx, no errors can be raise in this method.

node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
```diff
@@ -32,6 +36,44 @@ func NewOsDeviceConnectivityIscsi(executer executer.ExecuterInterface) OsDeviceC
 	}
 }
 
+func (r OsDeviceConnectivityIscsi) iscsiCmd(args ...string) (string, error) {
+	out, err := r.Executer.ExecuteWithTimeout(int(iscsiCmdTimeout.Seconds()), "iscsiadm", args)
+	return string(out), err
+}
+
+func (r OsDeviceConnectivityIscsi) iscsiDiscoverAndLogin(targetName, targetPortal string) error {
```
27149chen: I think portal is enough (instead of targetPortal)

node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
```diff
@@ -32,6 +36,44 @@ func NewOsDeviceConnectivityIscsi(executer executer.ExecuterInterface) OsDeviceC
...
+	logger.Debugf("iscsiDiscoverAndLogin: target: {%s}, portal: {%s}", targetPortal, targetPortal)
+	output, err := r.iscsiCmd("-m", "discovery", "-t", "sendtargets", "-p", targetPortal)
```
27149chen: You can use -o new to add the record first, but usually we only run this command
`iscsiadm -m discoverydb -t st -p port_ip:3260 --discover`

node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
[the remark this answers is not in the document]
27149chen: MultipleDeviceNotFoundError returns after rescan, but we should prevent the rescan if it will be failed for sure, because you have to clean up the path.
If you want to support single path, set it as an env.

node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
```diff
@@ -32,6 +36,44 @@ func NewOsDeviceConnectivityIscsi(executer executer.ExecuterInterface) OsDeviceC
...
+
+	output, err = r.iscsiCmd("-m", "node", "-p", targetPortal+":3260", "-T", targetName, "--login")
+	if err != nil {
+		logger.Errorf("Failed to login iSCSI: {%s}, error: {%s}", output, err)
+		return err
+	}
+	return nil
+}
+
+func (r OsDeviceConnectivityIscsi) EnsureLogin(ipsByArrayIdentifier map[string][]string) error {
```
27149chen: portalsByTarget

node/pkg/driver/node.go
27149chen: Yes, get current sessions and parse them to see if it is already login is better.
But you still need to run get_session in every request, is it fast enough that we can ignore its effect?

node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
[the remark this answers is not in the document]
oriyarde: 1. Do you mean that the login success count should be per iqn?
2. Could you elaborate about which path do I have to clean up?
3. I wish to allow a user to use a single path, but I don't want to over complicate things for them with an env var.

node/pkg/driver/node.go
oriyarde: if there are a lot of sessions (even sessions that are not related to us), it might take a lot of time (for every request)

27149chen: so you need find another way.

node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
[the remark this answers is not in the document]
27149chen: 1. not per iqn.
2. If there is only one path, the host os will find one device after rescan, but MultipleDeviceNotFoundError will be returned because it is not a multipath device, so before raising this error to caller,  you should firstly delete this device from host os.
3. ok, so do you have a better idea?

node/pkg/driver/node.go
oriyarde: I couldn't find another way. If you find another way, please write it here and I will consider it.
Until then, I will leave it as is.

node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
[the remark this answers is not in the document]
oriyarde: 1. but I thought the default multipath settings ignores single iscsi paths per iqn (not in total), am I wrong?
2. and what happens if I don't? I assume the user will want it logged in anyway.
3. my idea is to allow both multipath configurations with the same code (without requiring additional actions from the user, without checking their configuration at runtime, etc...), by paying the price of an unnecessary rescan in the rare case of a single iscsi path user that has yet to configure their multipath.conf properly (as we will require them in the user guide).

igorgonibm:
___
*[controller/array_action/array_mediator_ds8k.py, line 294 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M29l9UX1_PL6jBvVLKD:-M2EpzWvBX9z4NCOj02d:b-3qylv8) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/f254e9837494d67511250f80502916447023627e/controller/array_action/array_mediator_ds8k.py#L294)):*
<details><summary><i>Previously, oriyarde wrote…</i></summary><blockquote>

This is a comment for ds8k support PR (I only changed [] to {})
</blockquote></details>

you still can add exception.

igorgonibm:
___
*[controller/array_action/array_mediator_svc.py, line 403 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M29mkSo3QvRhDT3HNeA:-M2EqPK4-PF_S_aw_AXY:bb1byt6) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/f254e9837494d67511250f80502916447023627e/controller/array_action/array_mediator_svc.py#L403)):*
<details><summary><i>Previously, oriyarde wrote…</i></summary><blockquote>

I think it means "passive" nodes. This filter avoids transmission of a lot of irrelevant items.
see [pysvc branch](https://github.com/IBM/pysvc/commit/16f0a470d20020aaf6911a73dcc9c6b59c964db3#diff-83bb0a7194f227f07255f1be8110b966R123)
</blockquote></details>

You sure we don't want to login passive anyway?

igorgonibm:
___
*[controller/array_action/array_mediator_svc.py, line 437 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M29oalSCVIhoOFiE7fR:-M2Eqvqx78fSjGq7zQNH:bvl5m7u) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/f254e9837494d67511250f80502916447023627e/controller/array_action/array_mediator_svc.py#L437)):*
<details><summary><i>Previously, oriyarde wrote…</i></summary><blockquote>

It is possible for a single iqn to fail on all portal logins, but there is already an error for every login failure.
I don't think the failure distribution between iqns is significant enough for an additional warning.
</blockquote></details>

but what you say is in node level. Warning here will e helpful.

igorgonibm:
___
*[node/pkg/driver/node.go, line 137 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M29sKr01MK-5XkbnPlB:-M2EsHvqEhP4DnfsDQAH:boton9l) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/f254e9837494d67511250f80502916447023627e/node/pkg/driver/node.go#L137)):*
<details><summary><i>Previously, oriyarde wrote…</i></summary><blockquote>

Depends on the error reason.
What is the scenario which automatic login fails, but manual succeeds?
</blockquote></details>

Some "unexpexcted" problem related to ruNning login from pod in the code) VS  manually (on the node)

igorgonibm:
___
*[node/pkg/driver/node_test.go, line 184 at r5](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M2EtG-q9LBaUF2ywNrv:-M2EtG-q9LBaUF2ywNrw:bk74za2) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/11fe1dfc1e702d70572031be8ba2d0919fa908d3/node/pkg/driver/node_test.go#L184)):*
> ```Go
> 						PublishContextParamArrayIqn:     "iqn.1994-05.com.redhat:686358c930fe",
> 					},
> 					StagingTargetPath: "/test/path",
> ```

use stagingPath variable

igorgonibm:
___
*[node/pkg/driver/node_test.go, line 162 at r5](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M2EtOdRCuTr9fZUQyU7:-M2EtOdRCuTr9fZUQyU8:b5pvqir) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/11fe1dfc1e702d70572031be8ba2d0919fa908d3/node/pkg/driver/node_test.go#L162)):*
> ```Go
> 				req := &csi.NodeStageVolumeRequest{
> 					PublishContext:    publishContext,
> 					StagingTargetPath: "/test/path",
> ```

use stagingPath variable
My bad. Forgot to replace this one.

igorgonibm:
___
*[node/pkg/driver/node_utils.go, line 110 at r5](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M2EvKcq4w4-YwFgMZGY:-M2EvKcq4w4-YwFgMZGZ:b-oz1wch) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/11fe1dfc1e702d70572031be8ba2d0919fa908d3/node/pkg/driver/node_utils.go#L110)):*
> ```Go
> 	connectivityType := publishContext[configYaml.Controller.Publish_context_connectivity_parameter]
> 	if connectivityType == device_connectivity.ConnectionTypeISCSI {
> 		iqns := strings.Split(publishContext[configYaml.Controller.Publish_context_array_iqn], ",")
> ```

create constant for iqns separaror "," and use it here

igorgonibm:
___
*[node/pkg/driver/node_utils.go, line 113 at r5](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M2EvYISDC7JJr7wYgGS:-M2EvYISDC7JJr7wYgGT:b9rz0mu) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/11fe1dfc1e702d70572031be8ba2d0919fa908d3/node/pkg/driver/node_utils.go#L113)):*
> ```Go
> 		for _, iqn := range iqns {
> 			if ips, iqnExists := publishContext[iqn]; iqnExists {
> 				ipsByArrayInitiator[iqn] = strings.Split(ips, ",")
> ```

create constant for initiators separaror "," and use it here

igorgonibm:
___
*[node/pkg/driver/node_utils.go, line 120 at r5](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M2EvgQR1l_s-KNyej7s:-M2EvgQR1l_s-KNyej7t:b-d8fwy5) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/11fe1dfc1e702d70572031be8ba2d0919fa908d3/node/pkg/driver/node_utils.go#L120)):*
> ```Go
> 	}
> 	if connectivityType == device_connectivity.ConnectionTypeFC {
> 		wwns := strings.Split(publishContext[configYaml.Controller.Publish_context_fc_initiators], ",")
> ```

use initiators separator constant here

igorgonibm:
___
*[node/pkg/driver/device_connectivity/device_connectivity_iscsi.go, line 47 at r5](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M2Ex2wKEiYJacspV5yN:-M2Ex2wKEiYJacspV5yO:b-spedqn) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/11fe1dfc1e702d70572031be8ba2d0919fa908d3/node/pkg/driver/device_connectivity/device_connectivity_iscsi.go#L47)):*
> ```Go
> }
> 
> func (r OsDeviceConnectivityIscsi) iscsiDiscoverAndLogin(targetName, portal string) error {
> ```

We need unitest for this method

oriyarde:
@igorgonibm @27149chen Thank you for your comments, they are very valuable.
If you don't have any more comments, please resolve your open conversations and approve.

node/pkg/driver/node.go
oriyarde: We currently have two options, for every request:
1. Login
or
2. Parse all sessions to avoid some (probably most) logins

Which do you prefer?

node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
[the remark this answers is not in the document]
oriyarde: 1. do you mean that by default, both iqns of the same svc is being represented by the same multipath device mapper (i.e. if one fails, it will continue to work)?
2. how is it different than never logging out? or never deleting the volume that cannot be acceded (due to login failure)?
3. why? the retry will just re-login it right after. and what if it was already logged it before the driver login?

igorgonibm:
___
*[node/pkg/driver/node.go, line 137 at r3](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M29sKr01MK-5XkbnPlB:-M2UKPMS-V12h6WH2Xw8:b2nign6) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/f254e9837494d67511250f80502916447023627e/node/pkg/driver/node.go#L137)):*
<details><summary><i>Previously, oriyarde wrote…</i></summary><blockquote>

So you think we should always continue to rescan even if all logins fail?
</blockquote></details>

yes

igorgonibm:
___
*[node/pkg/driver/device_connectivity/device_connectivity_iscsi.go, line 68 at r6](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/131#-M2ULaWf8vdqe5jWNhII:-M2ULaWf8vdqe5jWNhIJ:btown3t) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/0c9fd94a906cac745b8532faa20e1270d290a2c1/node/pkg/driver/device_connectivity/device_connectivity_iscsi.go#L68)):*
> ```Go
> 	var err error
> 	for targetName, portals := range portalsByTarget {
> 		for _, portal := range portals {
> ```

I didn't saw it before. 
Discover should be done only with one ip per iqn (loop util on e succeeds). 
Only login should be to all ips.

node/pkg/driver/node.go
27149chen: 2 for now. If we don't have a better way.

oriyarde: Done

node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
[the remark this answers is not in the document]
oriyarde: @27149chen can we resolve this? we need this code ASAP

node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
```diff
@@ -32,6 +43,113 @@ func NewOsDeviceConnectivityIscsi(executer executer.ExecuterInterface) OsDeviceC
...
+			if !loggedInPortalsByTarget[targetName][portal] {
+				portals := filteredPortalsByTarget[targetName]
+				filteredPortalsByTarget[targetName] = append(portals, portal)
+			}
+		}
+	}
+	return filteredPortalsByTarget, nil
+}
+
+func (r OsDeviceConnectivityIscsi) iscsiDiscoverAny(portals []string) bool {
```
27149chen: please explain or add comments for this function, it is confusing.
why error != nil and return true?

oriyarde: according to @igorgonibm, one portal discovery (per target name) is sufficient.
so this function tries to successfully discover any (one) portal.
if this confusing, could you specify what should I write to make it clearer?

27149chen: you fixed the bug

node/pkg/driver/device_connectivity/device_connectivity_iscsi.go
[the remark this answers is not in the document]
oriyarde: Thanks for the approval.

PR #136, opened 2020-03-20
implement xiv get iscsi targets

oriyarde:
add xiv implementation for iscsi login

---
This change is [Reviewable](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/136)

### DIFF
--- a/controller/array_action/array_mediator_xiv.py
+++ b/controller/array_action/array_mediator_xiv.py
@@ -270,10 +270,26 @@ class XIVArrayMediator(ArrayMediatorAbstract):
             else:
                 raise controller_errors.UnMappingError(vol_name, host_name, ex)
 
-    def get_iscsi_targets_by_iqn(self):
+    def _get_iscsi_targets(self):
+        ip_interfaces = self.client.cmd.ipinterface_list().as_list
+        iscsi_interfaces = (i for i in ip_interfaces if i.type == "iSCSI")
+        ips = []
+        for interface in iscsi_interfaces:
+            if interface.address:
+                ips.append(interface.address)
+            if interface.address6:
+                ipv6 = interface.address6.join('[]')
+                ips.append(ipv6)
+        return ips
+
+    def _get_array_iqn(self):
         config_get_list = self.client.cmd.config_get().as_list
-        array_iqn = [a for a in config_get_list if a["name"] == "iscsi_name"][0]["value"]
-        return {array_iqn: []}  # TODO: CSI-1166 replace [] with list of iscsi target IPs (ipinterface_list)
+        return next(c.value for c in config_get_list if c.name == "iscsi_name")
+
+    def get_iscsi_targets_by_iqn(self):
+        array_iqn = self._get_array_iqn()
+        iscsi_targets = self._get_iscsi_targets()
+        return {array_iqn: iscsi_targets}
 
     def get_array_fc_wwns(self, host_name):
         fc_wwns_objects = self.client.cmd.fc_port_list()
--- a/controller/array_action/array_mediator_xiv.py
+++ b/controller/array_action/array_mediator_xiv.py
@@ -271,7 +271,7 @@ class XIVArrayMediator(ArrayMediatorAbstract):
                 raise controller_errors.UnMappingError(vol_name, host_name, ex)
 
     def _get_iscsi_targets(self):
-        ip_interfaces = self.client.cmd.ipinterface_list().as_list
+        ip_interfaces = self.client.cmd.ipinterface_list()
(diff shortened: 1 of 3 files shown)


igorgonibm:
___
*[controller/array_action/array_mediator_xiv.py, line 280 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/136#-M316V0n3wCU-OrukpZe:-M316V0n3wCU-OrukpZf:b85nk32) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/f04466661cef82113d1ec3ca6bacf29683bf38c4/controller/array_action/array_mediator_xiv.py#L280)):*
> ```Python
>             if interface.address:
>                 ips.append(interface.address)
>             if interface.address6:
> ```

why such join is needed?

igorgonibm:
___
*[controller/array_action/array_mediator_xiv.py, line 280 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/136#-M316V0n3wCU-OrukpZe:-M31AsTZ4KkdGcjQifDL:b-lvzhjt) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/f04466661cef82113d1ec3ca6bacf29683bf38c4/controller/array_action/array_mediator_xiv.py#L280)):*
<details><summary><i>Previously, oriyarde wrote…</i></summary><blockquote>

same as in SVC, ipv6 should be in square brackets so when the csi node concatenate it to ":3260", the result is valid
</blockquote></details>

the concatination should be done in node

igorgonibm:
___
*[controller/array_action/array_mediator_xiv.py, line 280 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/136#-M316V0n3wCU-OrukpZe:-M31SzLJ2UqE3Trg66yV:b-tct0ti) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/f04466661cef82113d1ec3ca6bacf29683bf38c4/controller/array_action/array_mediator_xiv.py#L280)):*
<details><summary><i>Previously, oriyarde wrote…</i></summary><blockquote>

node doesn't know which type of IP it is, and I don't think it should
</blockquote></details>

Can't node understand what ip type it is by its structure? 123.4 olr 11:22:33:44:55:66?
Some options here: https://stackoverflow.com/questions/22751035/golang-distinguish-ipv4-ipv6

igorgonibm:
___
*[controller/array_action/array_mediator_xiv.py, line 280 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/136#-M316V0n3wCU-OrukpZe:-M31kV_rAWodvsuriPmq:b901o7k) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/f04466661cef82113d1ec3ca6bacf29683bf38c4/controller/array_action/array_mediator_xiv.py#L280)):*
<details><summary><i>Previously, oriyarde wrote…</i></summary><blockquote>

It just makes it complicated for the node, the controller has a guarantee of ip type, the node should not have to guess
</blockquote></details>

but this is ugly
for ip4 u return clean ip   but for ipv6 something weird  - ip in brackets.

igorgonibm:
___
*[controller/array_action/array_mediator_xiv.py, line 280 at r1](https://reviewable.io/reviews/ibm/ibm-block-csi-driver/136#-M316V0n3wCU-OrukpZe:-M32AG9u15zPCuFIvByJ:bdau5wi) ([raw file](https://github.com/ibm/ibm-block-csi-driver/blob/f04466661cef82113d1ec3ca6bacf29683bf38c4/controller/array_action/array_mediator_xiv.py#L280)):*
<details><summary><i>Previously, oriyarde wrote…</i></summary><blockquote>

It is not less ugly than adding the brackets later in the node (it is even more efficient, since it saves the parsing check), and it is a common representation of a literal ipv6, so I don't find it weird or unclean.

there is also an option to remove the `+":3260"` logic in the node, since it is already the default port in iscsi-utils, and will probably never change.
</blockquote></details>

you mean [1:2:3:4:56:6] is valid ipv6?

controller/array_action/array_mediator_xiv.py
```diff
@@ -270,10 +270,26 @@ def unmap_volume(self, volume_id, host_name):
             else:
                 raise controller_errors.UnMappingError(vol_name, host_name, ex)
 
-    def get_iscsi_targets_by_iqn(self):
+    def _get_iscsi_targets(self):
+        ip_interfaces = self.client.cmd.ipinterface_list().as_list
```
27149chen: I guess the response of the command is iterable, you don't need as_list here.

oriyarde: Done.

PR #138, opened 2020-03-23
CSI-1109_license

igorgonibm:


NOTICES
```diff
@@ -1175,23 +1168,8 @@ END OF yaml version 2.2.2 NOTICES AND INFORMATION
...
 
 
-=========================================================================
-Section 3 - Other IBM Open Source Project codes
-=========================================================================
-
-Use of other IBM Open Source Project codes in IBM Block
-Storage CSI Driver Version 1.0.0
-
-https://github.com/IBM/pyxcli (Apache v2)
```
oriyarde: what was changed in 1.1.0 that led to this removal?

oriyarde:
@igorgonibm there's an unresolved conversation here, this should be addressed if not before merge then at least after.
could you also delete this branch once finished here?

PR #140, opened 2020-03-29
convert lunid to int in get_volume_mappings

27149chen:
fix CSI-1302


oriyarde:
@27149chen could you open a ticket to add a UT to cover this bug?

PR #143, opened 2020-03-30
convert zlinux lunid to the format in by-path folder

27149chen:
convert zlinux lunid to the format in by-path folder like this:
ccw-0.0.a200-fc-0x5005076306005329-lun-0x4010401f00000000

### DIFF
--- a/node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
+++ b/node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
@@ -128,11 +128,12 @@ see the udev function format_lun_number
 https://github.com/systemd/systemd/blob/c4ae2704b7e921a0b05486a7b201be6770a04ea7/src/udev/udev-builtin-path_id.c#L58
 */
 func convertIntToScsilun(lunId int) string {
-    if lunId < 256:
-        return strconv.Itoa(lunId)
-    else:
+	if lunId < 256 {
+		return strconv.Itoa(lunId)
+	} else {
 		converted := (pretreated_lun >> 16 & 0xFFFF) | (pretreated_lun&0xFFFF)<<16
-        return fmt.Sprintf("0x%x00000000", converted)
+		return fmt.Sprintf("0x%x00000000", converted)
+	}
 }
 
 func (r OsDeviceConnectivityHelperScsiGeneric) GetMpathDevice(volumeId string, lunId int, arrayIdentifiers []string, connectivityType string) (string, error) {
--- a/node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
+++ b/node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
@@ -118,8 +118,25 @@ func (r OsDeviceConnectivityHelperScsiGeneric) RescanDevices(lunId int, arrayIde
 	return nil
 }
 
+/*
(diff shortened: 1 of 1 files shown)


oriyarde:
> Is this for all CPU architectures? or just Z?

Also, is this for all storage types? of just ds8k?

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
```diff
@@ -144,7 +162,7 @@ func (r OsDeviceConnectivityHelperScsiGeneric) GetMpathDevice(volumeId string, l
 
 	for _, arrayIdentifier := range arrayIdentifiers {
 		dp := strings.Join([]string{targetPath, connectivityType, arrayIdentifier, "lun", lunIdStr}, "-")
-		logger.Infof("GetMpathDevice: Get the mpath devices related to connectivityType=%s initiator=%s and lunID=%s : {%v}", connectivityType, arrayIdentifier, lunIdStr, dp)
+		logger.Infof("GetMpathDevice: Get the mpath devices related to connectivityType=%s initiator=%s and lunID=%d : {%v}", connectivityType, arrayIdentifier, lunId, dp)
```
oriyarde: I think the log should still include somewhere the value after conversion (lunIdStr)

27149chen: yes, last params (dp) has it

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
```diff
@@ -118,8 +118,26 @@ func (r OsDeviceConnectivityHelperScsiGeneric) RescanDevices(lunId int, arrayIde
...
 }
 
+/*
+There are two style lun number, one's decimal value is <256 and the other
+is full as 16 hex digit. According to T10 SAM, when decimal value is more
+than 256 and it is converted to the full 16 hex digit, it should be
+swapped and converted into hex.
+
+see the udev function format_lun_number
+https://github.com/systemd/systemd/blob/c4ae2704b7e921a0b05486a7b201be6770a04ea7/src/udev/udev-builtin-path_id.c#L58
```
igorgonibm: i don't like such links. They may change in the future
f you want to leave it better write in more generic way "github systemd project file udev-builtin-path_id.c"

27149chen:
It is not releated to arch or storage types.

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
igorgonibm: I would like both before and after. (rawLunID=%s lunID=%s)

oriyarde: ok

oriyarde:
@27149chen please add UT coverage for this bug to a ticket, same as:
https://github.com/IBM/ibm-block-csi-driver/pull/140#issuecomment-605635793

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
27149chen: It is a blob object, it won't change.

oriyarde:
@27149chen can we delete this branch?

PR #146, opened 2020-04-06
Story/csi 1304 fix ds8k host ports

igorgonibm:


controller/array_action/array_mediator_ds8k.py
```diff
@@ -353,19 +356,16 @@ def get_iscsi_targets_by_iqn(self):
...
-            else:
-                fc_ports = self.client.get_fcports()
-
-            wwpns = [p.wwpn for p in fc_ports if p.state == IOPORT_STATUS_ONLINE]
+            host = self.client.get_host(host_name)
+            wwpns = [port[LOGIN_PORT_WWPN] for port in host.login_ports if
+                     port[LOGIN_PORT_STATE] == LOGIN_PORT_STATE_ONLINE]
             logger.debug("Found wwpns: {}".format(wwpns))
             return wwpns
+        except exceptions.NotFound:
```
oriyarde, line 367: change `try` content to include only the `host = self.client.get_host(host_name)` line

controller/array_action/ds8k_rest_client.py
```diff
@@ -135,9 +135,6 @@ def get_fcports(self):
     def get_host_mappings(self, host_name):
         return self._client.get_mappings_by_host(host_name)
 
-    def get_ioports_by_host(self, host_name):
```
oriyarde, line 138: what about removing `get_fcports`?

controller/tests/array_action/ds8k/test_array_mediator_ds8k.py
```diff
@@ -1,6 +1,6 @@
 import unittest
 from munch import Munch
-from mock import patch, NonCallableMagicMock
+from mock import patch, NonCallableMagicMock, Mock
```
oriyarde, line 3: what is this for?

controller/array_action/array_mediator_ds8k.py
```diff
@@ -11,6 +11,9 @@
 from controller.array_action import config
 from controller.array_action.array_action_types import Volume
 
+LOGIN_PORT_WWPN = 'wwpn'
+LOGIN_PORT_STATE = 'state'
+LOGIN_PORT_STATE_ONLINE = 'online'
```
oriyarde, line 16: remove the replaced const `IOPORT_STATUS_ONLINE`

controller/array_action/array_mediator_ds8k.py
igorgonibm: I prefer this way
host is defined inside try
other methods are done similarly. See ds8k mediator method get_volume_mappings for example

controller/array_action/ds8k_rest_client.py
igorgonibm: removed

controller/tests/array_action/ds8k/test_array_mediator_ds8k.py
igorgonibm: removed
code scanning shold had warn i think......

controller/array_action/array_mediator_ds8k.py
oriyarde: `get_volume_mappings` has a loop in the try, so at least it avoids multiplying the try overhead.

if you don't want a big `try/except` block in the middle of the function, make a separate function for it. as written in the Clean Code book:
"if the keyword 'try' exists in a function, it should be the very first word in the function and that there should be nothing after the catch/finally blocks."

besides that, I don't see another reason to `try/except` irrelevant code lines.

controller/tests/array_action/ds8k/test_array_mediator_ds8k.py
oriyarde: CSI-1196

controller/array_action/array_mediator_ds8k.py
oriyarde: `IOPORT_STATUS_ONLINE` does not equal to `'state'`, it's equal to `'online'`, please remove it, even though it is not yours. thanks.

controller/array_action/array_mediator_ds8k.py
igorgonibm: " there should be nothing after the catch/finally blocks."
so where to put the rest of the  code? (i don't want separate func. this is overkill)

igorgonibm: see create_volume in xiv mediator. he same - command after xcli command.

oriyarde: I don't think it's an overkill, it's just another small function. small functions are good.
and I don't agree with how other functions did it. they are big and ugly.

igorgonibm: if others work this way i won'g change the style

controller/array_action/array_mediator_ds8k.py
```diff
@@ -11,6 +11,9 @@
 from controller.array_action import config
 from controller.array_action.array_action_types import Volume
 
+LOGIN_PORT_WWPN = 'wwpn'
```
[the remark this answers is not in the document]
oriyarde: then import them `from pyds8k.resources.ds8k.v1.common.attr_names import IOPORT_WWPN, IOPORT_STATUS`.
it is not our responsibility to define their value.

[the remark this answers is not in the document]
igorgonibm: What you specified relates to IPORTS object and i use login ports of HOST object
Do u want me top use them anyway?

controller/array_action/array_mediator_ds8k.py
igorgonibm: Please comment on the correct line next time.

oriyarde: I cannot commit outside the diff

controller/array_action/array_mediator_ds8k.py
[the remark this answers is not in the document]
oriyarde: then we can compromise on:
`LOGIN_PORT_WWPN = IOPORT_WWPN`
`LOGIN_PORT_STATE = IOPORT_STATUS`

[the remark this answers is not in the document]
igorgonibm: this is exactly what i intended to do 
done

controller/array_action/array_mediator_ds8k.py
igorgonibm: this i why i didn't push resolve before pushing

PR #149, opened 2020-04-08
CSI-1318_fix_ds8k_host_ports_node_fixed

igorgonibm:
Wain for mpaths in parallel using go  chanels

### DIFF
--- a/node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
+++ b/node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
@@ -167,23 +167,23 @@ func (r OsDeviceConnectivityHelperScsiGeneric) GetMpathDevice(volumeId string, l
 	}
 
 	logger.Debugf("GetMpathDevice: Start concurrent multipath devices search for volume : [%s]", volumeId)
-	ch := make(chan *WaitForMpathResult)
+	mpathResChannel := make(chan *WaitForMpathResult)
 	for _, arrayIdentifier := range arrayIdentifiers {
-		go r.waitForMpath(targetPath, connectivityType, arrayIdentifier, lunId, volumeId, ch)
+		go r.waitForMpath(targetPath, connectivityType, arrayIdentifier, lunId, volumeId, mpathResChannel)
 	}
 
 	for i := 1; i <= len(arrayIdentifiers); i++ {
-		mpathRes := <-ch
+		mpathRes := <-mpathResChannel
 		devicePaths = append(devicePaths, mpathRes.devicesPaths...)
 		if mpathRes.err != nil {
 			errStrings = append(errStrings, mpathRes.err.Error())
 		}
 	}
-	close(ch)
+	close(mpathResChannel)
 	logger.Debugf("GetMpathDevice: Finished concurrent multipath devices search for volume : [%s]", volumeId)
 
 	if len(devicePaths) == 0 && len(errStrings) != 0 {
-		err := errors.New(strings.Join(errStrings, ","))
+		err := errors.New(strings.Join(errStrings, GetMpahDevErrorsSep))
 		return "", err
 	}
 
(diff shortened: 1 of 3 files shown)


node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric_test.go
```diff
@@ -277,7 +281,7 @@ func TestGetMpathDevice(t *testing.T) {
 						t.Fatalf("Expected error type %v, got different error %v", tc.expErrType, reflect.TypeOf(err))
 					}
 				} else {
-					if err.Error() != tc.expErr.Error() {
+					if !areSringsEqualsAsSet(err.Error(), tc.expErr.Error()) {
 						t.Fatalf("Expected error code %s, got %s", tc.expErr, err.Error())
```
oriyarde, line 285: I don't think this is error "code" anymore

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric_test.go
```diff
@@ -220,7 +228,7 @@ func TestGetMpathDevice(t *testing.T) {
 
 		{
 			name:             "Should fail when WaitForPathToExist return error with the first array iqn, and found no sd device with the second array iqn",
-			arrayIdentifiers: []string{"X", "Y"},
+			arrayIdentifiers: []string{"X", "X"},
```
oriyarde, line 231: why are these necessary?

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
```diff
@@ -160,20 +166,21 @@ func (r OsDeviceConnectivityHelperScsiGeneric) GetMpathDevice(volumeId string, l
...
-			logger.Errorf("GetMpathDevice: No device found error : %v ", e.Error())
-			errStrings = append(errStrings, e.Error())
-		} else if !exists {
-			e := &MultipleDeviceNotFoundForLunError{volumeId, lunId, []string{arrayIdentifier}}
-			logger.Errorf(e.Error())
-			errStrings = append(errStrings, e.Error())
+		go r.waitForMpath(targetPath, connectivityType, arrayIdentifier, lunId, volumeId, ch)
+	}
+
+	for i := 1; i <= len(arrayIdentifiers); i++ {
```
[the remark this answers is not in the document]
oriyarde: I'm not sure I follow.
`i` is not really used *inside* the loop, even though it doesn't cause an error.
I agree that there are "no new variables" on left side of := when using  `for _, __ := range arrayIdentifiers`, but I still think it's better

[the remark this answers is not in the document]
igorgonibm: _,_ is better woulod be better but for go  it is compilation error
for with index is ok for it because i is compared and updated in loop line itself so it treats i is used.

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric_test.go
igorgonibm: X is identifier name. When error occures it contains X as part of error message
Before my change there was predefined order - first handle X, simulate result for for X, next same for Y.
Now it is not defined which identifier will be handled  so we simulate first

oriyarde:

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
```diff
@@ -50,14 +50,24 @@ type OsDeviceConnectivityHelperScsiGeneric struct {
...
-	DevPath             = "/dev"
-	ConnectionTypeISCSI = "iscsi"
-	ConnectionTypeFC    = "fc"
+	DevPath                     = "/dev"
+	ConnectionTypeISCSI         = "iscsi"
+	ConnectionTypeFC            = "fc"
+	WaitForMpathRetries         = 5
+	WaitForMpathWaitIntervalSec = 1
+	FC_HOST_SYSFS_PATH          = "/sys/class/fc_remote_ports/rport-*/port_name"
+	IscsiHostRexExPath          = "/sys/class/iscsi_host/host*/device/session*/iscsi_session/session*/targetname"
```
oriyarde, line 69: RegEx*

PR #151, opened 2020-04-22
Feature/csi 1088 compile on z systems

oriyarde:
align `Dockerfile-csi-controller.test` to `Dockerfile-csi-controller`


Dockerfile-csi-controller.test
```diff
@@ -30,11 +30,11 @@ RUN pip3 install -r /driver/controller/requirements-tests.txt
 
 COPY ./common /driver/common
 COPY ./controller /driver/controller
+USER root
```
igorgonibm, line 33: This is build stage right? So it is root on  our system?

oriyarde: yes, ubi default user is non-root, and we need root to chown the `/driver` dir to `appuser`

PR #152, opened 2020-04-22
Story/csi 750 snap deployment 1 17

igorgonibm:
Create snapshot


controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -2,19 +2,242 @@
...
+        self.request.secrets = {"username": "user", "password": "pass", "management_address": "mg"}
+        self.request.parameters = {}
+        self.capacity_bytes = 10
+        self.request.name = snap_name
+        self.request.source_volume_id = "A9000:12345678"
+
+    @patch("controller.array_action.array_connection_manager.ArrayConnectionManager.__enter__")
+    @patch("controller.array_action.array_connection_manager.ArrayConnectionManager.__exit__")
+    def test_create_snapshot_with_empty_name(self, a_enter, a_exit):
+        self._test_create_object_with_empty_name(a_enter)
```
oriyarde, line 103: I thought the order of patch is [reversed](https://docs.python.org/3/library/unittest.mock.html#quick-guide) to the parameters order. how does it work? is this a false positive?

controller/array_action/array_mediator_xiv.py
```diff
@@ -163,6 +183,41 @@ def delete_volume(self, volume_id):
...
 
+    def get_snapshot(self, snapshot_name):
+        logger.debug("Get snapshot : {}".format(snapshot_name))
+        try:
+            cli_snapshot = self.client.cmd.vol_list(vol=snapshot_name).as_single_element
+        except xcli_errors.IllegalNameForObjectError as ex:
+            logger.exception(ex)
+            raise controller_errors.IllegalObjectName(ex.status)
+        if not cli_snapshot:
+            return None
```
oriyarde, line 194: so why did you define `SnapshotNotFoundError`?

controller/array_action/array_mediator_xiv.py
```diff
@@ -163,6 +183,41 @@ def delete_volume(self, volume_id):
...
+        if not cli_snapshot.master_name:
+            raise controller_errors.SnapshotNotFoundVolumeWithSameNameExists(cli_snapshot.name, self.endpoint)
```
oriyarde, line 196: this error seems very specific (and lacks the word "error").
can we at least shorten it to `VolumeWithSameNameExistsError`?

controller/array_action/array_mediator_ds8k.py
```diff
@@ -99,6 +99,16 @@ def max_connections(self):
         # max for rest api is 128.
         return 50
 
+    @classproperty
+    def max_snapshot_name_length(self):
+        # TODO: CSI-1339
+        pass
```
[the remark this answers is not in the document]
igorgonibm: Dud it at first. It caused code scanning to fall.

controller/array_action/array_mediator_xiv.py
igorgonibm: VolumeWithSameName not clear enough. It could be understood that we search for volume and another volume with same name exists.
Will add word "Error"

controller/array_action/array_mediator_xiv.py
```diff
@@ -76,17 +84,26 @@ def disconnect(self):
         if self.client and self.client.is_connected():
             self.client.close()
 
-    def _convert_size_blocks_to_bytes(self, size_in_blocks):
-        return size_in_blocks * self.BLOCK_SIZE_IN_BYTES
+    def _get_volume_or_snapshot_size_in_bytes(self, volume_or_snapshot):
```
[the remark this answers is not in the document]
igorgonibm: I didn't change. I added another method which uses this one.
Do u have a better suggestion for name? We can't call it "object" since it is specific for volume and snapshot

controller/controller_server/config.py
```diff
@@ -9,17 +9,22 @@
 VOLUME_CAPABILITIES_FIELD_ACCESS_TYPE_MOUNT = 'mount'
 VOLUME_CAPABILITIES_FIELD_ACCESS_TYPE_BLOCK = 'block'
 
+SECRET_USERNAME_PARAMETER = "username"
```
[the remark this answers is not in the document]
igorgonibm: probably copy/paste by mistake

controller/controller_server/csi_controller_server.py
```diff
@@ -331,6 +387,10 @@ def ControllerGetCapabilities(self, request, context):
         res = csi_pb2.ControllerGetCapabilitiesResponse(
             capabilities=[csi_pb2.ControllerServiceCapability(
                 rpc=csi_pb2.ControllerServiceCapability.RPC(type=types.Value("CREATE_DELETE_VOLUME"))),
+                csi_pb2.ControllerServiceCapability(
+                    rpc=csi_pb2.ControllerServiceCapability.RPC(type=types.Value("CREATE_DELETE_SNAPSHOT"))),
+                csi_pb2.ControllerServiceCapability(
+                    rpc=csi_pb2.ControllerServiceCapability.RPC(type=types.Value("LIST_SNAPSHOTS"))),
```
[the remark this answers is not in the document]
igorgonibm: removed for now. Currently no plan to implement

controller/array_action/array_mediator_xiv.py
igorgonibm: I had it in my [rev snapshot branch. Need it for next ticket
Removed for now

controller/tests/controller_server/csi_controller_server_test.py
igorgonibm: It was copy paste from volume test.

controller/controller_server/messages.py
```diff
@@ -7,16 +7,16 @@
...
 secret_missing_message = 'secret is missing'
-capabilities_not_set_message = "capbilities were not set"
+capabilities_not_set_message = "capabilities were not set"
 unsupported_fs_type_message = "unsupported fs_type : {}"
 unsupported_volume_access_type_message = "unsupported volume access type"
 unsupported_access_mode_message = "unsupported access mode : {}"
-name_should_be_empty_message = 'name should not be empty'
-size_bigget_then_0_message = 'size should be bigger then 0'
+name_should_not_be_empty_message = 'name should not be empty'
+size_bigger_then_0_message = 'size should be bigger then 0'
```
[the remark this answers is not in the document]
igorgonibm: Scanned it again. Fixed

controller/array_action/array_mediator_ds8k.py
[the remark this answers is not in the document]
oriyarde: I see. python is a little weird about it.
so just do it for all the methods (not properties)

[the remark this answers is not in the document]
igorgonibm: done

controller/array_action/array_mediator_xiv.py
oriyarde: 1. is it so bad if we use `VolumeWithSameNameExistsError` for both snapshot and volume searches?
2. is this even relevant for other storage types?
3. what about something like `FoundVolumeInsteadOfSnapshotError`?

igorgonibm: 1. Yes. There are different errors.
2. No. only volume and snapshot is the same object (at least in a9k storage)
3. I prefer my way. It is logner but more informative and clear.

controller/array_action/array_mediator_xiv.py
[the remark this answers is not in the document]
oriyarde: not sure which method you added (looks like a rename in this diff).
I'm still not sure why can't just we keep the `_convert_size_blocks_to_bytes` with the `size_in_blocks` parameter.
is it so bad for the `_generate_*_response` functions to access `.capacity`?

controller/array_action/array_mediator_xiv.py
oriyarde: but will you use this error here in the "revert to snapshot" ticket?
if so, please add its number here.
if not, then do you think we should return None (and in all mediators)?

igorgonibm: Here no. In CreateVolume
yes. We should return none in snapshot not found in all mediators

controller/tests/controller_server/csi_controller_server_test.py
oriyarde: we shouldn't use their buggy patch code for new tests.

I think it will be easier for you to fix the volume tests before using them (because the fix should be to remove the unnecessary-yet-required parameter),
but if not, then at least write your patches in the correct order

controller/array_action/array_mediator_xiv.py
oriyarde: 1. then what about `VolumeWithSameNameExistsSnapshotError`/`SnapshotNameFoundIsVolumeError`?
2. if this is only relevant to a9k, we might want a more general error here (since these errors are not storage-specific)
3. this shortening is better than nothing, but I'm sure we can find something better!

controller/tests/controller_server/csi_controller_server_test.py
igorgonibm: I started to remove what seems to be unused and the tests failed.

controller/array_action/array_mediator_xiv.py
igorgonibm: Ok SnapshotNameBelongsToVolumeError

controller/tests/array_action/xiv/array_mediator_xiv_tests.py
```diff
@@ -111,6 +111,72 @@ def test_delete_volume_succeeds(self):
...
+
+    def test_create_snapshot_return_volume_does_not_exists_error(self):
+        self._test_create_snapshot_error(xcli_errors.VolumeBadNameError, array_errors.VolumeNotFoundError)
+
+    def test_create_snapshot_return_permission_error(self):
+        self._test_create_snapshot_error(xcli_errors.OperationForbiddenForUserCategoryError,
+                                         array_errors.PermissionDeniedError)
+
+    @patch.object(XIVArrayMediator, "_generate_snapshot_response")
+    def test_create_snapshot_generate_snapshot_response_return_exception(self, response):
```
oriyarde, line 164: raise_exception (not return).
I don't think it's supposed to raise an exception

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -2,19 +2,238 @@
...
+        self._test_create_object_with_wrong_secrets(a_enter)
+
+    @patch("controller.array_action.array_connection_manager.ArrayConnectionManager.detect_array_type")
+    @patch("controller.array_action.array_connection_manager.ArrayConnectionManager.__enter__")
+    def test_create_snapshot_with_array_connection_exception(self, a_enter, array_type):
+        self._test_create_object_with_array_connection_exception(a_enter)
+
+    @patch("controller.array_action.array_connection_manager.ArrayConnectionManager.detect_array_type")
+    @patch("controller.array_action.array_connection_manager.ArrayConnectionManager.__enter__")
+    @patch("controller.array_action.array_connection_manager.ArrayConnectionManager.__exit__")
```
oriyarde, line 133: please [fix patch/parameter order](https://github.com/IBM/ibm-block-csi-driver/pull/152/files/#r415303275) **everywhere** in your diff

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -2,19 +2,238 @@
...
+    def test_create_snapshot_with_get_array_type_exception(self, a_enter, a_exit, array_type):
+        self._test_create_object_with_get_array_type_exception(a_enter, array_type)
+
+    @patch("controller.array_action.array_connection_manager.ArrayConnectionManager.detect_array_type")
+    @patch("controller.array_action.array_mediator_xiv.XIVArrayMediator.get_snapshot")
+    @patch("controller.array_action.array_connection_manager.ArrayConnectionManager.__enter__")
+    def test_create_snapshot_get_snapshot_exception(self, a_enter, get_volume, array_type):
```
oriyarde, line 140: get_snapshot (not get_volume)

controller/tests/array_action/xiv/array_mediator_xiv_tests.py
igorgonibm: It may if capacity is nut a number. I took existing test for volume

controller/tests/controller_server/csi_controller_server_test.py
igorgonibm: Fixed order/removed unused where possible

controller/controller_server/csi_controller_server.py
```diff
@@ -318,6 +303,71 @@ def ListVolumes(self, request, context):
...
+                try:
+                    snapshot_name = self._get_snapshot_name(request, array_mediator)
+                except controller_errors.IllegalObjectName as ex:
+                    context.set_details(ex.message)
+                    context.set_code(grpc.StatusCode.INVALID_ARGUMENT)
+                    return csi_pb2.CreateSnapshotResponse()
+
+                volume_name = array_mediator.get_volume_name(vol_id)
+                logger.info("Snapshot name : {}. Volume name : {}".format(snapshot_name, volume_name))
+                snapshot = array_mediator.get_snapshot(snapshot_name)
```
27149chen, line 335: do you handle these errors?
https://github.com/container-storage-interface/spec/blob/master/spec.md#createsnapshot-errors
For example: ALREADY_EXISTS: Indicates that a snapshot corresponding to the specified snapshot name already exists but is incompatible with the specified volume_id.

controller/controller_server/csi_controller_server.py
```diff
@@ -318,6 +303,71 @@ def ListVolumes(self, request, context):
...
+                if not snapshot:
+                    logger.debug(
+                        "Snapshot doesn't exist. Creating a new snapshot {0} from volume {1}".format(snapshot_name,
+                                                                                                     volume_name))
+                    snapshot = array_mediator.create_snapshot(snapshot_name, volume_name)
```
27149chen, line 340: what if there are two create snapshot call? the first is faster so it creates the snapshot, should the second return error?

controller/controller_server/csi_controller_server.py
igorgonibm: How can it be icompatible? (at least a9k. This ticket is for a9k)
For creating volume from snapshot next ticket) i will check of course such issues.

controller/controller_server/csi_controller_server.py
igorgonibm: two calls for the same snapshot?
We have the same issue for create volume. In  such case slower create will fall,the retry will occur an d succeed.

controller/controller_server/csi_controller_server.py
igorgonibm: Will check if it is of other volume

controller/array_action/array_mediator_xiv.py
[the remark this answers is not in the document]
igorgonibm: left old method name

controller/controller_server/csi_controller_server.py
```diff
@@ -318,6 +303,79 @@ def ListVolumes(self, request, context):
...
+        except (controller_errors.IllegalObjectName, controller_errors.VolumeNotFoundError) as ex:
+            context.set_details(ex.message)
+            context.set_code(grpc.StatusCode.INVALID_ARGUMENT)
+            return csi_pb2.CreateSnapshotResponse()
+        except controller_errors.PermissionDeniedError as ex:
+            context.set_code(grpc.StatusCode.PERMISSION_DENIED)
+            context.set_details(ex)
+            return csi_pb2.CreateSnapshotResponse()
+        except (controller_errors.SnapshotAlreadyExists,
+                controller_errors.SnapshotNameBelongsToVolumeError) as ex:
```
oriyarde, line 363: this is not a snapshot, so I'm 100% sure we can't use ALREADY_EXISTS for this.
[the spec is very clear about it](https://github.com/container-storage-interface/spec/blob/v1.2.0/spec.md#createsnapshot-errors)

igorgonibm: will return INTERNAL now

controller/controller_server/csi_controller_server.py
igorgonibm: Ori created ticket to handle this it it will be decided to fix.  CSI-1380.
I don't add TODO unless we'll decide to handle it this way

controller/array_action/array_mediator_interface.py
```diff
@@ -144,13 +157,45 @@ def unmap_volume(self, volume_id, host_name):
...
+    @abstractmethod
+    def get_snapshot(self, snapshot_name):
+        """
+        This function return snapshot info about the snapshot.
+        Args:
+            snapshot_name : name of the snapshot in the storage system
+        Returns:
+           Snapshot
+        Raises:
+            SnapshotNotFoundVolumeWithSameNameExists
```
oriyarde, line 176: don't forget to update the docstrings on renames

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -2,19 +2,254 @@
...
+        self.assertEqual(context.code, grpc.StatusCode.OK)
+        self.mediator.create_snapshot.assert_called_once_with("prefix_some_name", "snap_vol")
+
+
+class TestControllerServerCreateVolume(AbstractControllerTest):
+
+    def get_create_object_method(self):
+        return self.servicer.CreateVolume
+
+    def get_create_object_method_response(self):
```
oriyarde, line 251: `get_create_object_response_method`?

controller/array_action/array_mediator_ds8k.py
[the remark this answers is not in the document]
oriyarde: you still [haven't changed it](https://github.com/IBM/ibm-block-csi-driver/pull/152/files/d665b590e05109e4780f6d732e606f2fa99b7875#diff-249ba19401c4a58c36225e8fa741fe02R165) everywhere

oriyarde:
@igorgonibm what about adding a pvc yaml example with `dataSource`?

PR #153, opened 2020-04-26
Feature/csi 1319 create z build jobs in jenkins

ArbelNathan:
CSI-1319

### DIFF
--- a/scripts/run_yamlcheck.sh
+++ b/scripts/run_yamlcheck.sh
@@ -1,7 +1,7 @@
 #!/bin/bash -xe
 
 arch=$(uname -m)
-if [ "$arch" = "x86_64" ];
+if [[ "$arch" == "x86"* ]];
 then
 docker run -t -v `pwd`/deploy/kubernetes/v1.13:/deploy/kubernetes/v1.13 garethr/kubeval deploy/kubernetes/v1.13/*.yaml
 docker run -t -v `pwd`/deploy/kubernetes/v1.14:/deploy/kubernetes/v1.14 garethr/kubeval deploy/kubernetes/v1.14/*.yaml
--- a/scripts/run_yamlcheck.sh
+++ b/scripts/run_yamlcheck.sh
@@ -1,6 +1,6 @@
 #!/bin/bash -xe
 
-arch=$('/bin/uname -m')
+arch=$('/usr/bin/uname -m')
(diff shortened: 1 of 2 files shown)


scripts/run_yamlcheck.sh
```diff
@@ -1,4 +1,9 @@
 #!/bin/bash -xe
+
+arch=$(uname -m)
+echo arch
```
oriyarde, line 4: remove the echo line

scripts/run_yamlcheck.sh
```diff
@@ -1,4 +1,9 @@
...
+if [ "$arch" = "x86_64" ];
+then
 docker run -t -v `pwd`/deploy/kubernetes/v1.13:/deploy/kubernetes/v1.13 garethr/kubeval deploy/kubernetes/v1.13/*.yaml
 docker run -t -v `pwd`/deploy/kubernetes/v1.14:/deploy/kubernetes/v1.14 garethr/kubeval deploy/kubernetes/v1.14/*.yaml
-
+fi
```
oriyarde, line 9: add `else` echo "skipping k8s yaml validations since it only supports x86" (something like this)

ArbelNathan: done

scripts/run_yamlcheck.sh
ArbelNathan: done

scripts/run_yamlcheck.sh
```diff
@@ -1,4 +1,10 @@
 #!/bin/bash -xe
+
+arch=$(uname -m)
+if [ "$arch" = "x86_64" ];
```
igorgonibm, line 4: shoulodn't we skip also for x86/x86_32?

ArbelNathan: do we have slaves that work on x86/x86_32?

ArbelNathan: do we have any slaves that work on x86/x86_32?

ArbelNathan: change it to be on the safe side

oriyarde: I think they only support 64:
https://hub.docker.com/r/garethr/kubeval/tags

PR #154, opened 2020-05-06
Story/csi 1276 svc perf get host (reconsider open issues CSI-1492 as marked in conversations)

igorgonibm:
Improve performance of getting host by identifiers for SVC

### DIFF
--- a/controller/array_action/array_mediator_svc.py
+++ b/controller/array_action/array_mediator_svc.py
@@ -1,4 +1,5 @@
 from collections import defaultdict
+from io import StringIO
 
 from pysvc import errors as svc_errors
 from pysvc.unified.client import connect
@@ -8,11 +9,9 @@ import controller.array_action.config as config
 import controller.array_action.errors as controller_errors
 from controller.array_action.array_action_types import Volume, Host
 from controller.array_action.array_mediator_abstract import ArrayMediatorAbstract
+from controller.array_action.svc_cli_result_reader import SVCListResultsReader
 from controller.array_action.utils import classproperty, bytes_to_string
 from controller.common.csi_logger import get_stdout_logger
-from controller.array_action.svc_cli_result_reader import SVCListResultsReader
-
-from io import StringIO
 
 array_connections_dict = {}
 logger = get_stdout_logger()
--- a/controller/array_action/array_mediator_svc.py
+++ b/controller/array_action/array_mediator_svc.py
@@ -271,6 +271,11 @@ class SVCArrayMediator(ArrayMediatorAbstract):
     def get_host_by_host_identifiers(self, initiators):
         logger.debug("Getting host id for initiators : {0}".format(initiators))
         hosts_list = self.client.svcinfo.lshost()
+        if not hosts_list:
+            logger.debug(
+                "Can not found host by using initiators: {0}. No hosts defined on the array {1} ".format(initiators,
+                                                                                                         self.endpoint))
+            raise controller_errors.HostNotFoundError(initiators)
 
         # for each host get its detailoed info from array by sending batch of commands
         detailed_hosts_list_cmd = self._get_detailed_hosts_list_cmd(hosts_list)
--- a/controller/array_action/array_mediator_svc.py
+++ b/controller/array_action/array_mediator_svc.py
@@ -277,11 +277,7 @@ class SVCArrayMediator(ArrayMediatorAbstract):
         logger.debug("Getting detailed hosts list")
         # TODO
         logger.debug(hosts_list)
-        logger.debug(detailed_hosts_list_cmd)
-        logger.debug(type(self.client.send_raw_command))
         detailed_hosts_list_output, detailed_hosts_list_errors = self.client.send_raw_command(detailed_hosts_list_cmd)
-        # TODO
-        logger.debug("done")
         detailed_hosts_list_string = bytes_to_string(detailed_hosts_list_output)
         detailed_hosts_list_error_msg = self._get_formatted_hosts_list_error_msg(detailed_hosts_list_errors)
         if not detailed_hosts_list_errors:
@@ -289,6 +285,9 @@ class SVCArrayMediator(ArrayMediatorAbstract):
 
         logger.debug("Finding the correct host")
         hosts_reader = SVCListResultsReader(detailed_hosts_list_string)
+        # TODO
+        logger.debug(hosts_reader)
+        logger.debug(type(hosts_reader))
         iscsi_host, fc_host = None, None
         for host_details in hosts_reader:
             host_name = host_details.get_as_list(HOST_NAME_PARAM)
--- a/controller/array_action/array_mediator_svc.py
+++ b/controller/array_action/array_mediator_svc.py
@@ -33,7 +33,7 @@ HOST_ID_PARAM = 'id'
 HOST_NAME_PARAM = 'name'
 HOST_ISCSI_NAMES_PARAM = 'iscsi_name'
 HOST_WWPNS_PARAM = 'WWPN'
-MAX_HOSTS_LIST_ERROR_LOG_LENGTH = 100
+MAX_HOSTS_LIST_ERR_MSG_LENGTH = 300
 
 def is_warning_message(ex):
     """ Return True if the exception message is warning """
@@ -321,9 +321,9 @@ class SVCArrayMediator(ArrayMediatorAbstract):
         detailed_host_list_errors = bytes_to_string(detailed_host_list_errors_bytes)
         if not detailed_host_list_errors:
             return ""
-        elif len(detailed_host_list_errors) <= MAX_HOSTS_LIST_ERROR_LOG_LENGTH
+        elif len(detailed_host_list_errors) <= MAX_HOSTS_LIST_ERR_MSG_LENGTH:
             return detailed_host_list_errors
-        return "{0} ...".format(detailed_host_list_errors[:MAX_HOSTS_LIST_ERROR_LOG_LENGTH])
+        return "{0} ...".format(detailed_host_list_errors[:MAX_HOSTS_LIST_ERR_MSG_LENGTH])
 
     def get_volume_mappings(self, volume_id):
         logger.debug("Getting volume mappings for volume id : "
--- a/controller/array_action/array_mediator_svc.py
+++ b/controller/array_action/array_mediator_svc.py
@@ -336,7 +336,8 @@ class SVCArrayMediator(ArrayMediatorAbstract):
         output_as_bytes, errors_as_bytes = self.client.send_raw_command(cmd)
         output_as_str = bytes_to_string(output_as_bytes)
         errors_as_str = bytes_to_string(errors_as_bytes)
-        return output_as_str, errors_as_str
+        formatted_errors_as_str = self._get_formatted_hosts_list_error_msg(errors_as_str)
+        return output_as_str, formatted_errors_as_str
 
     def _get_formatted_hosts_list_error_msg(self, detailed_host_list_errors):
         if len(detailed_host_list_errors) <= HOSTS_LIST_ERR_MSG_MAX_LENGTH:
--- a/controller/array_action/array_mediator_svc.py
+++ b/controller/array_action/array_mediator_svc.py
@@ -27,8 +27,7 @@ VOL_NOT_FOUND = 'CMMVC8957E'
 POOL_NOT_MATCH_VOL_CAPABILITIES = 'CMMVC9292E'
 NOT_REDUCTION_POOL = 'CMMVC9301E'
 
-LIST_HOSTS_CMD = "lshost"
-LIST_CMDS_SEPARATOR = ";"
+LIST_HOSTS_CMD_FORMAT = 'lshost {HOST_ID};'
 HOST_ID_PARAM = 'id'
 HOST_NAME_PARAM = 'name'
 HOST_ISCSI_NAMES_PARAM = 'iscsi_name'
@@ -329,20 +328,17 @@ class SVCArrayMediator(ArrayMediatorAbstract):
         writer = StringIO()
         for host in host_list:
             host_id = host.get(HOST_ID_PARAM)
-            writer.write(LIST_HOSTS_CMD)
-            writer.write(" ")
-            writer.write(host_id)
-            writer.write(LIST_CMDS_SEPARATOR)
+            writer.write(LIST_HOSTS_CMD_FORMAT.format(HOST_ID=host_id))
         return writer.getvalue()
 
     def _send_raw_cli_command(self, cmd):
         output_as_bytes, errors_as_bytes = self.client.send_raw_command(cmd)
         output_as_str = bytes_to_string(output_as_bytes)
         errors_as_str = bytes_to_string(errors_as_bytes)
-        formatted_errors_as_str = self._get_formatted_hosts_list_error_msg(errors_as_str)
+        formatted_errors_as_str = self._truncate_error_msg(errors_as_str)
         return output_as_str, formatted_errors_as_str
 
-    def _get_formatted_hosts_list_error_msg(self, detailed_host_list_errors):
+    def _truncate_error_msg(self, detailed_host_list_errors):
         if len(detailed_host_list_errors) <= HOSTS_LIST_ERR_MSG_MAX_LENGTH:
             return detailed_host_list_errors
         return "{0} ...".format(detailed_host_list_errors[HOSTS_LIST_ERR_MSG_MAX_LENGTH])
--- a/controller/array_action/array_mediator_svc.py
+++ b/controller/array_action/array_mediator_svc.py
@@ -310,7 +310,7 @@ class SVCArrayMediator(ArrayMediatorAbstract):
         if detailed_hosts_list_errors:
             logger.error("Errors returned from getting detailed hosts list: {0}".format(detailed_hosts_list_errors))
 
-        return self._get_detailed_hosts_by_raw_output*detailed_hosts_list_output
+        return self._get_detailed_hosts_by_raw_output(detailed_hosts_list_output)
 
     def _get_detailed_hosts_by_raw_output(self, detailed_hosts_list_raw_output):
         logger.debug("Reading detailed hosts list commands batch response")
--- a/controller/array_action/array_mediator_svc.py
+++ b/controller/array_action/array_mediator_svc.py
@@ -13,7 +13,6 @@ from controller.common.csi_logger import get_stdout_logger
 from controller.array_action.svc_cli_result_reader import SVCListResultsReader
 
 from io import StringIO
-import textwrap
 
 array_connections_dict = {}
 logger = get_stdout_logger()
@@ -34,7 +33,7 @@ HOST_ID_PARAM = 'id'
 HOST_NAME_PARAM = 'name'
 HOST_ISCSI_NAMES_PARAM = 'iscsi_name'
 HOST_WWPNS_PARAM = 'WWPN'
-MAX_HOSTS_LIST_ERR_MSG_LENGTH = 300
+HOSTS_LIST_ERR_MSG_MAX_LENGTH = 300
 
 
 def is_warning_message(ex):
@@ -307,14 +306,12 @@ class SVCArrayMediator(ArrayMediatorAbstract):
         # get all hosts details by sending a single batch of commands, in which each command is per host
         detailed_hosts_list_cmd = self._get_detailed_hosts_list_cmd(hosts_list)
         logger.debug("Sending getting detailed hosts list commands batch")
-        detailed_hosts_list_output, detailed_hosts_list_errors = self.client.send_raw_command(detailed_hosts_list_cmd)
-        detailed_hosts_list_as_string = bytes_to_string(detailed_hosts_list_output)
-        detailed_hosts_list_error_msg = self._get_formatted_hosts_list_error_msg(detailed_hosts_list_errors)
+        detailed_hosts_list_output, detailed_hosts_list_errors = send_raw_cli_command(detailed_hosts_list_cmd)
         if not detailed_hosts_list_errors:
-            logger.error("Errors returned from getting detailed hosts list: {0}".format(detailed_hosts_list_error_msg))
+            logger.error("Errors returned from getting detailed hosts list: {0}".format(detailed_hosts_list_errors))
 
         logger.debug("Reading detailed hosts list commands batch response")
-        hosts_reader = SVCListResultsReader(detailed_hosts_list_as_string)
+        hosts_reader = SVCListResultsReader(detailed_hosts_list_output)
         res = []
         for host_details in hosts_reader:
             host_id = host_details.get(HOST_ID_PARAM)
@@ -335,9 +332,16 @@ class SVCArrayMediator(ArrayMediatorAbstract):
             writer.write(LIST_CMDS_SEPARATOR)
         return writer.getvalue()
 
-    def _get_formatted_hosts_list_error_msg(self, detailed_host_list_errors_bytes):
-        detailed_host_list_errors = bytes_to_string(detailed_host_list_errors_bytes)
-        return textwrap.shorten(detailed_host_list_errors, width=MAX_HOSTS_LIST_ERR_MSG_LENGTH, placeholder="...")
+    def send_raw_cli_command(self, cmd):
+        output_as_bytes, errors_as_bytes = self.client.send_raw_command(cmd)
+        output_as_str = bytes_to_string(output_as_bytes)
+        errors_as_str = bytes_to_string(errors_as_bytes)
+        return output_as_str, errors_as_str
+
+    def _get_formatted_hosts_list_error_msg(self, detailed_host_list_errors):
+        if len(detailed_host_list_errors) <= HOSTS_LIST_ERR_MSG_MAX_LENGTH:
+            return detailed_host_list_errors
+        return "{0} ...".format(detailed_host_list_errors[HOSTS_LIST_ERR_MSG_MAX_LENGTH])
 
     def get_volume_mappings(self, volume_id):
         logger.debug("Getting volume mappings for volume id : "
--- a/controller/array_action/array_mediator_svc.py
+++ b/controller/array_action/array_mediator_svc.py
@@ -33,7 +33,7 @@ HOST_ID_PARAM = 'id'
 HOST_NAME_PARAM = 'name'
 HOST_ISCSI_NAMES_PARAM = 'iscsi_name'
 HOST_WWPNS_PARAM = 'WWPN'
-
+MAX_HOSTS_LIST_ERROR_LOG_LENGTH = 100
 
 def is_warning_message(ex):
     """ Return True if the exception message is warning """
@@ -273,9 +273,9 @@ class SVCArrayMediator(ArrayMediatorAbstract):
         detailed_hosts_list_cmd = self._get_detailed_hosts_list_cmd(hosts_list)
         logger.debug("Getting detailed hosts list")
         detailed_host_list_output, detailed_host_list_errors = self.client.send_raw_command(detailed_hosts_list_cmd)
-        detailed_host_list_errors = bytes_to_string(detailed_host_list_errors)
+        detailed_host_list_error_msg = self._get_formatted_hosts_list_error_msg(detailed_host_list_errors)
         if not detailed_host_list_errors:
-            logger.error("Errors returned from getting detailed hosts list: {0}".format(detailed_host_list_errors))
+            logger.error("Errors returned from getting detailed hosts list: {0}".format(detailed_host_list_error_msg))
         logger.debug("Finding the correct host")
         hosts_reader = SVCListResultsReader(detailed_host_list_output)
         iscsi_host, fc_host = None, None
@@ -316,6 +316,14 @@ class SVCArrayMediator(ArrayMediatorAbstract):
             writer.write(host_id)
             writer.write(LIST_CMDS_SEPARATOR)
         return writer.getvalue()
+
+    def _get_formatted_hosts_list_error_msg(self, detailed_host_list_errors_bytes):
+        detailed_host_list_errors = bytes_to_string(detailed_host_list_errors_bytes)
+        if not detailed_host_list_errors:
+            return ""
+        elif len(detailed_host_list_errors) <= MAX_HOSTS_LIST_ERROR_LOG_LENGTH
+            return detailed_host_list_errors
+        return "{0} ...".format(detailed_host_list_errors[:MAX_HOSTS_LIST_ERROR_LOG_LENGTH])
 
     def get_volume_mappings(self, volume_id):
         logger.debug("Getting volume mappings for volume id : "
--- a/controller/array_action/array_mediator_svc.py
+++ b/controller/array_action/array_mediator_svc.py
@@ -307,7 +307,7 @@ class SVCArrayMediator(ArrayMediatorAbstract):
         detailed_hosts_list_cmd = self._get_detailed_hosts_list_cmd(hosts_list)
         logger.debug("Sending getting detailed hosts list commands batch")
         detailed_hosts_list_output, detailed_hosts_list_errors = self._send_raw_cli_command(detailed_hosts_list_cmd)
-        if not detailed_hosts_list_errors:
+        if detailed_hosts_list_errors:
             logger.error("Errors returned from getting detailed hosts list: {0}".format(detailed_hosts_list_errors))
 
         logger.debug("Reading detailed hosts list commands batch response")
--- a/controller/array_action/array_mediator_svc.py
+++ b/controller/array_action/array_mediator_svc.py
@@ -27,8 +27,8 @@ VOL_NOT_FOUND = 'CMMVC8957E'
 POOL_NOT_MATCH_VOL_CAPABILITIES = 'CMMVC9292E'
 NOT_REDUCTION_POOL = 'CMMVC9301E'
 
-LIST_HOSTS_CMD = "lshost "
-LIST_CMDS_SEPARATOR = "; "
+LIST_HOSTS_CMD = "lshost"
+LIST_CMDS_SEPARATOR = ";"
 HOST_ID_PARAM = 'id'
 HOST_NAME_PARAM = 'name'
 HOST_ISCSI_NAMES_PARAM = 'iscsi_name'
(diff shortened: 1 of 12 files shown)


controller/array_action/array_mediator_svc.py
```diff
@@ -294,6 +297,51 @@ def get_host_by_host_identifiers(self, initiators):
...
+            wwns = host_details.get_as_list(HOST_WWPNS_PARAM)
+            host = Host(host_id, host_name, iscsi_names, wwns)
+            res.append(host)
+        return res
+
+    def _get_detailed_hosts_list_cmd(self, host_list):
+        writer = StringIO()
+        for host in host_list:
+            host_id = host.get(HOST_ID_PARAM)
+            writer.write(LIST_HOSTS_CMD)
```
oriyarde, line 331: I think something like
```
LIST_HOSTS_CMD_FORMAT = 'lshost {HOST_ID};'
writer.write(LIST_HOSTS_CMD_FORMAT.format(HOST_ID=host_id)
```
would be more readable and with the same "big O notation" efficiency

controller/array_action/array_mediator_svc.py
```diff
@@ -294,6 +297,51 @@ def get_host_by_host_identifiers(self, initiators):
             logger.debug("can not found host by using initiators: {0} ".format(initiators))
             raise controller_errors.HostNotFoundError(initiators)
 
+    def _get_detailed_hosts_list(self):
+        logger.debug("Getting detailed hosts list on array {0}".format(self.endpoint))
+        hosts_list = self.client.svcinfo.lshost()
+        if not hosts_list:
+            return []
+
+        # for each host get its detailed info from array by sending batch of commands
```
oriyarde, line 306: this might imply that we're sending a batch "for each host".
maybe:
`get all hosts details by sending a single batch of commands, in which each command is per host`?

controller/array_action/array_mediator_svc.py
```diff
@@ -24,6 +27,14 @@
 POOL_NOT_MATCH_VOL_CAPABILITIES = 'CMMVC9292E'
 NOT_REDUCTION_POOL = 'CMMVC9301E'
 
+LIST_HOSTS_CMD = "lshost"
+LIST_CMDS_SEPARATOR = ";"
+HOST_ID_PARAM = 'id'
+HOST_NAME_PARAM = 'name'
+HOST_ISCSI_NAMES_PARAM = 'iscsi_name'
+HOST_WWPNS_PARAM = 'WWPN'
+MAX_HOSTS_LIST_ERR_MSG_LENGTH = 300
```
oriyarde, line 36: `HOSTS_LIST_ERR_MSG_MAX_LENGTH`?

controller/array_action/array_mediator_svc.py
```diff
@@ -294,6 +297,51 @@ def get_host_by_host_identifiers(self, initiators):
...
+    def _get_detailed_hosts_list(self):
+        logger.debug("Getting detailed hosts list on array {0}".format(self.endpoint))
+        hosts_list = self.client.svcinfo.lshost()
+        if not hosts_list:
+            return []
+
+        # for each host get its detailed info from array by sending batch of commands
+        detailed_hosts_list_cmd = self._get_detailed_hosts_list_cmd(hosts_list)
+        logger.debug("Sending getting detailed hosts list commands batch")
+        detailed_hosts_list_output, detailed_hosts_list_errors = self.client.send_raw_command(detailed_hosts_list_cmd)
```
oriyarde, line 309: `send_raw_command` needs a wrapper that would convert the bytes to string for both stdout and stderr

controller/array_action/array_mediator_svc.py
```diff
@@ -294,6 +297,51 @@ def get_host_by_host_identifiers(self, initiators):
...
+            host_id = host.get(HOST_ID_PARAM)
+            writer.write(LIST_HOSTS_CMD)
+            writer.write(" ")
+            writer.write(host_id)
+            writer.write(LIST_CMDS_SEPARATOR)
+        return writer.getvalue()
+
+    def _get_formatted_hosts_list_error_msg(self, detailed_host_list_errors_bytes):
+        detailed_host_list_errors = bytes_to_string(detailed_host_list_errors_bytes)
+        if not detailed_host_list_errors:
```
oriyarde, line 339: no need for this `if`

controller/array_action/svc_cli_result_reader.py
```diff
@@ -0,0 +1,117 @@
...
+import controller.array_action.errors as controller_errors
+
+logger = get_stdout_logger()
+
+ID_PARAM_NAME = "id"
+
+
+class SVCListResultsReader:
+    """
+    Iterable object used to read raw command response from m SVS array.
```
roysahar-ibm, line 11: typo - "from m and SVS" should be "from SVC"

controller/array_action/array_mediator_svc.py
```diff
@@ -294,6 +297,51 @@ def get_host_by_host_identifiers(self, initiators):
...
+            return ""
+        elif len(detailed_host_list_errors) <= MAX_HOSTS_LIST_ERR_MSG_LENGTH:
```
oriyarde, line 341: this can be shortened to:
`return textwrap.shorten(errors, width=MAX, placeholder="...")`

controller/array_action/svc_cli_result_reader.py
```diff
@@ -0,0 +1,117 @@
...
+    Each object in response is translated to SVCListResultsElement.
+    Input is received as string which represents '\n'-separated list of returned lines from output.
+    Line with param 'id' (e.g. 'id 1') is recognized as first line of object ans used as separator between objects
+    (e.g. in input "id 1\nname n3\nid 2\nname n2" first object starts with line 'id 1' and ends with 'id 2')
```
roysahar-ibm, line 15: It is hard to understand the comment with the "\nn"

controller/array_action/array_mediator_svc.py
```diff
@@ -294,6 +297,51 @@ def get_host_by_host_identifiers(self, initiators):
...
+        if not hosts_list:
+            return []
+
+        # for each host get its detailed info from array by sending batch of commands
+        detailed_hosts_list_cmd = self._get_detailed_hosts_list_cmd(hosts_list)
+        logger.debug("Sending getting detailed hosts list commands batch")
+        detailed_hosts_list_output, detailed_hosts_list_errors = self.client.send_raw_command(detailed_hosts_list_cmd)
+        detailed_hosts_list_as_string = bytes_to_string(detailed_hosts_list_output)
+        detailed_hosts_list_error_msg = self._get_formatted_hosts_list_error_msg(detailed_hosts_list_errors)
+        if not detailed_hosts_list_errors:
```
oriyarde, line 312: why `not`? perhaps we need a UT for this...

controller/array_action/svc_cli_result_reader.py
```diff
@@ -0,0 +1,117 @@
...
+            if param_name == ID_PARAM_NAME:
+                self._next_object_id = param_value
+                return res
+            res.add(param_name, param_value)
+        return res
+
+    def _parse_param(self, line):
+        """
+        Args:
+            line : lin e representing <param-name> <param-value> (para value may be empty or contains whitespaces)
```
roysahar-ibm, line 75: typo - "lin e" should be "line"
"(para " should be "(param"

controller/array_action/svc_cli_result_reader.py
igorgonibm: because it is \nname or all the comment?
If i'll use other example it ill be ok? E.g. \nwwns

controller/array_action/svc_cli_result_reader.py
igorgonibm: fixed

controller/array_action/svc_cli_result_reader.py
igorgonibm: fixed

controller/array_action/utils.py
```diff
@@ -1,3 +1,10 @@
+from controller.array_action.config import UTF_8
+
+
+def bytes_to_string(input_as_bytes):
+    return input_as_bytes.decode(UTF_8) if input_as_bytes else ""
```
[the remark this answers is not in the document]
igorgonibm: I added to handle None value which probably may not be. But when I use this method I assume it never raises exception
This is called "defensive programming"

controller/tests/array_action/svc/array_mediator_svc_test.py
```diff
@@ -1,15 +1,18 @@
...
+    HOST_ID_PARAM, HOST_NAME_PARAM, HOST_ISCSI_NAMES_PARAM, HOST_WWPNS_PARAM
+from controller.array_action.svc_cli_result_reader import SVCListResultsElement
+from controller.array_action.config import UTF_8
 import controller.array_action.errors as array_errors
 from pysvc.unified.response import CLIFailureError
 from pysvc import errors as svc_errors
 import controller.array_action.config as config
 from controller.common.node_info import Initiators
 
+EMPTY_BYTES = bytes("", UTF_8)
```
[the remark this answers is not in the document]
igorgonibm: done

controller/array_action/array_mediator_svc.py
```diff
@@ -294,6 +297,51 @@ def get_host_by_host_identifiers(self, initiators):
             logger.debug("can not found host by using initiators: {0} ".format(initiators))
             raise controller_errors.HostNotFoundError(initiators)
 
+    def _get_detailed_hosts_list(self):
```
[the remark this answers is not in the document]
oriyarde: 1. no need for this general solution, since the py-cli libs don't (and has no plan to) support loading into memory in chunks.
2. the function is still too long (regardless of the general solution), and should still be divided into two.
3. as I wrote, Munch *is* different (avoids consts, avoids the wrongly-public Host class, conforms to the project's current convention, etc...).

controller/array_action/array_mediator_svc.py
oriyarde: if my suggested name is better

controller/array_action/array_mediator_svc.py
oriyarde: then let me rephrase: why does `if not errors` mean that "there are errors"?

controller/array_action/array_mediator_svc.py
oriyarde: it is still linear to the amount of hosts (and I assume "thousands" is only a theoretical case).
so how do you know that 1 short string construction requires more instructions than 4 "append"s?
did you compare the performance between the two?

controller/array_action/utils.py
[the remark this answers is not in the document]
oriyarde: I understand that this could protect against future `None`.
but as I wrote, since it is only used in SVC (and only in one place), I think it's too soon to put it in common utils

controller/array_action/array_mediator_svc.py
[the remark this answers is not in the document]
igorgonibm: think the Reader is correct. It is not open for discussion.
Host object is generic. It doesn't have anything svc related and there is a ticket to generalize get_hosts_for_init where it will be used.

controller/array_action/utils.py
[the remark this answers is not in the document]
igorgonibm: I do not agree. Leaving this way

controller/array_action/array_mediator_svc.py
igorgonibm: It doubles the time. It is not theoretical. We should support 4000 hosts.. 
Format passes over all characters and then  write passes again.

controller/array_action/array_mediator_svc.py
igorgonibm: ok

controller/array_action/array_mediator_svc.py
igorgonibm: ok

controller/array_action/array_mediator_svc.py
igorgonibm: We don't have UT to check logs

controller/array_action/array_mediator_svc.py
igorgonibm: seem s it doesn't work
>>textwrap.shorten('abcdefgh', width=5, placeholder="...")
'...'

oriyarde: my bad.
try `'{}...'.format(textwrap.wrap('abcdefgh', width=5)[0], '...')`

controller/array_action/array_mediator_svc.py
oriyarde: it shouldn't be hard, but could you address my question? it's still not clear to me

controller/array_action/array_mediator_svc.py
oriyarde: why double? it should also do 3 times less appends.
anyway, you can also try something like `map(writer.write, ('lshost ', host_id, ';'))`

controller/array_action/array_mediator_svc.py
igorgonibm: This is the same thing i did except it is not one line
i pefere m y way

controller/array_action/array_mediator_svc.py
igorgonibm: If not <empty list>

controller/array_action/utils.py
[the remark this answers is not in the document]
oriyarde: do you know if any other storage type does this conversion?
or you just think we should put storage-specific logic here?

controller/array_action/array_mediator_svc.py
igorgonibm: It first "formats" pass for each char in "lshost 1" and then "writes to writer" for each char.in "lshost 1"

controller/array_action/config.py
```diff
@@ -8,3 +8,5 @@
 
 # volume context
 CONTEXT_POOL = "pool"
+
+UTF_8 = "utf-8"
```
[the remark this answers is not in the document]
oriyarde: why did you resolve this? I still thinks this is the wrong location for it

controller/array_action/array_mediator_svc.py
oriyarde: ok. I guess it's good enough

controller/array_action/utils.py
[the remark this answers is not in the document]
igorgonibm: This func is not storage specific. It is general purpose func
I am stopping this discussion

controller/array_action/array_mediator_svc.py
oriyarde:
```
import timeit
from io import StringIO

def _first(host_list):
	writer = StringIO()
	for host_id in host_list:
		writer.write('lshost')
		writer.write(' ')
		writer.write(host_id)
		writer.write(';')
	return writer.getvalue()


def _second(host_list):
    writer = StringIO()
    for host_id in host_list:
        writer.write('lshost {HOST_ID};'.format(HOST_ID=host_id))
    return writer.getvalue()


four_k = map(str, range(4000))

t1 = timeit.Timer("_first(four_k)", "from __main__ import _first, four_k")
first_time = t1.timeit(1)
print("1st took %fs" % (first_time,))

t2 = timeit.Timer("_second(four_k)", "from __main__ import _second, four_k")
second_time = t2.timeit(1)
print("2nd took %fs" % (second_time,))
```
1st took 0.005819s
2nd took 0.000017s

oriyarde: I'm not sure I follow the theory.
did you try to compare their time?

controller/array_action/array_mediator_svc.py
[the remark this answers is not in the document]
oriyarde: what do you mean it's not open for discussion? the review is exactly the place for these discussions.
I don't understand; couldn't find any "get_hosts_for_init" - could you point me to the ticket?

controller/array_action/svc_cli_result_reader.py
```diff
@@ -0,0 +1,117 @@
...
+    """
+    def __init__(self, hosts_raw_list_as_string):
+        self._hosts_raw_list = hosts_raw_list_as_string.splitlines()
+        self._current_index = 0
+        self._next_object_id = None
+        self._init_first_object_id()
+
+    def _init_first_object_id(self):
+        """
+        Set _init_first_object_id to id o the first object and _current_index to point to next line
```
oriyarde, line 25: did you mean `id 0`? if so, does it have to be 0?

controller/array_action/svc_cli_result_reader.py
```diff
@@ -0,0 +1,117 @@
...
+            if not line:
+                continue
+            param_name, param_value = self._parse_param(line)
+            if param_name == ID_PARAM_NAME:
+                self._next_object_id = param_value
+                return res
+            res.add(param_name, param_value)
+        return res
+
+    def _parse_param(self, line):
```
oriyarde, line 72: all this function for `name, _, value = line.partition(' ')`? what am I missing here?
you can also add `value = value.strip()`, but I am not aware of such need

controller/array_action/config.py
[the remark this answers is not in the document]
igorgonibm: This is general constant. Not svc related. Even if it currently is used in svc.

controller/array_action/svc_cli_result_reader.py
igorgonibm: no
there may be host without id 0

controller/array_action/svc_cli_result_reader.py
igorgonibm: 1) I don't assume space separator but and whitespace
2) For empty value there ay be only single sting e.g. line "iscsi_name iscsi.ibm " or "iscsi_name" if no iscsi
3) value can contain spaces

controller/tests/array_action/svc/array_mediator_svc_test.py
```diff
@@ -231,162 +234,158 @@ def test_properties(self):
...
-                                  'WWPN': ['abc3'],
-                                  'iscsi_name': 'iqn.test.3'})
-        ret1 = [host_munch_ret_1, host_munch_ret_2]
-        ret2 = Munch(as_single_element=host_munch_ret_2)
-        ret3 = Munch(as_single_element=host_munch_ret_3)
-        self.svc.client.svcinfo.lshost.side_effect = [ret1, ret2, ret3]
+    @patch("controller.array_action.svc_cli_result_reader.SVCListResultsReader.__iter__")
+    def test_get_host_by_identifiers_return_iscsi_host(self, result_reader_iter):
+        host_1 = self._get_host_as_dictonary('host_id_1', 'test_host_1', [], ['abc1'])
+        host_2 = self._get_host_as_dictonary('host_id_2', 'test_host_2', ['iqn.test.2'], ['abc3'])
```
roysahar-ibm, line 275: host2 is 'abc3'?

controller/tests/array_action/svc/array_mediator_svc_test.py
```diff
@@ -231,162 +234,158 @@ def test_properties(self):
...
-        host_munch_ret_3 = Munch({'id': 'host_id_3', 'name': 'test_host_3',
-                                  'WWPN': ['abc3'],
-                                  'iscsi_name': 'iqn.test.3'})
-        ret1 = [host_munch_ret_1, host_munch_ret_2]
-        ret2 = Munch(as_single_element=host_munch_ret_2)
-        ret3 = Munch(as_single_element=host_munch_ret_3)
-        self.svc.client.svcinfo.lshost.side_effect = [ret1, ret2, ret3]
+    @patch("controller.array_action.svc_cli_result_reader.SVCListResultsReader.__iter__")
+    def test_get_host_by_identifiers_return_iscsi_host(self, result_reader_iter):
+        host_1 = self._get_host_as_dictonary('host_id_1', 'test_host_1', [], ['abc1'])
```
roysahar-ibm, line 274: host1 does not have iqn? this is for coverage of cases?

controller/array_action/svc_cli_result_reader.py
igorgonibm: but this is the exact example. Lines with \n char separator
Do u gave any other idea how to write the example? I can do this way  but less like it
"id 1
name name1
id 2
name name2"

controller/array_action/svc_cli_result_reader.py
oriyarde: if it's fixed, one of you should resolve this

controller/array_action/array_mediator_svc.py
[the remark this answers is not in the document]
oriyarde: 2. could you also address this function length issue?

controller/array_action/config.py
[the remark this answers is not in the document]
oriyarde: even if you want single-use consts here, the way I see this file - "config" means values that we want the option to easily change in the future, and since this value won't change in python 3, we won't need to re-"configure" it.

regardless of where you eventually place it, make sure you don't duplicate the same const (e.g. `utils.py`)

controller/array_action/svc_cli_result_reader.py
oriyarde: I still don't understand this line.
what does "Set _init_first_object_id" mean? `_init_first_object_id` is the function name, not a data member

controller/array_action/svc_cli_result_reader.py
oriyarde: I don't think 1. is a real scenario, but I'm pretty sure my suggestion works with 2. + 3...

controller/tests/array_action/svc/array_mediator_svc_test.py
igorgonibm: yes This host is without iscsi. only fc

controller/tests/array_action/svc/array_mediator_svc_test.py
igorgonibm: abc3 is wwn.
This was original test value. I ddidn't change it

controller/tests/array_action/svc/array_mediator_svc_test.py
```diff
@@ -231,162 +234,158 @@ def test_properties(self):
...
+        self.svc.client.svcinfo.lshost.return_value = self._get_hosts_list_result(hosts)
+        self.svc.client.send_raw_command = Mock()
+        self.svc.client.send_raw_command.return_value = EMPTY_BYTES, EMPTY_BYTES
+        result_reader_iter.return_value = self._get_detailed_hosts_list_result(hosts)
+        host, connectivity_type = self.svc.get_host_by_host_identifiers(Initiators('iqn.test.2', ['ABC3']))
         self.assertEqual('test_host_3', host)
         self.assertEqual([config.ISCSI_CONNECTIVITY_TYPE,
                           config.FC_CONNECTIVITY_TYPE], connectivity_type)
 
+    def _get_host_as_dictonary(self, id, name, iscsi_names_list, wwpns_list):
```
oriyarde, line 359: *dictionary

controller/array_action/array_mediator_svc.py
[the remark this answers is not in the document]
igorgonibm: Your loop doesn't handle list values. See the reader class to understand

controller/array_action/array_mediator_svc.py
oriyarde: I agree that the logic is very similar, but what about the naming?
could you make the names more general? e.g.:
`_get_formatted_hosts_list_error_msg` -> `truncate`
`detailed_host_list_errors` -> `error_msgs`

controller/array_action/array_mediator_svc.py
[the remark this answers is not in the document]
oriyarde: please address all numbered issues (with numbers).
I think your last comment is for a different existing conversation (simply adding `value = value.split() if ' ' in value else value` after my `value = value.strip()`)

controller/array_action/array_mediator_svc.py
oriyarde: I agree the logic is similar.
but could you still rename it to something more general? e.g.:
`_get_formatted_hosts_list_error_msg` -> `truncate_errors`
`detailed_host_list_errors` -> `errors_text`

btw I don't see this function being called anywhere.

controller/array_action/array_mediator_svc.py
igorgonibm: fixed

oriyarde: thanks! but I still think the fix should include a Unit Test (to avoid this confusion in the future).
if you think it's too much for this ticket, then at least open a ticket and add it here

controller/array_action/array_mediator_svc.py
oriyarde: if you want to save the time comparisons, this is the fastest and most pythonic approach:
`return 'lshost {}'.format(';lshost '.join(host_list))`

controller/array_action/svc_cli_result_reader.py
igorgonibm: @roysahar-ibm how do u wan t me to change it?

controller/array_action/svc_cli_result_reader.py
igorgonibm: @roysahar-ibm yes typo. Fixing

igorgonibm: Fixed the comment

controller/array_action/svc_cli_result_reader.py
igorgonibm: done

controller/array_action/array_mediator_svc.py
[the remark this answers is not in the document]
oriyarde: yes, the loop should be in a new (second) function.
but there should also be another new (first) function for the first part (from `self.client.svcinfo.lshost()` to `if not detailed_hosts_list_errors`)

controller/array_action/config.py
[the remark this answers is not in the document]
oriyarde: I don't remember when we talked about it, but I don't think I should open a ticket for a small rename.
if you think it is too much for this ticket, please open a new one and add the number here (we shouldn't treat `config` as `constants` without at least a ticket for renaming it)

controller/array_action/array_mediator_svc.py
oriyarde: I don't understand from your last response why the writer is faster - is this about the `map` overhead?

controller/array_action/svc_cli_result_reader.py
```diff
@@ -0,0 +1,107 @@
...
+            line = self._hosts_raw_list[self._current_index].strip()
+            self._current_index += 1
+            if line:
+                param_name, _, param_value = line.partition(' ')
+                param_value = param_value.strip()
+                if param_name == ID_PARAM_NAME:
+                    self._next_object_id = param_value
+                else:
+                    raise controller_errors.InvalidCliResponseError(
+                        "First element is {0}. Expected {1}".format(line, ID_PARAM_NAME))
```
oriyarde, line 40: shouldn't it be `First param is ...`? I thought "element" is the complete object, e.g. cli host

controller/tests/array_action/svc/array_mediator_svc_test.py
```diff
@@ -231,162 +234,158 @@ def test_properties(self):
...
+        host, connectivity_type = self.svc.get_host_by_host_identifiers(Initiators('iqn.test.2', ['ABC3']))
         self.assertEqual('test_host_3', host)
         self.assertEqual([config.ISCSI_CONNECTIVITY_TYPE,
                           config.FC_CONNECTIVITY_TYPE], connectivity_type)
 
+    def _get_host_as_dictionary(self, id, name, iscsi_names_list, wwpns_list):
+        res = {}
+        res[HOST_ID_PARAM] = id
+        res[HOST_NAME_PARAM] = name
+        if iscsi_names_list is not None:
```
oriyarde, line 363: I don't think they are ever `None`.
did you mean to add `None` as a default parameter value?

controller/tests/array_action/svc/array_mediator_svc_test.py
```diff
@@ -231,162 +234,158 @@ def test_properties(self):
...
+        self.svc.client.send_raw_command = Mock()
+        self.svc.client.send_raw_command.return_value = EMPTY_BYTES, EMPTY_BYTES
+        result_reader_iter.return_value = self._get_detailed_hosts_list_result(hosts)
+        host, connectivity_type = self.svc.get_host_by_host_identifiers(Initiators('iqn.test.2', ['ABC3']))
         self.assertEqual('test_host_3', host)
         self.assertEqual([config.ISCSI_CONNECTIVITY_TYPE,
                           config.FC_CONNECTIVITY_TYPE], connectivity_type)
 
+    def _get_host_as_dictionary(self, id, name, iscsi_names_list, wwpns_list):
+        res = {}
```
oriyarde, line 360: you can put the unconditional values in this line (`{key: value}`)

controller/tests/array_action/svc/array_mediator_svc_test.py
```diff
@@ -231,162 +234,158 @@ def test_properties(self):
...
+        res[HOST_ID_PARAM] = id
+        res[HOST_NAME_PARAM] = name
+        if iscsi_names_list is not None:
+            res[HOST_ISCSI_NAMES_PARAM] = iscsi_names_list
+        if wwpns_list is not None:
+            res[HOST_WWPNS_PARAM] = wwpns_list
+        return res
+
+    def _get_hosts_list_result(self, hosts_dict):
```
oriyarde, line 369: remove this function and change `_get_host_as_dictionary` to `_get_host_as_munch`

controller/array_action/array_mediator_svc.py
```diff
@@ -294,6 +297,53 @@ def get_host_by_host_identifiers(self, initiators):
...
+
+        logger.debug("Reading detailed hosts list commands batch response")
+        hosts_reader = SVCListResultsReader(detailed_hosts_list_output)
+        res = []
+        for host_details in hosts_reader:
+            host_id = host_details.get(HOST_ID_PARAM)
+            host_name = host_details.get(HOST_NAME_PARAM)
+            iscsi_names = host_details.get_as_list(HOST_ISCSI_NAMES_PARAM)
+            wwns = host_details.get_as_list(HOST_WWPNS_PARAM)
+            host = Host(host_id, host_name, iscsi_names, wwns)
```
oriyarde, line 321: opening the Host class vs Munch issue here, since the [conversation with numbered issues](https://github.com/IBM/ibm-block-csi-driver/pull/154/files#r421501818) is not convenient.
as I wrote there: could you please point me to the "get_hosts_for_init" ticket?

regardless of this ticket, I still think a Munch will save us the new consts, and conforms to the project's current convention for cli objects.

controller/array_action/svc_cli_result_reader.py
roysahar-ibm: Replace the "\n" with "<New Line>"

igorgonibm: but the input contains nerwline

controller/array_action/array_mediator_svc.py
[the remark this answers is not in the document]
igorgonibm: Moved loop to separate function
but not agree with the second part. This way the original method will just call two other methods so it is meaningless.

controller/array_action/array_mediator_svc.py
igorgonibm: How return 'lshost {}'.format(';lshost '.join(host_list))
do the job?

igorgonibm: I'll use: LIST_HOSTS_CMD_FORMAT = 'lshost {HOST_ID};'
writer.write(LIST_HOSTS_CMD_FORMAT.format(HOST_ID=host_id)

controller/array_action/array_mediator_svc.py
igorgonibm: done

controller/array_action/array_mediator_svc.py
igorgonibm: Resonsider about log assert here CSI-1492

controller/array_action/svc_cli_result_reader.py
igorgonibm: fixed

controller/tests/array_action/svc/array_mediator_svc_test.py
igorgonibm: done

controller/array_action/array_mediator_svc.py
igorgonibm: consider here CSI-1492

controller/tests/array_action/svc/array_mediator_svc_test.py
igorgonibm: consider CSI-1492

controller/tests/array_action/svc/array_mediator_svc_test.py
igorgonibm: removed

controller/array_action/config.py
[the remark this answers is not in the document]
igorgonibm: Removed from config

controller/array_action/svc_cli_result_reader.py
oriyarde: I meant that the "element" should be replaced by "param", and not remove "param"

igorgonibm: updated

controller/array_action/svc_cli_result_reader.py
igorgonibm: Will use <new line>

controller/array_action/array_mediator_svc.py
oriyarde: @igorgonibm why don't you use `:` slicing?

controller/array_action/utils.py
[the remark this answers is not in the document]
oriyarde: I think we can avoid all this raw bytes conversion by using `self.client.transport.send_command` (instead of `self.client.send_raw_command`, similar to [the capacity type conversion avoidance trick in ds8k](https://github.com/IBM/ibm-block-csi-driver/pull/122#discussion_r381554222))
so I'm unresolving this for CSI-1492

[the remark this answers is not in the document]
oriyarde: fixed in https://github.com/IBM/ibm-block-csi-driver/pull/432

PR #156, opened 2020-05-09
Improvement/csi 109 make pylint ignore explicitly

oriyarde:
instead of checking errors only, check any issue not explicitly disabled (by adding explicit whitelist of current messages)


Dockerfile-csi-controller.test
```diff
@@ -33,7 +33,7 @@ COPY ./controller /driver/controller
 USER root
 RUN groupadd -g 9999 appuser && \
```
ArbelNathan, line 34: isn't better practice to write the group name in cap? (e.g. APPUSER)

oriyarde: not that I'm aware of. but I didn't write this, I only chown another dir for some pylint pickle

PR #157, opened 2020-05-18
Story/csi 1449 unpublish vol no volume

igorgonibm:
If during unpublish volume already deleted OK should be returned
https://jira.xiv.ibm.com/browse/CSI-1449

### DIFF
--- a/controller/controller_server/csi_controller_server.py
+++ b/controller/controller_server/csi_controller_server.py
@@ -271,10 +271,12 @@ class ControllerServicer(csi_pb2_grpc.ControllerServicer):
 
         except controller_errors.VolumeAlreadyUnmappedError:
             logger.debug("Idempotent case. volume is already unmapped.")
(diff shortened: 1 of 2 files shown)


oriyarde:
> If during unpublish volume already deleted OK should be returned
> https://jira.xiv.ibm.com/browse/CSI-1449

@igorgonibm please remove internal link (you may keep the ticket number, if you think it's needed in addition to the branch and PR names)

controller/controller_server/csi_controller_server.py
```diff
@@ -273,12 +273,16 @@ def ControllerUnpublishVolume(self, request, context):
             logger.debug("Idempotent case. volume is already unmapped.")
             return csi_pb2.ControllerUnpublishVolumeResponse()
 
+        except controller_errors.VolumeNotFoundError as ex:
```
oriyarde, line 276: why not change the mediator interface (+implementations) to be idempotent?

PR #159, opened 2020-05-20
Story/csi 752 del snap a9k v120

igorgonibm:
create volume from snapshot a9k


controller/controller_server/csi_controller_server.py
```diff
@@ -8,6 +8,7 @@
 from retry import retry
 
 import controller.array_action.errors as controller_errors
+from controller.controller_server.errors import ObjectIdError
```
oriyarde, line 11:
```
import ...
from ...
import ...
```
please fix the import order

controller/controller_server/utils.py
```diff
@@ -1,7 +1,7 @@
 from controller.common.csi_logger import get_stdout_logger
 import controller.controller_server.config as config
 from controller.csi_general import csi_pb2
-from controller.controller_server.errors import ValidationException
+from controller.controller_server.errors import ValidationException, ObjectIdError
```
oriyarde, line 4: I know your addition here is valid, but still, since you already have a change here, please fix the existing imports order

igorgonibm: ok

controller/controller_server/csi_controller_server.py
igorgonibm: i let it auto fix

controller/array_action/array_mediator_xiv.py
```diff
@@ -264,6 +267,25 @@ def create_snapshot(self, name, volume_name):
...
+    def delete_snapshot(self, snapshot_id):
+        logger.info("Deleting snapshot with id : {0}".format(snapshot_id))
+        try:
+            snapshot_name = self._get_vol_by_wwn(snapshot_id)
+        except controller_errors.VolumeNotFoundError:
+            raise controller_errors.SnapshotNotFoundError(snapshot_id)
+
+        try:
+            self.client.cmd.snapshot_delete(snapshot=snapshot_name)
+        except xcli_errors.VolumeBadNameError as ex:
```
oriyarde, line 279: how can this happen, if the name comes from the storage?

controller/array_action/array_mediator_xiv.py
```diff
@@ -264,6 +267,25 @@ def create_snapshot(self, name, volume_name):
...
 
+    def delete_snapshot(self, snapshot_id):
+        logger.info("Deleting snapshot with id : {0}".format(snapshot_id))
+        try:
+            snapshot_name = self._get_vol_by_wwn(snapshot_id)
+        except controller_errors.VolumeNotFoundError:
+            raise controller_errors.SnapshotNotFoundError(snapshot_id)
+
+        try:
+            self.client.cmd.snapshot_delete(snapshot=snapshot_name)
```
oriyarde, line 278: why does `SnapshotIdBelongsToVolumeError` exist in the interface only for "create volume from snapshot" scenario, and not here? can't we get `CommandFailedRuntimeError("The operation is permitted on snapshots only.")` here?

oriyarde:
@igorgonibm do we also plan to support "retain" deletion policy?

PR #162, opened 2020-05-27
object_id_error_refactring

igorgonibm:


igorgonibm:
Eran said to push without review since this change passed review as part of delete snapshot

oriyarde:
@igorgonibm could you delete this branch?

oriyarde:
> Eran said to push without review since this change passed review as part of delete snapshot

could you add a link here to said review?

PR #175, opened 2020-06-14
ds8k_test_is_active

igorgonibm:
CSI-1551: don't use not valid ds8k connections in connection pool

### DIFF
--- a/controller/array_action/array_mediator_ds8k.py
+++ b/controller/array_action/array_mediator_ds8k.py
@@ -158,7 +158,7 @@ class DS8KArrayMediator(ArrayMediatorAbstract):
         pass
 
     def is_active(self):
-        return True
+        return self.client.is_valid()
 
     def get_system_info(self):
         return self.client.get_system()
--- a/controller/array_action/ds8k_rest_client.py
+++ b/controller/array_action/ds8k_rest_client.py
@@ -74,6 +74,16 @@ class RESTClient(object):
 
         self._client = Client(**client_kwargs)
 
+    def is_valid(self):
+        try:
+            # Send command to check if client is still valid. IO enclosures is chosen since it is fast.
+            self._client.get_io_enclosures()
+            return True
+        except Exception as ex:
+            logger.info("The client is not valid")
+            logger.exception(ex)
+            return False
+
     def get_system(self):
         return self._client.get_systems()[0]
 


controller/array_action/ds8k_rest_client.py
```diff
@@ -74,6 +74,16 @@ def __init__(self, service_address, user, password,
 
         self._client = Client(**client_kwargs)
 
+    def is_valid(self):
+        try:
+            # Send command to check if client is still valid. IO enclosures is chosen since it is fast.
+            self._client.get_io_enclosures()
+            return True
+        except Exception as ex:
```
oriyarde, line 82: shouldn't there be a specific error to expect here?
see: https://github.com/IBM/ibm-block-csi-driver/pull/135#discussion_r437391502

controller/array_action/array_mediator_ds8k.py
```diff
@@ -158,7 +158,7 @@ def disconnect(self):
         pass
 
     def is_active(self):
-        return True
+        return self.client.is_valid()
```
oriyarde, line 161: why `valid` and not `active`?

controller/array_action/ds8k_rest_client.py
igorgonibm: Possible but since I don't know what is returned for example when token is up-to-date I prefer to handle any exception this way.
In  the attacked comment it can't happen since we create new connection there.

controller/array_action/array_mediator_ds8k.py
igorgonibm: because it can be with illegal password which means it is active but invalid

oriyarde: but, if I understand correctly, `is_active` currently returns `False` on invalid password exception

igorgonibm: yes. Because we can't really check is active for ds8k so instead we assume it is active if and only if it is valid.

oriyarde:
> CSI-1551

@igorgonibm could you add it to the PR name as well?

controller/array_action/array_mediator_ds8k.py
oriyarde: why can't we check in ds8k if the exception is due to invalid credentials?

igorgonibm: as i wrote - we can't check if token is not up-to-date
and anyway i want to handle any unexpected error

controller/array_action/ds8k_rest_client.py
oriyarde: why can't we know what is returned when token is (not) up-to-date?
besides the generally bad practice of a catch-all `Exception`, I think we risk all ds8k functionality if there would be any issue specifically with `get_io_enclosures` that is not related to authentication/connectivity.
and why can't it happen when we create a new connection there?

controller/array_action/array_mediator_ds8k.py
oriyarde: if token is not up-to-date, what does it mean in terms of active and valid?

oriyarde: and do "invalid password" and "expired token" raise the same exceptions?

controller/array_action/ds8k_rest_client.py
igorgonibm: 1) get_io_enclosures  is list command without arguments which should never fall when connection is valid and storage is up
2) we can't check since i don't know when is its expiration. it may be days.,
3) This is bad practice but i don't know exactly which exceptions can happen here

igorgonibm: 4. Token can't be expired a second after creating connection.

controller/array_action/array_mediator_ds8k.py
igorgonibm: 1) I'm not entering discussion about method name. 
2) I don't know which exception is raised

oriyarde: so could you just add a ticket number for after GA to resolve remaining conversations here?

controller/array_action/ds8k_rest_client.py
oriyarde: 1. I think `which should never fail` is not a good assumption 
2. just because you don't know, doesn't mean it cannot be checked
3. I think it's worth checking
4. so is it called there only right after creation? I thought it is used for later checks as well

igorgonibm:
c open issues: CSI-1554

PR #177, opened 2020-06-15
Fix/csi 1550 svc create volume from snapshot gap

oriyarde:
keep snapshot to volume mapping until snapshot deletion

### DIFF
--- a/controller/array_action/array_mediator_svc.py
+++ b/controller/array_action/array_mediator_svc.py
@@ -40,7 +40,12 @@ HOSTS_LIST_ERR_MSG_MAX_LENGTH = 300
 CLI_OBJECT_TYPE_VOLUME = 'Volume'
 CLI_OBJECT_TYPE_FCMAP = 'FlashCopy Mapping'
 
+FCMAP_STATUS_DONE = 'idle_or_copied'
+
 YES = 'yes'
+
+ENDPOINT_TYPE_SOURCE = 'source'
+ENDPOINT_TYPE_TARGET = 'target'
 
 
 def is_warning_message(ex):
@@ -174,13 +179,14 @@ class SVCArrayMediator(ArrayMediatorAbstract):
         return self.client.transport.transport.get_transport().is_active()
 
     def _generate_volume_response(self, cli_volume):
+        source_volume_wwn = self._get_source_volume_wwn_if_exists(cli_volume)
         return Volume(
             int(cli_volume.capacity),
             cli_volume.vdisk_UID,
             cli_volume.name,
             self.endpoint,
             cli_volume.mdisk_grp_name,
-            None,
+            source_volume_wwn,
             self.array_type)
 
     def _generate_snapshot_response(self, cli_snapshot, source_volume_name):
@@ -236,6 +242,19 @@ class SVCArrayMediator(ArrayMediatorAbstract):
             raise controller_errors.VolumeNotFoundError(volume_name_or_id)
         return cli_volume
 
+    def _get_fcmap_as_target_if_exists(self, volume_name):
+        fcmaps_as_target = self._get_fcmaps(volume_name, ENDPOINT_TYPE_TARGET)
+        if len(fcmaps_as_target) != 1:
+            return None
+        return fcmaps_as_target[0]
+
+    def _get_source_volume_wwn_if_exists(self, target_cli_volume):
+        fcmap = self._get_fcmap_as_target_if_exists(target_cli_volume.name)
+        if not fcmap:
+            return None
+        source_volume_name = fcmap.source_vdisk_name
+        return self._get_wwn_by_volume_name_if_exists(source_volume_name)
+
     def get_volume(self, volume_name, volume_context=None, volume_prefix=""):
         cli_volume = self._get_cli_volume(volume_name)
         return self._generate_volume_response(cli_volume)
@@ -243,10 +262,14 @@ class SVCArrayMediator(ArrayMediatorAbstract):
     def get_volume_name(self, volume_id):
         return self._get_vol_by_wwn(volume_id)
 
+    def _get_fcmaps(self, volume_name, endpoint_type):
+        filter_value = '{0}_vdisk_name={1}'.format(endpoint_type, volume_name)
+        return self.client.svcinfo.lsfcmap(filtervalue=filter_value).as_list
+
     def is_volume_has_snapshots(self, volume_id):
         volume_name = self._get_vol_by_wwn(volume_id)
-        filter_value = 'source_vdisk_name={0}'.format(volume_name)
-        return bool(self.client.svcinfo.lsfcmap(filtervalue=filter_value).as_list)
+        fcmaps = self._get_fcmaps(volume_name, ENDPOINT_TYPE_SOURCE)
+        return bool(fcmaps)
 
     def validate_supported_capabilities(self, capabilities):
         logger.debug("validate_supported_capabilities for "
@@ -272,7 +295,16 @@ class SVCArrayMediator(ArrayMediatorAbstract):
             return size_in_bytes - ret + 512
         return size_in_bytes
 
-    def _get_volume_name_by_wwn_if_exist(self, volume_id):
+    def _get_wwn_by_volume_name_if_exists(self, volume_name):
+        cli_volume = self._get_cli_volume_if_exists(volume_name)
+        if not cli_volume:
+            return None
+
+        wwn = cli_volume.vdisk_UID
+        logger.debug("found wwn : {0}".format(wwn))
+        return wwn
+
+    def _get_volume_name_by_wwn_if_exists(self, volume_id):
         filter_value = 'vdisk_UID=' + volume_id
         vol_by_wwn = self.client.svcinfo.lsvdisk(
             filtervalue=filter_value).as_single_element
@@ -284,7 +316,7 @@ class SVCArrayMediator(ArrayMediatorAbstract):
         return vol_name
 
     def _get_vol_by_wwn(self, volume_id):
-        vol_name = self._get_volume_name_by_wwn_if_exist(volume_id)
+        vol_name = self._get_volume_name_by_wwn_if_exists(volume_id)
         if not vol_name:
             raise controller_errors.VolumeNotFoundError(volume_id)
         return vol_name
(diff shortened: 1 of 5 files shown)


controller/array_action/array_mediator_svc.py
```diff
@@ -443,29 +479,45 @@ def _stop_and_delete_fcmap(self, fcmap_id):
...
-        self._stop_and_delete_fcmap(target_cli_volume.FC_id)
-        self._delete_volume_by_name(target_cli_volume.name)
+    def _delete_all_fcmaps_as_source_if_exist(self, snapshot_name):
+        snapshot_as_source_fcmaps = self._get_fcmaps(snapshot_name, ENDPOINT_TYPE_SOURCE)
+        unfinished_fcmaps = [fcmap.name for fcmap in snapshot_as_source_fcmaps
+                             if fcmap.status != FCMAP_STATUS_DONE]
+        if unfinished_fcmaps:
+            raise controller_errors.SnapshotIsStillInUseError(snapshot_name, unfinished_fcmaps)
+        for fcmap in snapshot_as_source_fcmaps:
+            self._delete_fcmap(fcmap.id, force=False)
```
ArbelNathan, line 489: no need to stop? ( self._stop_fcmap)

controller/array_action/array_mediator_svc.py
```diff
@@ -443,29 +479,45 @@ def _stop_and_delete_fcmap(self, fcmap_id):
...
+
+    def _delete_snapshot(self, cli_volume):
+        fcmap_id = cli_volume.FC_id
+        snapshot_name = cli_volume.name
+        if fcmap_id == 'many':
+            self._delete_all_fcmaps_as_source_if_exist(snapshot_name)
+        fcmap = self._get_fcmap_as_target_if_exists(snapshot_name)
+        if not fcmap:
+            raise controller_errors.SnapshotNotFoundError(snapshot_name)
```
ArbelNathan, line 498: is that what it means if "not fcmap"? that no Snapshot found?

controller/array_action/array_mediator_svc.py
oriyarde: no, since they should all be finished (same reason for `force=False`)

controller/array_action/array_mediator_svc.py
oriyarde: it's not a snapshot if it doesn't have a mapping to it, and this is the existing exception in the mediator interface.
do you think we should add "snapshot is volume" error/s for `delete_snapshot` in the interface?

ArbelNathan: well, I dont understand how do we really get to here if its not a snapshot..
the mediator server already verified that, right?

controller/array_action/array_mediator_svc.py
ArbelNathan: ok, I dont understand svc so I have no really idea what it means..

controller/array_action/array_mediator_svc.py
oriyarde: not that I'm aware of, but let me know if I missed it

controller/array_action/array_mediator_svc.py
```diff
@@ -236,17 +242,34 @@ def _get_cli_volume(self, volume_name_or_id):
             raise controller_errors.VolumeNotFoundError(volume_name_or_id)
         return cli_volume
 
+    def _get_fcmap_as_target_if_exists(self, volume_name):
+        fcmaps_as_target = self._get_fcmaps(volume_name, ENDPOINT_TYPE_TARGET)
+        if len(fcmaps_as_target) != 1:
```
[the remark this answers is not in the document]
oriyarde: no, volume can have many snapshots

controller/array_action/array_mediator_svc.py
```diff
@@ -236,17 +242,34 @@ def _get_cli_volume(self, volume_name_or_id):
...
+            return None
+        return fcmaps_as_target[0]
+
+    def _get_source_volume_wwn_if_exists(self, target_cli_volume):
```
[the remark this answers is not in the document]
oriyarde: but that's how we know it doesn't raise an exception if not exists (and usually to check for `None` instead)

controller/array_action/array_mediator_svc.py
```diff
@@ -236,17 +242,34 @@ def _get_cli_volume(self, volume_name_or_id):
             raise controller_errors.VolumeNotFoundError(volume_name_or_id)
         return cli_volume
 
+    def _get_fcmap_as_target_if_exists(self, volume_name):
```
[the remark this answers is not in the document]
oriyarde: I think 1 conversation (at least per PR) is enough to resolve this issue...

controller/array_action/array_mediator_svc.py
```diff
@@ -236,17 +242,34 @@ def _get_cli_volume(self, volume_name_or_id):
...
+        return self._get_wwn_by_volume_name_if_exists(source_volume_name)
+
     def get_volume(self, volume_name, volume_context=None, volume_prefix=""):
         cli_volume = self._get_cli_volume(volume_name)
         return self._generate_volume_response(cli_volume)
 
     def get_volume_name(self, volume_id):
         return self._get_vol_by_wwn(volume_id)
 
+    def _get_fcmaps(self, volume_name, endpoint_type):
```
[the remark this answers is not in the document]
oriyarde: done

controller/array_action/array_mediator_svc.py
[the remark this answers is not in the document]
oriyarde: I don't see any appropriate exception in the existing interface, and I couldn't find any scenario which would return an OK status for it

controller/array_action/array_mediator_svc.py
ArbelNathan: no verification.. but its odd, it will raise on snapshot object.. the k8s snapshot will say there is no snapshot..

oriyarde: I agree that this exception is not ideal for this scenario.
so what do you suggest?

1. adding an exception in the interface?
2. delete it without an error?
3. leave it as it is?
4. something else?

ArbelNathan: btw, what if its 'many' there? isn't a bad case?

ArbelNathan: I think exception and deletion

oriyarde: yes, `many` should cause a storage exception on deletion attempt

oriyarde: do you mean that we should delete and then raise an exception?
if so, the second retry would succeed, so the exception would not mean much

controller/array_action/array_mediator_svc.py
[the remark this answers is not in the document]
igorgonibm: In any case if there s more than 1 there is error. You added this code so it is in scope.
Not critical for GA you can open ticket for open conversations.

controller/array_action/array_mediator_svc.py
ArbelNathan: it will if we will do it before checking on target (if finished copying), if it not a snapshot, I cant see why to check it..

ArbelNathan: in other hand, if it not a snapshot, maybe its better if we wont delete it..

controller/array_action/array_mediator_svc.py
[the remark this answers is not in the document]
igorgonibm: Return None should be the standard behavior. We already have open conversations about it i other reviews (svc create snapshot i think).
You can leave it for open conversations ticket too.

controller/array_action/array_mediator_svc.py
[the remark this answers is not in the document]
oriyarde: I mean that this level of check is not covered at all in the project, so I think it should be done in a separate ticket for all relevant places (create snapshot, DS8K, etc...)

controller/array_action/array_mediator_svc.py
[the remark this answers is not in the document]
oriyarde: ok.
btw I think the most common behavior in the mediator(s) is raising an exception on not exist

oriyarde:
I opened CSI-1556 for non-GA conversations resolving

controller/array_action/array_mediator_svc.py
oriyarde: do you think it's mandatory for GA?
if not, then we would have more time to think about it until CSI-1556

ArbelNathan: no, really not mandatory

ArbelNathan: first verify its a snapshot, later delete all flashcopies

controller/array_action/array_mediator_svc.py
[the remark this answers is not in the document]
ArbelNathan: if you think that an error should be raise here, do it, if a new ticket is needed, open one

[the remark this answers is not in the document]
oriyarde: I don't think an error is needed here.
do you?

controller/array_action/array_mediator_svc.py
oriyarde: and if it's not a snapshot, should it still be deleted?

oriyarde: and is this current error ok?

ArbelNathan: if not snapshot, I think its not ok to remove flashcopies.. and the error seems right

oriyarde: so if there are any action items here, please specify.
if not, please resolve.

controller/array_action/array_mediator_svc.py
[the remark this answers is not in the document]
ArbelNathan: no, it seems to be handled later by the correct  context, as I understand it, right here it without context to just return an error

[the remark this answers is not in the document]
oriyarde: so if there are any action items here, please specify.
if not, please resolve.

controller/array_action/array_mediator_svc.py
ArbelNathan: first check if its a snapshot, if it is, remove flashcopies and continue, else, return snapshot not found and dont touch. that is my opinion.

oriyarde: so what exactly is missing in the current code?

ArbelNathan: self._delete_all_fcmaps_as_source_if_exist(snapshot_name) after if not fcmap:

oriyarde: but we already have this call right before `if not fcmap:`.
could you clarify what you mean?

PR #178, opened 2020-06-15
Fix/csi 1545 powerpc fc subsystem

oriyarde:
change PowerPC FC `/dev/disk/by-path` search from:
**`pci-*`**`fc-0x0000000000000000-lun-0 -> ../../sdX`
to:
`fc-0x0000000000000000-lun-0 -> ../../sdX` (without `pci-*` prefix)

### DIFF
--- a/node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
+++ b/node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
@@ -155,10 +155,10 @@ func (r OsDeviceConnectivityHelperScsiGeneric) GetMpathDevice(volumeId string, l
 	}
 	var devicePaths []string
 	var errStrings []string
-	var targetPath string
+	var subsystemPrefix string
 
 	if connectivityType == ConnectionTypeFC {
-		targetPath = fmt.Sprintf("/dev/disk/by-path/%s*", fcSubsystem)
+		subsystemPrefix = fcSubsystemPrefix
 		// In host, the path like this: /dev/disk/by-path/pci-0000:13:00.0-fc-0x500507680b25c0aa-lun-0
 		// So add prefix "0x" for the arrayIdentifiers
 		for index, wwn := range arrayIdentifiers {
@@ -166,8 +166,10 @@ func (r OsDeviceConnectivityHelperScsiGeneric) GetMpathDevice(volumeId string, l
 		}
 	}
 	if connectivityType == ConnectionTypeISCSI {
-		targetPath = "/dev/disk/by-path/ip*"
+		subsystemPrefix = "ip*"
 	}
+
+	var targetPath = fmt.Sprintf("/dev/disk/by-path/%s", subsystemPrefix)
 
 	logger.Debugf("GetMpathDevice: Start concurrent multipath devices search for volume : [%s]", volumeId)
 	mpathResChannel := make(chan *WaitForMpathResult)
--- a/node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
+++ b/node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
@@ -166,7 +166,7 @@ func (r OsDeviceConnectivityHelperScsiGeneric) GetMpathDevice(volumeId string, l
 		}
 	}
 	if connectivityType == ConnectionTypeISCSI {
-		subsystemPrefix = "ip*"
+		subsystemPrefix = "ip*-"
 	}
 
 	var targetPath = fmt.Sprintf("/dev/disk/by-path/%s", subsystemPrefix)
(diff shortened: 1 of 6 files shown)


node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
```diff
@@ -226,7 +228,7 @@ func (r OsDeviceConnectivityHelperScsiGeneric) GetMpathDevice(volumeId string, l
 func (r OsDeviceConnectivityHelperScsiGeneric) waitForMpath(targetPath string, connectivityType string, arrayIdentifier string, lunId int, volumeId string,
 	resChannel chan<- *WaitForMpathResult) {
 	lunIdStr := convertIntToScsilun(lunId)
-	dp := strings.Join([]string{targetPath, connectivityType, arrayIdentifier, "lun", lunIdStr}, "-")
+	dp := targetPath + strings.Join([]string{connectivityType, arrayIdentifier, "lun", lunIdStr}, "-")
```
igorgonibm, line 231: why this change?

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
```diff
@@ -155,20 +155,22 @@ func (r OsDeviceConnectivityHelperScsiGeneric) GetMpathDevice(volumeId string, l
...
 		// In host, the path like this: /dev/disk/by-path/pci-0000:13:00.0-fc-0x500507680b25c0aa-lun-0
 		// So add prefix "0x" for the arrayIdentifiers
 		for index, wwn := range arrayIdentifiers {
 			arrayIdentifiers[index] = "0x" + strings.ToLower(wwn)
 		}
 	}
 	if connectivityType == ConnectionTypeISCSI {
-		targetPath = "/dev/disk/by-path/ip*"
+		subsystemPrefix = "ip*-"
 	}
```
igorgonibm, line 171: leave devicePath instance variable as before and init it here. Don't define it here as local variable.
What is the reason u devicePath as instance variable?

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
oriyarde: we don't want the hyphen in case there is no prefix before `fc-0x0000000000000000-lun-0`

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
oriyarde: I'm not sure I understand what you mean.
could you clarify?

igorgonibm: I mean targetPath
Leave it as instance variable
and don't add subsystemPrefix  as instance variable

oriyarde: why?
and are you implying `targetPath` is not an instance variable in my current change? if so, why?

igorgonibm: My mistake sorry. They are not instance variables.

node/pkg/driver/device_connectivity/device_path_amd64.go
```diff
@@ -0,0 +1,4 @@
+package device_connectivity
+
+// for x86, the FC subsystem prefix is pci
```
igorgonibm, line 3: I like comments but here it just says exactly as the code line itself

node/pkg/driver/device_connectivity/device_path_ppc64le.go
```diff
@@ -0,0 +1,4 @@
+package device_connectivity
+
+// for PowerPC, no FC subsystem prefix is needed
```
igorgonibm, line 3: better change to "for Power PC there is no FC subsystem prefix"

node/pkg/driver/device_connectivity/device_path_s390x.go
```diff
@@ -1,4 +1,4 @@
 package device_connectivity
 
 // on Z-Systems, the udev subsystem is a Channel Command Word
-const fcSubsystem = "ccw"
```
igorgonibm, line 4: also here comment is not needed

node/pkg/driver/device_connectivity/device_path_ppc64le.go
oriyarde: do you know this for a fact? because I don't

node/pkg/driver/device_connectivity/device_path_s390x.go
oriyarde: the comment is not part of this PR (and you approved it in https://github.com/IBM/ibm-block-csi-driver/pull/124).
then why is it needed in Power?

node/pkg/driver/device_connectivity/device_path_amd64.go
oriyarde: then why do we want to keep the comment in Power?

node/pkg/driver/device_connectivity/device_path_ppc64le.go
igorgonibm: This is what you did here. Prefix is empty string -> no prefix.

node/pkg/driver/device_connectivity/device_path_amd64.go
igorgonibm: I would remove it in power too.

node/pkg/driver/device_connectivity/device_path_s390x.go
igorgonibm: I would remove it in power too.

node/pkg/driver/device_connectivity/device_path_ppc64le.go
oriyarde: doesn't mean that I know that FC **never** (not even outside CSI context) has a prefix on PowerPC

oriyarde: as I understood from the other conversations, I'm removing all 3 comments

node/pkg/driver/device_connectivity/device_path_amd64.go
oriyarde: done

node/pkg/driver/device_connectivity/device_path_s390x.go
oriyarde: done

PR #180, opened 2020-06-18
CSI-1575 fix volume name length error

oriyarde:


oriyarde:
I opened ticket CSI-1589 to finish review fixes

PR #182, opened 2020-06-21
Update README.md

rivkap:
i've updated some items  that were missing. did not include creating volume snapshots yet - let me know if this should be included as well

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Supported IBM storage systems:
 
 Supported operating systems:
   - RHEL 7.x (x86 architecture)
-  - RHCOS (x86, IBM Z, and IBM PowerPC architecture)
+  - RHCOS (x86, IBM Z, and IBM PowerPC architectures)
 
 Full documentation can be found on the [IBM Knowledge Center](https://www.ibm.com/support/knowledgecenter/SSRQ8T).
 
@@ -153,7 +153,7 @@ Use the 'kubectl create -f deploy/kubernetes/snapshot-controller' command to cre
 Apply each of the yamls, using the 'kubectl apply -f' command.
 
 #### 4. Configure storage system connectivity
-##### 4.1. Define the hostname of each Kubernetes node on the relevant storage systems with the valid WWPN(for Fibre Channel) or IQN(for iSCSI) of the node.
+##### 4.1. Define the hostname of each Kubernetes node on the relevant storage systems with the valid WWPN (for Fibre Channel) or IQN (for iSCSI) of the node.
 
 ##### 4.2. For Fibre Channel, configure the relevant zoning from the storage to the host.
 
--- a/README.md
+++ b/README.md
@@ -10,7 +10,6 @@ Supported container platforms:
   - Kubernetes v1.17
 
 Supported IBM storage systems:
-  - IBM FlashSystem 9100
   - IBM Spectrum Virtualize Family (including IBM Flash family members built with IBM Spectrum Virtualize (FlashSystem 5010, 5030, 5100, 7200, 9100, 9200, 9200R) and IBM SAN Volume Controller (SVC) models SV2, SA2)
   - IBM FlashSystem A9000/R
   - IBM DS8880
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -145,10 +144,18 @@ systemctl status multipathd
...
+##### 3.1 Install the Snapshot CRDs, once per cluster.
+Use the 'kubectl create -f config/crd' command to create each of the yaml files located in https://github.com/kubernetes-csi/external-snapshotter/tree/master/config/crd.
+Apply each of the yamls, using the 'kubectl apply - f' command.
+##### 3.2 Install the Common Snapshot Controller, once per cluster.
+Use the 'kubectl create -f deploy/kubernetes/snapshot-controller' command to create each of the yaml files located in https://github.com/kubernetes-csi/external-snapshotter/tree/master/deploy/kubernetes/snapshot-controller.
+Apply each of the yamls, using the 'kubectl apply -f' command.
 
-3.2. For Fibre Channel, configure the relevant zoning from the storage to the host.
+#### 4. Configure storage system connectivity
+##### 4.1. Define the hostname of each Kubernetes node on the relevant storage systems with the valid WWPN(for Fibre Channel) or IQN(for iSCSI) of the node.
```
igorgonibm, line 156: I think it is better to add add space after WWPN and after IQN

README.md
```diff
@@ -145,10 +144,18 @@ systemctl status multipathd
...
 ```
 
-#### 3. Configure storage system connectivity
-3.1. Define the hostname of each Kubernetes node on the relevant storage systems with the valid WWPN(for Fibre Channel) or IQN(for iSCSI) of the node.
+#### 3. Enable support for volume Snapshots on your Kubernetes Cluster.
+##### 3.1 Install the Snapshot CRDs, once per cluster.
+Use the 'kubectl create -f config/crd' command to create each of the yaml files located in https://github.com/kubernetes-csi/external-snapshotter/tree/master/config/crd.
+Apply each of the yamls, using the 'kubectl apply - f' command.
+##### 3.2 Install the Common Snapshot Controller, once per cluster.
+Use the 'kubectl create -f deploy/kubernetes/snapshot-controller' command to create each of the yaml files located in https://github.com/kubernetes-csi/external-snapshotter/tree/master/deploy/kubernetes/snapshot-controller.
```
igorgonibm, line 152: If we want to support different namespace add "Update the namespace to an appropriate value for your environment"

README.md
```diff
@@ -10,15 +10,14 @@ Supported container platforms:
   - Kubernetes v1.17
```
igorgonibm, line 10: I think these changes should also be added also o operator readme.

README.md
```diff
@@ -10,15 +10,14 @@ Supported container platforms:
...
-  - IBM FlashSystem 9100
   - IBM Spectrum Virtualize Family (including IBM Flash family members built with IBM Spectrum Virtualize (FlashSystem 5010, 5030, 5100, 7200, 9100, 9200, 9200R) and IBM SAN Volume Controller (SVC) models SV2, SA2)
   - IBM FlashSystem A9000/R
   - IBM DS8880
   - IBM DS8900
 
 Supported operating systems:
   - RHEL 7.x (x86 architecture)
-  - RHCOS (x86 and IBM Z architecture)
+  - RHCOS (x86, IBM Z, and IBM PowerPC architecture)
```
oriyarde, line 20: maybe architectures (plural)?

oriyarde:
> i've updated some items that were missing. did not include creating volume snapshots yet - let me know if this should be included as well

@rivkap I think "Enable support for volume Snapshots" should either be added together with "creating volume snapshots", or both shouldn't be added to the READMEs at all

rivkap:
@oriyarde  @igorgonibm I can't make the decision to keep or not keep the volume snapshot support information in the readme. Let me know if it should be there or not. and i can help delete it. Once we finish with this readme i'll update the second one as well.
Also - do we no longer support OpenShift 4.2?

oriyarde:
> @oriyarde @igorgonibm I can't make the decision to keep or not keep the volume snapshot support information in the readme. Let me know if it should be there or not. and i can help delete it. Once we finish with this readme i'll update the second one as well.
> Also - do we no longer support OpenShift 4.2?

IMO we should not include snapshot info in READMEs, I would even remove most of their existing content (in a different ticket). they are long enough and link to the full KC.

as of 1.2.0 we will no longer support OpenShift 4.2

PR #186, opened 2020-06-21
Fix/csi 1376 wrong errors ds8k

ArbelNathan:
follow up ticket: CSI-1597


controller/array_action/array_mediator_ds8k.py
```diff
@@ -23,7 +23,10 @@
 
 # response error codes
 ERROR_CODE_INVALID_CREDENTIALS = 'BE7A002D'
+KNOWN_ERROR_CODE_INVALID_CREDENTIALS = 'BE7A0029'
```
oriyarde, line 26: why `KNOWN_`?

controller/array_action/array_mediator_ds8k.py
```diff
@@ -23,7 +23,10 @@
...
 ERROR_CODE_RESOURCE_NOT_EXISTS = 'BE7A0001'
+INCORRECT_ID = 'BE7A0005'
+NO_TOKEN_IS_SPECIFIED = 'BE7A001A'
```
oriyarde, line 29: please ensure each error code is covered by a UT

controller/array_action/array_mediator_ds8k.py
ArbelNathan: I dont have the spec for the Errors and I dont know that error (same name without the known), so I didn't remove it.

oriyarde: I think it's true for all error codes here, and I think they should be aligned, so I would remove the prefix

oriyarde: if you want them to have different names, I would use `_1` + `_2` suffixes, or simply a list of them

PR #192, opened 2020-06-24
Fix/csi 1608 snapshot name prefix

ArbelNathan:
follow up on CSI-1609 to remove the use of shorten name methods

### DIFF
--- a/controller/array_action/array_mediator_ds8k.py
+++ b/controller/array_action/array_mediator_ds8k.py
@@ -95,6 +95,13 @@ def shorten_volume_name(name, prefix):
         return (prefix + settings.NAME_PREFIX_SEPARATOR + hashed)[:MAX_VOLUME_LENGTH]
 
 
+def shorten_snapshot_name(name):
+    prefix = name.split(settings.NAME_PREFIX_SEPARATOR, 2)[0]
+    if prefix != name:
+        return shorten_volume_name(name, prefix)
+    return shorten_volume_name(name, prefix="")
+
+
 class DS8KArrayMediator(ArrayMediatorAbstract):
     SUPPORTED_FROM_VERSION = '7.5.1'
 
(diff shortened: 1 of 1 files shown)


controller/array_action/array_mediator_ds8k.py
```diff
@@ -95,6 +95,13 @@ def shorten_volume_name(name, prefix):
         return (prefix + settings.NAME_PREFIX_SEPARATOR + hashed)[:MAX_VOLUME_LENGTH]
 
 
+def shorten_snapshot_name(name):
```
oriyarde, line 98: I think we can remove all name `shorten` logic in ds8k, and simply change `max_volume_name_length` from 63 to 16, since we now have generic logic: https://github.com/IBM/ibm-block-csi-driver/pull/180

controller/array_action/array_mediator_ds8k.py
```diff
@@ -583,8 +590,9 @@ def create_snapshot(self, name, volume_name, volume_context=None):
             logger.error(
                 "volume_context is not specified, can not get volumes from storage."
             )
+        full_name = shorten_snapshot_name(name)
```
oriyarde, line 593: why full? AFAIK `name` already contains the prefix here

controller/array_action/array_mediator_ds8k.py
```diff
@@ -95,6 +95,13 @@ def shorten_volume_name(name, prefix):
         return (prefix + settings.NAME_PREFIX_SEPARATOR + hashed)[:MAX_VOLUME_LENGTH]
 
 
+def shorten_snapshot_name(name):
+    prefix = name.split(settings.NAME_PREFIX_SEPARATOR)[0]
```
oriyarde, line 99: what if there is no prefix?

controller/array_action/array_mediator_ds8k.py
oriyarde, line 593: why is it not needed also in `get_snapshot`?

controller/array_action/array_mediator_ds8k.py
oriyarde, line 99: what happens when the prefix contains the separator?

PR #193, opened 2020-06-24
Fix/csi 1515 lun conversion

ArbelNathan:


controller/tests/array_action/ds8k/test_array_mediator_ds8k.py
```diff
@@ -308,7 +308,8 @@ def test_unmap_volume_failed_with_ClientException(self):
         self.client_mock.get_host_mappings.return_value = [
             Munch({
                 "volume": volume_id,
-                "lunid": lunid
+                "lunid": lunid,
```
oriyarde, line 311: why do we still need to mock the `lunid` attribute?

ArbelNathan: I dont, deleted it

PR #195, opened 2020-06-28
fix/CSI-1608 prefix snapshot ds8k

ArbelNathan:
follow up in ticket  CSI-1616


controller/array_action/array_mediator_ds8k.py
```diff
@@ -74,27 +73,6 @@ def hash_string(string):
...
-    :param name: the origin name with prefix
-    :param prefix: the prefix of the name
-    :return: a short name with the origin prefix
-    """
-
-    if len(name) <= MAX_VOLUME_LENGTH:
-        return name
-
-    if not prefix:
-        return hash_string(name)[:MAX_VOLUME_LENGTH]
```
oriyarde, line 91: what about removing `def hash_string`?

controller/array_action/array_mediator_ds8k.py
```diff
@@ -235,7 +211,7 @@ def create_volume(self, name, size_in_bytes, capabilities, pool_id, volume_prefi
         try:
             cli_kwargs = {}
             cli_kwargs.update({
-                'name': shorten_volume_name(name, volume_prefix),
```
oriyarde, line 238: I think `volume_prefix` can now be removed altogether (including mediator interface)

ArbelNathan: because its also from mediator interface I think its better for after GA

PR #199, opened 2020-07-05
fix/CSI-1616_FU_1608_prefix

ArbelNathan:
follow up for ticket CSI-1608
https://github.com/IBM/ibm-block-csi-driver/pull/195

### DIFF
--- a/controller/controller_server/csi_controller_server.py
+++ b/controller/controller_server/csi_controller_server.py
@@ -439,8 +439,9 @@ class ControllerServicer(csi_pb2_grpc.ControllerServicer):
                         return csi_pb2.CreateSnapshotResponse()
                 else:
                     logger.debug(
-                        "Snapshot doesn't exist. Creating a new snapshot {0} from volume {1}".format(snapshot_final_name,
-                                                                                                     volume_name))
+                        "Snapshot doesn't exist. Creating a new snapshot {0} from volume {1}".format(
+                            snapshot_final_name,
+                            volume_name))
                     snapshot = array_mediator.create_snapshot(snapshot_final_name, volume_name, request.parameters)
 
                 logger.debug("generating create snapshot response")
--- a/controller/controller_server/csi_controller_server.py
+++ b/controller/controller_server/csi_controller_server.py
@@ -84,7 +84,7 @@ class ControllerServicer(csi_pb2_grpc.ControllerServicer):
                 logger.debug(array_mediator)
                 # TODO: CSI-1358 - remove try/except
                 try:
-                    volume_full_name, volume_prefix = self._get_volume_name_and_prefix(request, array_mediator)
+                    volume_full_name = self._get_volume_name_and_prefix(request, array_mediator)
                 except controller_errors.IllegalObjectName as ex:
                     context.set_details(ex.message)
                     context.set_code(grpc.StatusCode.INVALID_ARGUMENT)
@@ -99,15 +99,14 @@ class ControllerServicer(csi_pb2_grpc.ControllerServicer):
                 try:
                     vol = array_mediator.get_volume(
                         volume_full_name,
-                        volume_context=request.parameters,
-                        volume_prefix=volume_prefix,
+                        volume_context=request.parameters
                     )
                 except controller_errors.VolumeNotFoundError:
                     logger.debug(
                         "volume was not found. creating a new volume with parameters: {0}".format(request.parameters))
 
                     array_mediator.validate_supported_capabilities(capabilities)
-                    vol = array_mediator.create_volume(volume_full_name, size, capabilities, pool, volume_prefix)
+                    vol = array_mediator.create_volume(volume_full_name, size, capabilities, pool)
                 else:
                     logger.debug("volume found : {}".format(vol))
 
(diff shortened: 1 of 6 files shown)


oriyarde:
> follow up for ticket CSI-1608
> passed CI:
> http://csijen.xiv.ibm.com:8080/view/CSI_Production_Matrix/job/Production_K8S_FC_DS8K/job/CSI_XAVI_production_FC_ds8k_tucs_k8s1.17/10/

please remove internal links

controller/controller_server/csi_controller_server.py
```diff
@@ -594,7 +593,7 @@ def _get_object_name_and_prefix(self, request, max_name_prefix_length, max_name_
         if len(full_name) > max_name_length:
             hashed_name = utils.hash_string(name)
             full_name = self._join_object_prefix_with_name(prefix, hashed_name)
-        return full_name[:max_name_length], prefix
+        return full_name[:max_name_length]
```
oriyarde, line 596: please update function name accordingly

ArbelNathan: done

oriyarde: what about other functions? e.g. `_get_volume_name_and_prefix`?

ArbelNathan: you right, missed it. fix

controller/controller_server/csi_controller_server.py
```diff
@@ -563,21 +562,21 @@ def GetPluginInfo(self, request, context):
...
-                                                   config.PARAMETERS_SNAPSHOT_NAME_PREFIX)
+        name = self._get_object_full_name(request, array_mediator.max_snapshot_prefix_length,
+                                          array_mediator.max_snapshot_name_length,
+                                          config.OBJECT_TYPE_NAME_SNAPSHOT,
+                                          config.PARAMETERS_SNAPSHOT_NAME_PREFIX)
         return name
 
-    def _get_object_name_and_prefix(self, request, max_name_prefix_length, max_name_length, object_type,
-                                    prefix_param_name):
+    def _get_object_full_name(self, request, max_name_prefix_length, max_name_length, object_type,
```
oriyarde, line 578: align related `get_x_name` function names, IMO by removing `full_`

controller/controller_server/csi_controller_server.py
```diff
@@ -84,7 +84,7 @@ def CreateVolume(self, request, context):  # pylint: disable=too-many-branches
                 logger.debug(array_mediator)
                 # TODO: CSI-1358 - remove try/except
                 try:
-                    volume_full_name, volume_prefix = self._get_volume_name_and_prefix(request, array_mediator)
+                    volume_full_name = self._get_volume_name(request, array_mediator)
```
oriyarde, line 87: consider also removing `full_` here (since there is no more prefix var to distinguish from)

ArbelNathan: there is a prefix var to distinguish from, its not part of the volume name from the request, that why it called full name, not because we used to get the prefix too..

controller/controller_server/csi_controller_server.py
ArbelNathan: actually, does "get" is the correct term here? its more like "create", "append" or "format", we connect the prefix and name, we don't get it as is.

controller/controller_server/csi_controller_server.py
oriyarde: it [used to be](https://github.com/IBM/ibm-block-csi-driver/blame/a189d1a1043f157051461176a2710a3c73252be4/controller/controller_server/csi_controller_server.py#L69) without `full_`, and [was added](https://github.com/IBM/ibm-block-csi-driver/blame/618a82a8bb6f20db5dc43bc515cbb7bd4e30f1fa/controller/controller_server/csi_controller_server.py#L60) during the ds8k hash addition.
but if you think it should still be distinguished from the original `request.name`, then I'm not sure `full_` is still accurate, since it might also be hashed.
`final`/`in_storage` may be more accurate (same for the "get" functions)

controller/controller_server/csi_controller_server.py
oriyarde: if adding a prefix is all it did, then we could've changed it to something more specific, like `join`.
but since it can also hash (and validates max lengths, etc...) then I don't know which verb can be more accurate.
but as I [wrote](https://github.com/IBM/ibm-block-csi-driver/pull/199/files#r450185142), perhaps something like `get_final_name`/`get_name_in_storage` can solve it

controller/controller_server/csi_controller_server.py

controller/controller_server/csi_controller_server.py
```diff
@@ -590,11 +589,11 @@ def _get_object_name_and_prefix(self, request, max_name_prefix_length, max_name_
...
                 )
-            full_name = self._join_object_prefix_with_name(prefix, name)
-        if len(full_name) > max_name_length:
+            final_name = self._join_object_prefix_with_name(prefix, name)
+        if len(final_name) > max_name_length:
             hashed_name = utils.hash_string(name)
-            full_name = self._join_object_prefix_with_name(prefix, hashed_name)
-        return full_name[:max_name_length], prefix
+            final_name = self._join_object_prefix_with_name(prefix, hashed_name)
+        return final_name[:max_name_length]
```
oriyarde, line 596: actually, I think full is more accurate in this context, since this slicing makes a final version from the full version

controller/controller_server/csi_controller_server.py

PR #185, opened 2020-06-21
Feature/csi 1552 cotroller fixs ds8k mediator

ArbelNathan:
separate ticket for finishing the review: CSI-1594

PR #200, opened 2020-07-08
Fix/csi 1594 fu 1552 mediator ds8k

ArbelNathan:
follow up for -
https://github.com/IBM/ibm-block-csi-driver/pull/185

### DIFF
--- a/controller/array_action/array_mediator_ds8k.py
+++ b/controller/array_action/array_mediator_ds8k.py
@@ -648,7 +648,6 @@ class DS8KArrayMediator(ArrayMediatorAbstract):
 
     def _check_snapshot_use_status(self, snapshot_id, flashcopies):
         for flashcopy in flashcopies:
-            logger.info("Deleting flashcopy: {}".format(flashcopy))
             if flashcopy.sourcevolume == snapshot_id:
                 flashcopy_obj = self._get_flashcopy(flashcopy.id)
                 if flashcopy_obj.out_of_sync_tracks != '0':
--- a/controller/array_action/array_mediator_ds8k.py
+++ b/controller/array_action/array_mediator_ds8k.py
@@ -14,6 +14,7 @@ from controller.array_action.utils import classproperty
 from controller.common import settings
 from controller.common.csi_logger import get_stdout_logger
 
+ARRAY_TYPE = "DS8K"
 LOGIN_PORT_WWPN = attr_names.IOPORT_WWPN
 LOGIN_PORT_STATE = attr_names.IOPORT_STATUS
 LOGIN_PORT_STATE_ONLINE = 'online'
@@ -308,7 +309,7 @@ class DS8KArrayMediator(ArrayMediatorAbstract):
             logger.error(
                 "pool_id is not specified, can not get volumes from storage."
             )
-            raise ValueError
+            raise array_errors.PoolIsRequired(ARRAY_TYPE)
 
         api_volume = self._get_api_volume_by_name(volume_name=name,
                                                   pool_id=pool_id)
@@ -410,7 +411,7 @@ class DS8KArrayMediator(ArrayMediatorAbstract):
             logger.error(
                 "pool_id is not specified, can not get volumes from storage."
             )
-            raise ValueError
+            raise array_errors.PoolIsRequired(ARRAY_TYPE)
         else:
             pools = [Munch({"id": pool_id})]
 
@@ -459,7 +460,7 @@ class DS8KArrayMediator(ArrayMediatorAbstract):
             logger.error(
                 "pool is not specified, can not get volumes from storage."
             )
-            raise ValueError
+            raise array_errors.PoolIsRequired(ARRAY_TYPE)
         target_api_volume = self._get_api_volume_by_name(volume_name=snapshot_name,
                                                          pool_id=pool_id)
         if not target_api_volume:
@@ -542,7 +543,7 @@ class DS8KArrayMediator(ArrayMediatorAbstract):
             logger.error(
                 "pool is not specified, can not get volumes from storage."
             )
-            raise ValueError
+            raise array_errors.PoolIsRequired(ARRAY_TYPE)
         target_api_volume = self._create_snapshot(name, pool_id, source_volume_name=volume_name)
         logger.info("finished creating snapshot '{0}' from volume '{1}'".format(name, volume_name))
         return self._generate_snapshot_response(target_api_volume, volume_name)
--- a/controller/array_action/array_mediator_ds8k.py
+++ b/controller/array_action/array_mediator_ds8k.py
@@ -309,7 +309,7 @@ class DS8KArrayMediator(ArrayMediatorAbstract):
         logger.info("Finished deleting volume {}".format(volume_id))
 
     def get_volume(self, name, pool_id=None, volume_prefix=""):
-        logger.debug("Getting volume {} under context {}".format(name, pool_id))
+        logger.debug("Getting volume {} in pool {}".format(name, pool_id))
         if not pool_id:
             logger.error(
                 "pool_id is not specified, can not get volumes from storage."
--- a/controller/array_action/array_mediator_ds8k.py
+++ b/controller/array_action/array_mediator_ds8k.py
@@ -14,7 +14,7 @@ from controller.array_action.utils import classproperty
 from controller.common import settings
 from controller.common.csi_logger import get_stdout_logger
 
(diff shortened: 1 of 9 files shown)


controller/array_action/array_mediator_ds8k.py
```diff
@@ -546,14 +535,14 @@ def get_snapshot_by_id(self, src_snapshot_id):
...
         logger.info("creating snapshot '{0}' from volume '{1}'".format(name, volume_name))
-        if not volume_context:
+        if not pool_id:
             logger.error(
-                "volume_context is not specified, can not get volumes from storage."
+                "pool is not specified, can not get volumes from storage."
             )
-        pool = volume_context[config.CONTEXT_POOL]
-        target_api_volume = self._create_snapshot(name, pool, source_volume_name=volume_name)
+            raise array_errors.SnapshotNotFoundError(name)
```
oriyarde, line 544: consider adding a separate Error for this scenario

controller/array_action/array_mediator_ds8k.py
```diff
@@ -650,10 +639,9 @@ def validate_flashcopy(self, flashcopy_id):
...
-    def _check_snapshot_use_status(self, snapshot_id, flashcopy_list):
-        for flashcopy in flashcopy_list:
-            logger.info("Deleting flashcopy: {}".format(flashcopy))
+    def _check_snapshot_use_status(self, snapshot_id, flashcopies):
+        for flashcopy in flashcopies:
             if flashcopy.sourcevolume == snapshot_id:
-                flashcopy_rel = self._get_flashcopy(flashcopy.id)
-                if flashcopy_rel.out_of_sync_tracks != '0':
-                    raise array_errors.SnapshotIsStillInUseError(snapshot_id, flashcopy_rel.targetvolume)
+                flashcopy_obj = self._get_flashcopy(flashcopy.id)
```
oriyarde, line 645: https://github.com/IBM/ibm-block-csi-driver/pull/185#discussion_r443222200
what is the difference between `flashcopy` and `flashcopy_obj` here?

controller/controller_server/csi_controller_server.py
```diff
@@ -406,6 +406,7 @@ def CreateSnapshot(self, request, context):
             context.set_code(grpc.StatusCode.INVALID_ARGUMENT)
             return csi_pb2.CreateSnapshotResponse()
 
+        pool = request.parameters.get(config.PARAMETERS_POOL)
```
oriyarde, line 409: why not `pool_id`?

controller/array_action/array_mediator_ds8k.py
ArbelNathan: flashcopy is small version of flashcopy in the storage, flashcopy_obj is the whole object from the storage the contain data of the copy process

controller/controller_server/csi_controller_server.py
ArbelNathan: because I copied the flow from create volume and tried to maintained the consistency

controller/array_action/array_mediator_ds8k.py
ArbelNathan: we have only PoolDoesNotExist and PoolDoesNotMatchCapabilities, and I thought that we add only cross platform  errors..

oriyarde: interesting gray area.
a new exception would match the ("cross-storage") interface `pool_id` parameter, but might actually be raised only in ds8k.
IMO it's not so bad, let's number this option as number 1.

here are a few other options to consider:
2. ensure `pool_id` value (if specified) matches the volume pool on storage, in other mediators as well, to actually use the new exception across storage types*
3. raise here a mediator-specific exception (a new "layer" of exceptions, between internal-storage and cross-storage exceptions)
4. raise here a **very** general, built-in exception (e.g. `ValueError`)
5. use existing `PoolDoesNotExist` (I don't like this option)

*a note about option 2: in xiv and svc, volumes with same name cannot exist, even in different pools.
so the only scenario I can think of that it may affect, is when a user changes the pool parameter for an existing pvc.

got any other options to consider?
which option do you prefer?

controller/array_action/array_mediator_ds8k.py
oriyarde: I feel this fact is not really conveyed with those current names.
consider renaming `flashcopy_obj` to something like `flashcopy_process` (everywhere relevant, i.e. including its `get` function)

controller/controller_server/csi_controller_server.py
oriyarde: i see. it might be the pool name on some of the other storage systems or something.

controller/array_action/array_mediator_ds8k.py
ArbelNathan: I think option 4 is my favorite.

controller/array_action/array_mediator_ds8k.py

controller/array_action/array_mediator_ds8k.py
oriyarde: I actually numbered the options from best to worst in my opinion.
IMO, the main downside of option 4 is that built-in exceptions may also be raised by 3rd parties, and I assumed we would like the controller server to `except` this scenario and then return CSI `INVALID_ARGUMENT` status code.
Do you have an idea how to mitigate this downside?

ArbelNathan: ok, done

controller/array_action/array_mediator_ds8k.py
```diff
@@ -14,6 +14,7 @@
 from controller.common import settings
 from controller.common.csi_logger import get_stdout_logger
 
+ARRAY_TYPE = "DS8K"
```
oriyarde, line 17: use existing `array_type`

controller/array_action/errors.py
```diff
@@ -99,6 +99,12 @@ def __init__(self, pool, array):
         self.message = messages.PoolDoesNotExist_message.format(pool, array)
 
 
+class PoolIsRequired(BaseArrayActionException):
```
oriyarde, line 102: this sounds more like a fact rather than an error.
consider renaming to something like `PoolParameterIsMissing/NotSpecified`

controller/array_action/array_mediator_ds8k.py
```diff
@@ -302,16 +303,16 @@ def delete_volume(self, volume_id):
         self._delete_volume(volume_id)
         logger.info("Finished deleting volume {}".format(volume_id))
 
-    def get_volume(self, name, volume_context=None):
-        logger.debug("Getting volume {} under context {}".format(name, volume_context))
-        if not volume_context:
+    def get_volume(self, name, pool_id=None):
+        logger.debug("Getting volume {} in pool {}".format(name, pool_id))
+        if not pool_id:
```
oriyarde, line 308: refactor to `if pool_id is not None:` to distinguish between unspecified vs. empty pool parameter

ArbelNathan: * if pool_id is None:

controller/array_action/array_mediator_ds8k.py

controller/array_action/errors.py
ArbelNathan: done

controller/array_action/array_mediator_ds8k.py

controller/array_action/messages.py
```diff
@@ -17,6 +17,8 @@
 
 PoolDoesNotExist_message = "Pool does not exist: {0} , array : {1}"
 
+PoolIsRequired = "pool parameter is mandatory in {0}"
```
oriyarde, line 20: sentences convention is to begin with Upper case

controller/array_action/array_mediator_ds8k.py
oriyarde: right

controller/array_action/errors.py
```diff
@@ -99,6 +99,12 @@ def __init__(self, pool, array):
         self.message = messages.PoolDoesNotExist_message.format(pool, array)
 
 
+class PoolParameterIsMissing(BaseArrayActionException):
+
+    def __init__(self, array_type):
+        self.message = messages.PoolIsRequired.format(array_type)
```
oriyarde, line 105: message const name should be aligned to exception class name

ArbelNathan: done

controller/array_action/messages.py
ArbelNathan: done

PR #201, opened 2020-07-08
 fix/CSI-1543_FU_1338_vol_from_snap_ds8k

ArbelNathan:
follow up for-
https://github.com/IBM/ibm-block-csi-driver/pull/173

### DIFF
--- a/controller/array_action/array_mediator_ds8k.py
+++ b/controller/array_action/array_mediator_ds8k.py
@@ -2,8 +2,8 @@ from munch import Munch
 from packaging.version import parse
 from pyds8k import exceptions
 from pyds8k.resources.ds8k.v1.common import attr_names
-from retry import retry
 from pyds8k.resources.ds8k.v1.common import types as ds8k_types
+from retry import retry
 
 import controller.array_action.errors as array_errors
 from controller.array_action import config
@@ -32,7 +32,7 @@ ERROR_CODE_VOLUME_NOT_FOUND_OR_ALREADY_PART_OF_CS_RELATIONSHIP = '00000013'
 
 FLASHCOPY_PERSISTENT_OPTION = ds8k_types.DS8K_OPTION_PER
 FLASHCOPY_NO_BACKGROUND_COPY_OPTION = ds8k_types.DS8K_OPTION_NBC
-FLASHCOPY_PERMIT_SPACE_EFFICIENT_TARGET = ds8k_types.DS8K_OPTION_PSET
+FLASHCOPY_PERMIT_SPACE_EFFICIENT_TARGET_OPTION = ds8k_types.DS8K_OPTION_PSET
 
 
 def parse_version(bundle):
@@ -504,7 +504,7 @@ class DS8KArrayMediator(ArrayMediatorAbstract):
         target_volume_id = get_volume_id_from_scsi_identifier(target_volume_id)
         if not options:
             options = []
-        options.append(FLASHCOPY_PERMIT_SPACE_EFFICIENT_TARGET)
+        options.append(FLASHCOPY_PERMIT_SPACE_EFFICIENT_TARGET_OPTION)
         try:
             api_flashcopy = self.client.create_flashcopy(source_volume_id=source_volume_id,
                                                          target_volume_id=target_volume_id,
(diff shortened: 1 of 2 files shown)


controller/array_action/array_mediator_ds8k.py
```diff
@@ -3,6 +3,7 @@
 from pyds8k import exceptions
 from pyds8k.resources.ds8k.v1.common import attr_names
 from retry import retry
+from pyds8k.resources.ds8k.v1.common import types as ds8k_types
```
oriyarde, line 6: please always use PyCharm's Optimize Imports (Ctrl+Alt+O)

controller/array_action/array_mediator_ds8k.py
```diff
@@ -29,9 +30,9 @@
 ERROR_CODE_ALREADY_FLASHCOPY = '000000AE'
 ERROR_CODE_VOLUME_NOT_FOUND_OR_ALREADY_PART_OF_CS_RELATIONSHIP = '00000013'
 
-FLASHCOPY_PERSISTENT_OPTION = "persistent"
-FLASHCOPY_NO_BACKGROUND_COPY_OPTION = "no_background_copy"
-FLASHCOPY_PERMIT_SPACE_EFFICIENT_TARGET = "permit_space_efficient_target"
+FLASHCOPY_PERSISTENT_OPTION = ds8k_types.DS8K_OPTION_PER
+FLASHCOPY_NO_BACKGROUND_COPY_OPTION = ds8k_types.DS8K_OPTION_NBC
+FLASHCOPY_PERMIT_SPACE_EFFICIENT_TARGET = ds8k_types.DS8K_OPTION_PSET
```
oriyarde, line 35: I know it's long, but consider adding `_OPTION`, for consistency

controller/array_action/array_mediator_ds8k.py
```diff
@@ -29,9 +30,9 @@
 ERROR_CODE_ALREADY_FLASHCOPY = '000000AE'
 ERROR_CODE_VOLUME_NOT_FOUND_OR_ALREADY_PART_OF_CS_RELATIONSHIP = '00000013'
 
-FLASHCOPY_PERSISTENT_OPTION = "persistent"
-FLASHCOPY_NO_BACKGROUND_COPY_OPTION = "no_background_copy"
-FLASHCOPY_PERMIT_SPACE_EFFICIENT_TARGET = "permit_space_efficient_target"
+FLASHCOPY_PERSISTENT_OPTION = ds8k_types.DS8K_OPTION_PER
```
oriyarde, line 33: btw, I know it's not really part of this ticket, but why is `FLASHCOPY_PERSISTENT_OPTION` **always** added to `options` outside `_create_flashcopy`, but `FLASHCOPY_PERMIT_SPACE_EFFICIENT_TARGET` is added inside?

ArbelNathan: because we always want to permit ESE but not necessarily always want the flashcopy to be persistent. now we do, but more functions will e added and that not always will be the case

controller/array_action/array_mediator_ds8k.py

controller/array_action/array_mediator_ds8k.py
ArbelNathan: done

controller/array_action/array_mediator_ds8k.py
oriyarde: when will we not want the flashcopy to be persistent?

ArbelNathan: when we wont be needing the source, like in clone..

oriyarde: why wouldn't we want to know the clone source?
does it matter if the source is a snapshot or a volume?

ArbelNathan: I dont know what will want to know right now, just tried to not be too specific..

oriyarde: so, since we currently don't have any solid reason to believe otherwise, I think that as long as the code requires both parameters everywhere, then it should be written accordingly (to add both parameters inside `_create_flashcopy`).
and if we ever want to change it, we can always do.

PR #207, opened 2020-07-15
CSI-1590 rename community csi log files

oriyarde:


scripts/ci/run_community_csi_test.sh
```diff
@@ -1,16 +1,14 @@
 #!/bin/bash -xe
 
-[ -z "${CONTROLLER_LOGS}" ] && { echo "CONTROLLER_LOGS env is mandatory"; exit 1; }
```
ArbelNathan, line 3: why no longer use CONTROLLER_LOGS var? what was its purpose?

oriyarde: It forced the job to specify a prefix for the log filename.
I don't think it's necessary, and especially not mandatory.

ArbelNathan: oh, ok

PR #216, opened 2020-08-18
CSI-109 fix controller imports

oriyarde:
fix controller imports once and for all

### DIFF
--- a/controller/scripts/lint.ini
+++ b/controller/scripts/lint.ini
@@ -23,7 +23,6 @@ ignore-patterns=csi_pb2.*[.]py
 # relevant categories: E - Error, W - Warning, R - Refactor, C - Convention
 disable=
  # easy:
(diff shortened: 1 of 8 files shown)


controller/scripts/lint.ini
```diff
@@ -23,7 +23,6 @@ ignore-patterns=csi_pb2.*[.]py
 # relevant categories: E - Error, W - Warning, R - Refactor, C - Convention
 disable=
  # easy:
- wrong-import-order, ungrouped-imports, unused-import, # C, C, W - fixable with PyCharm's Optimize Imports (Ctrl+Alt+O)
```
ArbelNathan, line 26: so now we will check imports in tests?

oriyarde: yes!

ArbelNathan: right now we are passing it?

oriyarde: yes...

PR #219, opened 2020-08-24
change to int and fix UT

ArbelNathan:


controller/scripts/lint.ini
```diff
@@ -23,6 +23,7 @@ ignore-patterns=csi_pb2.*[.]py
 # relevant categories: E - Error, W - Warning, R - Refactor, C - Convention
 disable=
  # easy:
+ raise-missing-from, # W - add "from" after some of the "raise"
```
oriyarde, line 26: forgot to mention that IMO it's worth adding **`happened only on Jenkins,`**`add "from"...`

ArbelNathan: as far as we know. I dont think its a fact that it cant happen in other places

oriyarde: that's why it's in past tense.
should be useful for reproduction purposes

ArbelNathan: done

oriyarde (approved):

matancarmeli7 (approved):

PR #226, opened 2020-09-08
fix/CSI-1657_wrong_response_class_return

ArbelNathan:
matching return type for the function returning it
a new follow up ticket opened CSI-1807


oriyarde (approved):
LGTM

lisyukanna (approved):
LGTM

PR #245, opened 2020-10-11
Dev ops/csi 1762 make build jobs pass registry credentials

matancarmeli7:


scripts/ci/jenkins_pipeline_csi
```diff
@@ -1,4 +1,11 @@
 pipeline {
+    environment {
+        registryCredential = 'csi_w3_user'
+        artifactory_path = 'build/reports/images_url'
+        registryUrl = 'https://stg-artifactory.haifa.ibm.com:5030'
```
oriyarde, line 5: this should come from the job configuration

scripts/ci/jenkins_pipeline_csi
```diff
@@ -25,9 +32,63 @@ pipeline {
...
+                    specific_tag = "${IMAGE_VERSION}_b${BUILD_NUMBER}_${branch}"
+                }
+            }
+        }
+        stage ('Build docker images') {
+            steps {
+                script {
+                    if (GIT_BRANCH == 'develop' || GIT_BRANCH == 'origin/develop' || GIT_BRANCH == 'master') {
+                        controller_image_latest = docker.build("${CSI_CONTROLLER_IMAGE}", "--build-arg VERSION=${IMAGE_VERSION} --build-arg BUILD_NUMBER=${BUILD_NUMBER} -f ${Dockerfile_controller} .")
+                        controller_image_with_generated_tag = docker.build("${DOCKER_REGISTRY}/${CSI_CONTROLLER_IMAGE}:${specific_tag}", "--build-arg VERSION=${IMAGE_VERSION} --build-arg BUILD_NUMBER=${BUILD_NUMBER} -f ${Dockerfile_controller} .")
```
oriyarde, line 49: this looks like a code dup of line 53.
are you sure we need an `else`?

scripts/ci/jenkins_pipeline_csi
```diff
@@ -25,9 +32,63 @@ pipeline {
...
+                    } else {
+                        docker.withRegistry(registryUrl, registryCredential ) {
+                            controller_image_with_generated_tag.push()
+                            node_image_with_generated_tag.push()
+                        }
+                    }
+                }
+            }
+        }
+        stage ('Save Artifactes') {
```
oriyarde, line 78: artifacts?

tikolsky (approved):

scripts/ci/jenkins_pipeline_csi
```diff
@@ -25,9 +32,63 @@ pipeline {
                 sh './scripts/run_yamlcheck.sh'
             }
         }
-        stage ('CSI-controller & node: Build and push images') {
+        stage ('Pre-build actions') {
```
oriyarde, line 35: I think we already have enough stages.
please move this to the next stage

scripts/ci/build_push_images.sh
```diff
@@ -1,51 +0,0 @@
-#!/bin/bash -xe
```
oriyarde, line 1: could we keep this file, and call it like this?
```
docker.withRegistry(...) {
    sh './scripts/ci/build_push_images.sh ...'
}
```
and if not, then could we at least call a `build_push_images.groovy` script?

scripts/ci/jenkins_pipeline_csi
oriyarde: I still see an `else` (can't comment on separate places in the same conversation).
let me rephrase - are you sure we need **any** else?

tikolsky (dismissed):

roysahar-ibm (dismissed):

oriyarde (changes requested):
@matancarmeli7 before merging, please address all remaining conversations in this pull request

scripts/ci/jenkins_pipeline_csi
```diff
@@ -1,4 +1,11 @@
 pipeline {
+    environment {
+        registryCredential = 'csi_w3_user'
+        artifactory_path = 'build/reports/images_url'
```
oriyarde, line 4: why not `registry_path`?

scripts/ci/jenkins_pipeline_csi
```diff
@@ -25,9 +32,46 @@ pipeline {
                 sh './scripts/run_yamlcheck.sh'
             }
         }
-        stage ('CSI-controller & node: Build and push images') {
+        stage ('Build docker images') {
```
oriyarde, line 35: could you please unify the three stages into one?

scripts/ci/build_push_images.sh
```diff
@@ -1,51 +0,0 @@
...
-
-# Validations
-MANDATORY_ENVS="IMAGE_VERSION BUILD_NUMBER DOCKER_REGISTRY CSI_NODE_IMAGE CSI_CONTROLLER_IMAGE GIT_BRANCH"
-for envi in $MANDATORY_ENVS; do 
-    [ -z "${!envi}" ] && { echo "Error - Env $envi is mandatory for the script."; exit 1; } || :
```
oriyarde, line 6: where did this check go?

scripts/ci/jenkins_pipeline_csi
```diff
@@ -25,9 +32,46 @@ pipeline {
...
+            steps {
+                script {
+                    branch = sh(script:"echo $GIT_BRANCH | sed 's|/|.|g'",returnStdout:true).trim()
+                    specific_tag = "${IMAGE_VERSION}_b${BUILD_NUMBER}_${branch}"
+                    if (GIT_BRANCH == 'develop' || GIT_BRANCH == 'origin/develop' || GIT_BRANCH == 'master') {
```
oriyarde, line 40: this check is duplicated. why not use a "tag_latest" var like in the original code?

scripts/ci/jenkins_pipeline_csi
matancarmeli7: why unify them? 
I think that it's more comfortable to see where is the problem from the Jenkins dashboard

scripts/ci/jenkins_pipeline_csi
matancarmeli7: because it's the path that the artifactory phase uses

scripts/ci/build_push_images.sh
matancarmeli7: done

scripts/ci/jenkins_pipeline_csi
matancarmeli7: done

scripts/ci/build_push_images.sh
matancarmeli7: why keep it if we don't use it?

scripts/ci/jenkins_pipeline_csi
matancarmeli7: done

scripts/ci/jenkins_pipeline_csi
oriyarde: but here we use artifactory only as a docker registry, and this pipeline code should work with non-artifactory registries as well.
so I'm not sure I follow. could you please clarify why `artifactory` is preferred over `registry`?

scripts/ci/jenkins_pipeline_csi
oriyarde: but we already have enough stages, and these stages are simple and rarely change.
@ArbelNathan what would you prefer?

scripts/ci/jenkins_pipeline_csi
matancarmeli7: I didn't mean to our local artifactory, I meant to the artifactory plugin that shows us at the end of the run the docker images that the pipeline created

scripts/ci/build_push_images.sh
oriyarde: we would use it in my 1st suggestion (as can be seen in the `sh` line).
both options I raised have these benefits:
1. reduces PR diff size (decreasing the chance of mistakes, e.g. [losing checks](https://github.com/IBM/ibm-block-csi-driver/pull/245#discussion_r507626972))
2. keeps logic outside the pipeline (makes it all more modular, reusable and in smaller, more manageable files).

the 1st option also has this benefit:
3. keeps logic in bash (a native cross-platform language that people are more familiar with - easier to maintain and more portable).

now that I answered your "why", could we go back to my question - could we do this?

scripts/ci/jenkins_pipeline_csi
oriyarde: you mean [this plugin](https://plugins.jenkins.io/artifactory/)? what does it have to do with [archiveArtifacts](https://www.jenkins.io/doc/pipeline/steps/core/) (which you're [not using](https://github.com/IBM/ibm-block-csi-driver/blob/ccc95bf4661a7b421b19e0b395faee03a33ec0ea/scripts/ci/jenkins_pipeline_csi#L86) with `artifactory_path` anyway)?

scripts/ci/jenkins_pipeline_csi
```diff
@@ -1,4 +1,12 @@
 pipeline {
+    environment {
+        registryCredential = 'csi_w3_user'
+        artifactory_path = 'build/reports/images_url'
+        registry_protocol = 'https'
+        Dockerfile_controller = 'Dockerfile-csi-controller'
+        Dockerfile_node = 'Dockerfile-csi-node'
+        latest_tag = "develop origin/develop master"
```
oriyarde, line 8: why not just keep a boolean var (like the original code)?
i.e. something like `latest_tag = "develop origin/develop master".contains(GIT_BRANCH)`

scripts/ci/jenkins_pipeline_csi
```diff
@@ -25,9 +33,47 @@ pipeline {
...
-        stage ('CSI-controller & node: Build and push images') {
+        stage ('Build docker images') {
+            steps {
+                script {
+                    sh 'bash ./scripts/ci/run_variables_verification.sh'
+                    branch = sh(script:"echo $GIT_BRANCH | sed 's|/|.|g'",returnStdout:true).trim()
+                    specific_tag = "${IMAGE_VERSION}_b${BUILD_NUMBER}_${branch}"
+                    if (latest_tag.contains(GIT_BRANCH)) {
+                        controller_image_latest = docker.build("${CSI_CONTROLLER_IMAGE}", "--build-arg VERSION=${IMAGE_VERSION} --build-arg BUILD_NUMBER=${BUILD_NUMBER} -f ${Dockerfile_controller} .")
+                        node_image_latest = docker.build("${DOCKER_REGISTRY}/${CSI_NODE_IMAGE}", "--build-arg VERSION=${IMAGE_VERSION} --build-arg BUILD_NUMBER=${BUILD_NUMBER} -f ${Dockerfile_node} .")
```
oriyarde, line 44: why not keep "partial" paths (e.g. `${DOCKER_REGISTRY}/${CSI_NODE_IMAGE}`) in a variable (like the original code did, e.g. with `node_registry`) which avoids duplicating the value?

scripts/ci/build_push_images.sh
oriyarde: this "done" raises a lot of questions, to name a few:

- where did `#!/bin/bash -xe` go?
- where did the newline at the EOF go?
- how can one understand the context of this file/variables from the `run_variables_verification.sh` name?

scripts/ci/jenkins_pipeline_csi
```diff
@@ -25,9 +33,47 @@ pipeline {
...
         }
-        stage ('CSI-controller & node: Build and push images') {
+        stage ('Build docker images') {
+            steps {
+                script {
+                    sh 'bash ./scripts/ci/run_variables_verification.sh'
+                    branch = sh(script:"echo $GIT_BRANCH | sed 's|/|.|g'",returnStdout:true).trim()
+                    specific_tag = "${IMAGE_VERSION}_b${BUILD_NUMBER}_${branch}"
+                    if (latest_tag.contains(GIT_BRANCH)) {
+                        controller_image_latest = docker.build("${CSI_CONTROLLER_IMAGE}", "--build-arg VERSION=${IMAGE_VERSION} --build-arg BUILD_NUMBER=${BUILD_NUMBER} -f ${Dockerfile_controller} .")
```
oriyarde, line 43: do we have to have separate build commands for latest tags? I don't think the original code had

scripts/ci/jenkins_pipeline_csi
```diff
@@ -25,9 +33,47 @@ pipeline {
...
+            }
+        }
+        stage ('push docker images') {
             steps {
-                sh './scripts/ci/build_push_images.sh build/reports/images_url'
+                script {
+                    registryUrl = "${registry_protocol}://${DOCKER_REGISTRY}"
+                    docker.withRegistry(registryUrl, registryCredential ) {
+                        if (latest_tag.contains(GIT_BRANCH)) {
+                            controller_image_latest.push()
```
oriyarde, line 57: if the `latest` tags push fails, will the stage stop and exit with failure? because I think the original code avoided that on purpose

scripts/ci/jenkins_pipeline_csi
```diff
@@ -25,9 +33,47 @@ pipeline {
...
+            }
+        }
+        stage ('Save Artifacts') {
+            steps {
+                script {
+                    sh 'dirname ${artifactory_path} | xargs mkdir -p'
+                    sh "echo ${DOCKER_REGISTRY}/${CSI_CONTROLLER_IMAGE}:$specific_tag > ${artifactory_path}"
+                    sh "echo ${DOCKER_REGISTRY}/${CSI_NODE_IMAGE}:$specific_tag >> ${artifactory_path}"
+                    if (latest_tag.contains(GIT_BRANCH)) {
+                        sh "echo ${DOCKER_REGISTRY}/${CSI_CONTROLLER_IMAGE}:latest >> ${artifactory_path}"
```
oriyarde, line 73: I don't think the original code saved the latest tags into a file.
why was this added?

scripts/ci/jenkins_pipeline_csi
```diff
@@ -25,9 +33,47 @@ pipeline {
...
+                    node_image_with_generated_tag = docker.build("${DOCKER_REGISTRY}/${CSI_NODE_IMAGE}:$specific_tag", "--build-arg VERSION=${IMAGE_VERSION} --build-arg BUILD_NUMBER=${BUILD_NUMBER} -f ${Dockerfile_node} .")
+                }
+            }
+        }
+        stage ('push docker images') {
             steps {
-                sh './scripts/ci/build_push_images.sh build/reports/images_url'
+                script {
+                    registryUrl = "${registry_protocol}://${DOCKER_REGISTRY}"
+                    docker.withRegistry(registryUrl, registryCredential ) {
```
oriyarde, line 55: is the space in `registryCredential )` on purpose?

scripts/ci/jenkins_pipeline_csi
```diff
@@ -25,9 +33,47 @@ pipeline {
...
+                    branch = sh(script:"echo $GIT_BRANCH | sed 's|/|.|g'",returnStdout:true).trim()
+                    specific_tag = "${IMAGE_VERSION}_b${BUILD_NUMBER}_${branch}"
+                    if (latest_tag.contains(GIT_BRANCH)) {
+                        controller_image_latest = docker.build("${CSI_CONTROLLER_IMAGE}", "--build-arg VERSION=${IMAGE_VERSION} --build-arg BUILD_NUMBER=${BUILD_NUMBER} -f ${Dockerfile_controller} .")
+                        node_image_latest = docker.build("${DOCKER_REGISTRY}/${CSI_NODE_IMAGE}", "--build-arg VERSION=${IMAGE_VERSION} --build-arg BUILD_NUMBER=${BUILD_NUMBER} -f ${Dockerfile_node} .")
+                    }
+                    controller_image_with_generated_tag = docker.build("${DOCKER_REGISTRY}/${CSI_CONTROLLER_IMAGE}:$specific_tag", "--build-arg VERSION=${IMAGE_VERSION} --build-arg BUILD_NUMBER=${BUILD_NUMBER} -f ${Dockerfile_controller} .")
+                    node_image_with_generated_tag = docker.build("${DOCKER_REGISTRY}/${CSI_NODE_IMAGE}:$specific_tag", "--build-arg VERSION=${IMAGE_VERSION} --build-arg BUILD_NUMBER=${BUILD_NUMBER} -f ${Dockerfile_node} .")
```
oriyarde, line 47: in the original code, if the node build would fail, we would at least have the controller image available to pull.
why should we change that?

scripts/ci/jenkins_pipeline_csi
```diff
@@ -1,4 +1,12 @@
 pipeline {
+    environment {
+        registryCredential = 'csi_w3_user'
+        artifactory_path = 'build/reports/images_url'
+        registry_protocol = 'https'
```
oriyarde, line 5: do we expect this value ever to change/be used in more than 1 place?
if not, consider using the value directly without an additional variable.

scripts/ci/build_push_images.sh
matancarmeli7: I don't think that we should, when you are using Jenkinsfile you want to use the Jenkins plugins. 
When there is an option to use a Jenkins plugin, it's better to use it, because it gives you the option of better error handling and in my opinion more readable code, the plugin's syntax is pretty simple on most of the time and more readable.
Also, I think that we should avoid using shell scripts in Jenkinsfile instead of plugins for the same reason that you try to avoid shell module in Ansible and try to use builtin modules if you can.

scripts/ci/jenkins_pipeline_csi
matancarmeli7: It's supposed to be used by the archiveArtifacts, but it wasn't in the scope of my task, I will open a ticket to go through the pipelines and to fix some issues that in there (I saw a few).

scripts/ci/build_push_images.sh
matancarmeli7: to which name do you suggest that I will change it?

scripts/ci/jenkins_pipeline_csi
matancarmeli7: in the original code, if the node build fails you won't know about it, the pipeline can finish in succeded mode and still don't push the node image.

scripts/ci/jenkins_pipeline_csi
matancarmeli7: I don't think that we will change it but maybe it would change.
I don't think that it's harmful to use this variable, but if you still think that you prefer to use it directly I will change it.

scripts/ci/jenkins_pipeline_csi
matancarmeli7: I thought that it would be a nice extra, if you think that it's not necessary, I can erase it.

scripts/ci/jenkins_pipeline_csi
matancarmeli7: I didn't find a way to do it in Jenkins

scripts/ci/jenkins_pipeline_csi
matancarmeli7: done

scripts/ci/jenkins_pipeline_csi
matancarmeli7: That how Jenkins works, you can't build the same image in the command with two tagges, I can add a docker tag instead of two docker build for each image. And I think that we should keep the variable even if we will have one build command, because it will be easier to change it if the Dockerfile path will change.

scripts/ci/jenkins_pipeline_csi
matancarmeli7: In the original code, the run will be green even If the latest tags push will fail or any push or build as long as the last command will succeded.

scripts/ci/jenkins_pipeline_csi
matancarmeli7: fixed it
done

scripts/ci/jenkins_pipeline_csi
oriyarde: I think you are wrong - not **any** push or build fail will result in a green run, only `latest` (or failure to save tags to file).
could you prove your statement? what makes you think that?

scripts/ci/jenkins_pipeline_csi
oriyarde: I think it's not necessary

scripts/ci/jenkins_pipeline_csi
oriyarde: 1. I disagree - did you test your statement?
2. I don't see how this is related to this conversation.

- if you think my statement is wrong, please elaborate.
- if you agree with my statement, please answer "why should we change that?".

scripts/ci/jenkins_pipeline_csi
oriyarde: I still think I prefer to use it directly

scripts/ci/jenkins_pipeline_csi
oriyarde: why did you decide to no longer use the `${DOCKER_REGISTRY}` value in this line?

scripts/ci/jenkins_pipeline_csi
oriyarde: what do you mean "how Jenkins works"?
can't you just build once (i.e. `controller_image = docker.build(...)`) and then push two tags (i.e. `controller_image.push("${controller_registry}:$specific_tag")` and `controller_image.push("${controller_registry}:latest`)?

scripts/ci/build_push_images.sh
oriyarde: you only (partially) addressed the 3rd benefit (and thus only to my 1st option) - please address as much options and arguments as possible.

why would the plugin give better error handling? could you elaborate?

I think you are ignoring the limitations of the plugin's "simple" syntax - most of our logic cannot be used with it (env vars validation, sed, saving a tags file, etc...) and so our logic becomes separated into different languages and places, coupled to jenkins, with no real added value.

scripts/ci/jenkins_pipeline_csi
oriyarde: this does not answer my questions, the lack of `artifactory_path` usage by `archiveArtifacts` was only a side note.
could you please clarify what does `build/reports/images_url` has to do with `artifactory`?

scripts/ci/build_push_images.sh
oriyarde: I assume `verify_env_vars_for_build_push.sh` would add more context, but this is just 1 of the many bad symptoms of the unnecessary separation of our code.

scripts/ci/jenkins_pipeline_csi
ArbelNathan: agreed that its to many stages, the debugging process wont change anyway due to fact that the same log will be written minus the the stage name which is too much of a small value for that clutter of stages

scripts/ci/jenkins_pipeline_csi
oriyarde: what do you mean by "find"? what's wrong with my suggestion?
and as I wrote, if this value remains, then at least the name should be changed (to e.g. `latest_tag_branches`).

scripts/ci/jenkins_pipeline_csi
matancarmeli7: because in bash script it doesn't stop if there is an error in one of the lines. In addition, the bash script will exit with exit code 0 if the last command of the script succeeded even when all other commands in the script failed.
To prove it I run a test pipeline that there are two Docker build commands in it' the first one will fail and the last one will succeed. The pipeline resulted in green.

scripts/ci/jenkins_pipeline_csi
matancarmeli7: because this is not the syntax this plugin works, I checked in a lot of places and everyone said that in order to build one image with two tags in Jenkins, the only way is how I did it. 
You can't do:
`controller_image.push("${controller_registry}:$specific_tag")
controller_image.push("${controller_registry}:latest")`

I can try and check it again

scripts/ci/jenkins_pipeline_csi
oriyarde: let me know when the proof is ready for me to see

scripts/ci/jenkins_pipeline_csi
oriyarde: do check again, and please add your links here.
this is what I found:
https://stackoverflow.com/a/53858257

scripts/ci/jenkins_pipeline_csi
```diff
@@ -1,4 +1,12 @@
 pipeline {
+    environment {
+        registryCredential = 'csi_w3_user'
```
oriyarde, line 3: `registryCredentialsID` is more accurate

matancarmeli7: done

scripts/ci/jenkins_pipeline_csi
matancarmeli7: done

scripts/ci/jenkins_pipeline_csi
matancarmeli7: I was wrong, I work now with the original bash file

scripts/ci/jenkins_pipeline_csi
matancarmeli7: done

scripts/ci/jenkins_pipeline_csi
matancarmeli7: I was wrong, I work now with the original bash file

scripts/ci/jenkins_pipeline_csi
matancarmeli7: done

scripts/ci/jenkins_pipeline_csi
matancarmeli7: done

scripts/ci/jenkins_pipeline_csi
matancarmeli7: done

scripts/ci/build_push_images.sh
matancarmeli7: removed it, done

scripts/ci/jenkins_pipeline_csi
matancarmeli7: done

scripts/ci/jenkins_pipeline_csi
matancarmeli7: removed it, done

oriyarde (approved):
LGTM

PR #247, opened 2020-10-14
Add ControllerPublishVolume and Node community tests

lisyukanna:


scripts/csi_test/csi_tests_to_run
```diff
@@ -12,3 +12,5 @@ NodeUnstageVolume
 GetPluginCapabilities
 GetPluginInfo
 Probe
+ControllerPublishVolume
+Node
```
ArbelNathan, line 16: what is "Node"?

lisyukanna: in this task there was a request to add 2 tests: ControllerPublishVolume and Node Service

oriyarde: will it work if `Node` be `Node Service` instead?

scripts/csi_test/csi_secrets
```diff
@@ -63,3 +63,9 @@ NodeUnstageVolumeSecret:
  username: USERNAME
  password: PASSWORD
  alias_in_storageclass: arr1
+
+ControllerPublishVolumeSecret:
```
oriyarde, line 67: what about ControllerUnpublishVolume?

scripts/csi_test/csi_tests_to_run
lisyukanna: why Node be Node Service?

scripts/csi_test/csi_secrets
lisyukanna: adding ControllerUnpublishVolume was not defined as part of this ticket, if adding ControllerUnpublishVolume will require minor changes I will add it as part of this task otherwise I will open a new ticket

scripts/csi_test/csi_tests_to_run
oriyarde: this is the "tests to run" file, and I got the impression that these tests are the "Node Service" tests.
so if I understand correctly, the reason is to have the full name, which would be more clear.

scripts/csi_test/csi_secrets
oriyarde: ok. btw I think the same goes for two "volume lifecycle" tests in "Controller Service"

lisyukanna: "ControllerUnpublishVolume" test was added, for "volume lifecycle" tests a ticket was opened

scripts/csi_test/csi_tests_to_run
lisyukanna: done

scripts/csi_test/csi_secrets
oriyarde: thanks for the addition and for opening CSI-2155!

scripts/csi_test/csi_tests_to_run
oriyarde: @ArbelNathan since you opened this conversation, please resolve it once you feel it's resolved

oriyarde (approved):
Nice job!

PR #249, opened 2020-11-01
add fix in the docker run that solved a problem with the openssl version in power

matancarmeli7:


ArbelNathan (dismissed):
looks good

oriyarde (changes requested):
1. do we really want to run our tests with an unsupported lib? it might cause issues
2. should we really "patch" for every specific version of old builder libs?
as the error implies, I think the correct fix here is to upgrade the builder lib instead - is there a reason not to?
3. if we still must patch, why not put it in the `Dockerfile-csi-controller.test` (similar to https://github.com/IBM/ibm-block-csi-driver/pull/120)? `run_unitests.sh` is only used in CI, but it would be better if manual local PowerPC runs would also work (i.e. not necessarily using `run_unitests.sh`)

matancarmeli7:
I think that it's good for now because the CI was broke in power slaves, I agree that it's not the ideal problem, so I opened a ticket to fix it by upgrading the builder lib.

oriyarde:
> I think that it's good for now because the CI was broke in power slaves, I agree that it's not the ideal problem, so I opened a ticket to fix it by upgrading the builder lib.

@matancarmeli7 then please add a comment `TODO: CSI-<number> remove env var` in the script, and ensure that ticket description contains this removal

and please also address my [3rd question](https://github.com/IBM/ibm-block-csi-driver/pull/249#pullrequestreview-521267033).

oriyarde (approved):

matancarmeli7:
I am testing the changes, if it will work I will merge it

ArbelNathan (approved):
lgtm

PR #250, opened 2020-11-03
Feature/csi 2339 update csi pb2

ArbelNathan:
updating csi_pb2 files and adding the command


controller/scripts/csi_pb2.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -x
+
+version="$1" # currently k8s on v1.2.0
+curl -o csi.proto https://raw.githubusercontent.com/container-storage-interface/spec/"$version"/csi.proto
+python -m grpc_tools.protoc --proto_path=. --grpc_python_out=./controller/csi_general --python_out=./controller/csi_general csi.proto
```
oriyarde, line 6: image
I'm starting to think we need to add a linter for [these gaps](https://github.com/IBM/ibm-block-csi-operator/pull/103#discussion_r507624526)

controller/scripts/csi_pb2.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -x
+
+version="$1" # currently k8s on v1.2.0
```
oriyarde, line 4: then let's keep the version value hardcoded for now, because it is important info, and comments are more likely to be removed

oriyarde (changes requested):

controller/scripts/csi_pb2.sh
ArbelNathan: done.
its not tested so if you miss it its hard to see it later..

controller/scripts/csi_pb2.sh
ArbelNathan: done

oriyarde (approved):
nice!

lisyukanna (approved):
LGTM

PR #212, opened 2020-07-22
Feature/csi 1601 implement validate volume capabilities

ArbelNathan:

PR #251, opened 2020-11-09
Feature/csi 2121 volume expansion controller server and interface

ArbelNathan:


controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -1195,6 +1199,156 @@ def test_controller_get_capabilities(self):
...
+    @patch("controller.controller_server.csi_controller_server.get_agent")
+    def test_expand_volume_with_array_connection_exception(self, storage_agent):
+        self._test_create_object_with_array_connection_exception(storage_agent)
+
+    @patch("controller.controller_server.csi_controller_server.get_agent")
+    def test_expand_volume_with_get_volume_name_too_long_success(self, storage_agent):
+        self._prepare_expand_volume_mocks(storage_agent)
+        self.mediator.max_volume_name_length = 63
+
+        self.request.name = "a" * 128
```
oriyarde, line 1321: where is this being used?

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -1195,6 +1199,156 @@ def test_controller_get_capabilities(self):
...
+        self.servicer.ControllerExpandVolume(self.request, self.context)
+        self.assertEqual(self.context.code, grpc.StatusCode.OK)
+
+    @patch("controller.array_action.array_mediator_xiv.XIVArrayMediator.expand_volume")
```
oriyarde, line 1325: try to move this to `setUp`/`_prepare...`

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -1195,6 +1199,156 @@ def test_controller_get_capabilities(self):
...
+        self.assertTrue(expand_respond.node_expansion_required)
+        self.mediator.expand_volume.assert_called_once_with(volume_id=self.volume_id,
+                                                            required_bytes=self.capacity_bytes)
+
+    @patch("controller.controller_server.csi_controller_server.get_agent")
+    def test_expand_volume_with_wrong_secrets(self, a_enter):
+        self._test_create_object_with_wrong_secrets(a_enter)
+
+    @patch("controller.controller_server.csi_controller_server.get_agent")
+    def test_expand_volume_with_wrong_volume_capabilities(self, storage_agent):
```
oriyarde, line 1293: split this into two unit tests (something like `test_expand_volume_with_invalid_fstype_fail` and `test_expand_volume_with_invalid_access_mode_fail`)

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -1195,6 +1199,156 @@ def test_controller_get_capabilities(self):
...
+    def test_expand_volume_with_zero_size(self, storage_agent):
+        self._prepare_expand_volume_mocks(storage_agent)
+
+        self.request.capacity_range.required_bytes = 0
+        self.servicer.ControllerExpandVolume(self.request, self.context)
+        self.assertEqual(self.context.code, grpc.StatusCode.OK)
+        self.mediator.expand_volume.assert_called_once_with(volume_id=self.volume_id, required_bytes=0)
+
+    @patch("controller.controller_server.csi_controller_server.get_agent")
+    def test_expand_volume_with_big_size(self, storage_agent):
```
oriyarde, line 1256: `test_expand_volume_with_size_too_large_fail`?

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -1195,6 +1199,156 @@ def test_controller_get_capabilities(self):
...
+        self.fqdn = "fqdn"
+        self.mediator = XIVArrayMediator("user", "password", self.fqdn)
+        self.mediator.client = Mock()
+
+        self.storage_agent = MagicMock()
+        self.storage_agent.get_mediator.return_value.__enter__.return_value = self.mediator
+
+        self.servicer = ControllerServicer(self.fqdn)
+
+        self.request = Mock()
```
oriyarde, line 1221: can't these (+ secrets) move higher up the `setUp` hierarchy?

controller/controller_server/csi_controller_server.py
```diff
@@ -534,6 +539,83 @@ def GetCapacity(self, request, context):
...
+                context.set_code(grpc.StatusCode.INVALID_ARGUMENT)
+                return csi_pb2.ControllerExpandVolumeResponse()
+
+            with get_agent(user, password, array_addresses, array_type).get_mediator() as array_mediator:
+                logger.debug(array_mediator)
+
+                size = request.capacity_range.required_bytes
+
+                if size > array_mediator.maximal_volume_size_in_bytes:
+                    raise controller_errors.SizeOutOfRange(size, array_mediator.maximal_volume_size_in_bytes)
```
oriyarde, line 565: do we really need to add a new type of exception just for this line? why not simply return `OUT_OF_RANGE` response here?

controller/array_action/messages.py
```diff
@@ -62,3 +62,7 @@
 ObjectIsStillInUseError_message = "Object {0} is still in use by {1}"
 
 InvalidCliResponseError_message = "Invalid CLI response. Details : {0}"
+
+NotEnoughSpaceInPoolError_message = "Not enough space in pool"
```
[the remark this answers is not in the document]
ArbelNathan: I dont have it..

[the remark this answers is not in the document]
oriyarde: what do you mean? why can't a mediator fill the pool value?

controller/controller_server/csi_controller_server.py
```diff
@@ -534,6 +539,83 @@ def GetCapacity(self, request, context):
...
+        try:
+            utils.validate_expand_volume_request(request)
+
+            user, password, array_addresses = utils.get_array_connection_info_from_secret(secrets)
+
+            try:
+                array_type, volume_id = utils.get_volume_id_info(request.volume_id)
+            except ObjectIdError as ex:
+                logger.warning("volume id is invalid. error : {}".format(ex))
+                context.set_code(grpc.StatusCode.INVALID_ARGUMENT)
```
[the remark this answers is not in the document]
ArbelNathan: but in that case, what are suggesting to do?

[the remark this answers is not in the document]
oriyarde: return a different non-OK status code? e.g. `NOT_FOUND`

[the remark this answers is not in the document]
ArbelNathan: btw,according to [spec](https://github.com/container-storage-interface/spec/blob/v1.3.0/spec.md#error-scheme): Indicates that the one or more fields in this field is either not allowed by the Plugin or has an invalid value.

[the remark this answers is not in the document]
oriyarde:
```
Each RPC defines a set of gRPC error codes that MUST be returned by the plugin when specified conditions are encountered. In addition to those, if the conditions defined below are encountered, the plugin MUST return the associated gRPC error code.
```
interesting, so a status code should be used where it's mentioned in the spec for both specific method scenarios and general scenarios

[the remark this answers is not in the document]
ArbelNathan: so we can close this conversation?

controller/tests/controller_server/csi_controller_server_test.py
ArbelNathan: its i only one place and its kind of prepare one (several method using it). I changing it to private

controller/controller_server/csi_controller_server.py
ArbelNathan: no.
done

controller/tests/controller_server/csi_controller_server_test.py
ArbelNathan: yep
done

controller/tests/controller_server/csi_controller_server_test.py
ArbelNathan: seems better.
done

controller/tests/controller_server/csi_controller_server_test.py
ArbelNathan: ok
done

controller/array_action/array_mediator_interface.py
```diff
@@ -131,6 +131,23 @@ def get_volume_name(self, volume_id):
...
 
+    @abstractmethod
+    def expand_volume(self, volume_id, required_bytes):
+        """
+        This expand the volume size on storage.
+        Args:
+           volume_id        : volume id
+           required_bytes   : capacity requirements of the volume after expansion
+        Returns:
+           volume
```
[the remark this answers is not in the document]
ArbelNathan: done

controller/controller_server/csi_controller_server.py
```diff
@@ -534,6 +539,83 @@ def GetCapacity(self, request, context):
...
+
+                if size > array_mediator.maximal_volume_size_in_bytes:
+                    raise controller_errors.SizeOutOfRange(size, array_mediator.maximal_volume_size_in_bytes)
+
+                try:
+
+                    logger.debug("expanding volume {0}".format(volume_id))
+                    array_mediator.expand_volume(
+                        volume_id=volume_id,
+                        required_bytes=request.capacity_range.required_bytes)
```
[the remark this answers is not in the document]

controller/controller_server/csi_controller_server.py
```diff
@@ -534,6 +539,83 @@ def GetCapacity(self, request, context):
...
+                    res = utils.generate_csi_expand_volume_response(request)
+                    logger.info("finished expanding volume")
+                    return res
+
+                except controller_errors.PermissionDeniedError as ex:
```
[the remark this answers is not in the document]

controller/controller_server/csi_controller_server.py
```diff
@@ -534,6 +539,83 @@ def GetCapacity(self, request, context):
...
+            context.set_code(grpc.StatusCode.RESOURCE_EXHAUSTED)
+            return csi_pb2.ControllerExpandVolumeResponse()
+
+        except controller_errors.SizeOutOfRange as ex:
+            logger.exception(ex)
+            context.set_details(ex.message)
+            context.set_code(grpc.StatusCode.OUT_OF_RANGE)
+            return csi_pb2.ControllerExpandVolumeResponse()
+
+        except (controller_errors.ObjectIsStillInUseError, Exception) as ex:
```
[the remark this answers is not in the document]

controller/controller_server/utils.py
```diff
@@ -158,6 +158,32 @@ def validate_delete_snapshot_request(request):
     logger.debug("request validation finished.")
 
 
+def validate_expand_volume_request(request):
+    logger.debug("validating expand volume request")
+
+    if request.volume_id == "":
+        raise ValidationException("Volume id cannot be empty")
```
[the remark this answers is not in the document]

controller/controller_server/utils.py
```diff
@@ -158,6 +158,32 @@ def validate_delete_snapshot_request(request):
...
+
+    logger.debug("validating volume capacity")
+    if request.capacity_range:
+        if request.capacity_range.required_bytes < 0:
+            raise ValidationException(messages.size_bigger_than_0_message)
+    else:
+        raise ValidationException(messages.no_capacity_range_message)
+
+    logger.debug("validating secrets")
+    if request.secrets:
```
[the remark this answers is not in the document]

controller/controller_server/utils.py
```diff
@@ -158,6 +158,32 @@ def validate_delete_snapshot_request(request):
...
+        validate_secret(request.secrets)
+
+    logger.debug("validating volume capabilities")
+    capability = request.volume_capability
```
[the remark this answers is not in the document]

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -1195,6 +1199,156 @@ def test_controller_get_capabilities(self):
...
+        self.mediator.expand_volume.assert_called_once_with(volume_id=self.volume_id, required_bytes=0)
+
+    @patch("controller.controller_server.csi_controller_server.get_agent")
+    def test_expand_volume_with_big_size(self, storage_agent):
+        self._prepare_expand_volume_mocks(storage_agent)
+
+        self.request.capacity_range.required_bytes = 1125899906842625
+        self.servicer.ControllerExpandVolume(self.request, self.context)
+        self.assertEqual(self.context.code, grpc.StatusCode.OUT_OF_RANGE)
+        self.mediator.expand_volume.assert_not_called()
```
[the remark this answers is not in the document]
ArbelNathan: its not, its a fixed value taken from the XIV mediator

controller/controller_server/utils.py
```diff
@@ -201,6 +227,17 @@ def generate_csi_create_snapshot_response(new_snapshot, source_volume_id):
     return res
 
 
+def generate_csi_expand_volume_response(request):
+    logger.debug("creating response for expand volume")
+    res = csi_pb2.ControllerExpandVolumeResponse(
+        capacity_bytes=request.capacity_range.required_bytes,
```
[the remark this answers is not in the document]

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -1195,6 +1199,156 @@ def test_controller_get_capabilities(self):
...
+
+    @patch("controller.controller_server.csi_controller_server.get_agent")
+    def test_expand_volume_succeeds_mount_access_mode(self, storage_agent):
+        self._prepare_expand_volume_mocks(storage_agent)
+
+        expand_respond = self.servicer.ControllerExpandVolume(self.request, self.context)
```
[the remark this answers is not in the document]

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -1195,6 +1199,156 @@ def test_controller_get_capabilities(self):
...
+        self.servicer.ControllerExpandVolume(self.request, self.context)
+        self.assertEqual(self.context.code, grpc.StatusCode.INVALID_ARGUMENT, "wrong fs_type")
+        self.assertTrue("fs_type" in self.context.details)
+
+        self.request.volume_capability = csi_pb2.VolumeCapability(
+            access_mode=csi_pb2.VolumeCapability.AccessMode(mode=self.access_types.MULTI_NODE_SINGLE_WRITER),
+            mount=csi_pb2.VolumeCapability.MountVolume(fs_type="ext4"))
+
+        self.servicer.ControllerExpandVolume(self.request, self.context)
+        self.assertEqual(self.context.code, grpc.StatusCode.INVALID_ARGUMENT, "wrong access_mode")
```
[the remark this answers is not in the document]

controller/tests/controller_server/csi_controller_server_test.py
ArbelNathan: that test is not relevant and I removed it completely

controller/array_action/array_mediator_ds8k.py
```diff
@@ -122,6 +122,10 @@ def max_snapshot_prefix_length(self):
     def minimal_volume_size_in_bytes(self):
         return 512  # 1 block, 512 bytes
 
+    @classproperty
+    def maximal_volume_size_in_bytes(self):
+        return 17592186044416  # 16 TB
```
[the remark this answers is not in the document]

controller/array_action/array_mediator_interface.py
```diff
@@ -131,6 +131,23 @@ def get_volume_name(self, volume_id):
         """
         raise NotImplementedError
 
+    @abstractmethod
+    def expand_volume(self, volume_id, required_bytes):
+        """
+        This expand the volume size on storage.
+        Args:
+           volume_id        : volume id
+           required_bytes   : capacity requirements of the volume after expansion
```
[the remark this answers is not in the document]
ArbelNathan: done

controller/array_action/array_mediator_interface.py
```diff
@@ -406,6 +424,14 @@ def minimal_volume_size_in_bytes(self):
         """
         raise NotImplementedError
 
+    @property
+    @abstractmethod
+    def maximal_volume_size_in_bytes(self):
+        """
+        The maximal volume size in bytes.
```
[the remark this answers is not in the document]
ArbelNathan: done

controller/array_action/errors.py
```diff
@@ -206,3 +206,15 @@ class InvalidCliResponseError(BaseArrayActionException):
...
+class NotEnoughSpaceInPool(BaseArrayActionException):
+
+    def __init__(self):
+        self.message = messages.NotEnoughSpaceInPoolError_message
+
+
+class SizeOutOfRange(BaseArrayActionException):
+
+    def __init__(self, requested_size, limit):
+        self.message = messages.SiseOutOfRangeError_message.format(requested_size, limit)
```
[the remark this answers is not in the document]
ArbelNathan: done

controller/array_action/messages.py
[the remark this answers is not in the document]
ArbelNathan: sure, but only to pass the pool parameter to the mediators for log purposes seems too much

controller/array_action/messages.py
```diff
@@ -62,3 +62,7 @@
...
+
+SiseOutOfRangeError_message = "requested size is above limit. requested: {0}, limit is: {1}"
```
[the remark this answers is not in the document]
ArbelNathan: done

controller/controller_server/utils.py
```diff
@@ -158,6 +158,32 @@ def validate_delete_snapshot_request(request):
     logger.debug("request validation finished.")
 
 
+def validate_expand_volume_request(request):
+    logger.debug("validating expand volume request")
+
+    if request.volume_id == "":
```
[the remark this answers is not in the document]

controller/controller_server/csi_controller_server.py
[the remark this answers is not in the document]
oriyarde: yes

controller/tests/controller_server/csi_controller_server_test.py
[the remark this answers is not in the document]
oriyarde: then please mock it to decouple it from XIV mediator implementation

controller/array_action/array_mediator_ds8k.py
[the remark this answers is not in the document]
oriyarde: it's still `TiB` (not `TB`) - either fix the comment or remove it

controller/array_action/messages.py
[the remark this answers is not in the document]
oriyarde: why would the pool parameter be passed to the mediators? the controller server doesn't have the pool parameter

controller/controller_server/csi_controller_server.py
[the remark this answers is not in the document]
oriyarde: why `FAILED_PRECONDITION`? I thought `ObjectIsStillInUseError` is for still being used by a flashcopy mapping, not a worker node (since we support online expansion).

controller/controller_server/utils.py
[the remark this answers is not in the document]
oriyarde: **similar** (id is not a name)

controller/controller_server/utils.py
```diff
@@ -158,6 +158,29 @@ def validate_delete_snapshot_request(request):
...
+        if request.capacity_range.required_bytes < 0:
+            raise ValidationException(messages.size_bigger_than_0_message)
+    else:
+        raise ValidationException(messages.no_capacity_range_message)
+
+    validate_secret(request.secrets)
+
+    logger.debug("validating volume capabilities")
+    if not request.volume_capability:
+        raise ValidationException(messages.capabilities_not_set_message)
```
oriyarde, line 178: I thought `This is an OPTIONAL field.`

controller/controller_server/csi_controller_server.py
```diff
@@ -534,6 +539,81 @@ def GetCapacity(self, request, context):
...
+                    context.set_details(message)
+                    context.set_code(grpc.StatusCode.OUT_OF_RANGE)
+                    return csi_pb2.ControllerExpandVolumeResponse()
+
+                logger.debug("expanding volume {0}".format(volume_id))
+                array_mediator.expand_volume(
+                    volume_id=volume_id,
+                    required_bytes=size)
+
+                volume = array_mediator.get_object_by_id(volume_id, config.VOLUME_TYPE_NAME)
```
oriyarde, line 576: wouldn't it be more efficient if `expand_volume` would just return the volume (as was initially documented in the mediator interface)?

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -1264,6 +1260,151 @@ def test_controller_get_capabilities(self):
...
+    def test_expand_volume_with_zero_size(self, storage_agent):
+        self._prepare_expand_volume_mocks(storage_agent)
+
+        self.request.capacity_range.required_bytes = 0
+        self.servicer.ControllerExpandVolume(self.request, self.context)
+        self.assertEqual(self.context.code, grpc.StatusCode.OK)
+        self.mediator.expand_volume.assert_called_once_with(volume_id=self.volume_id, required_bytes=0)
+
+    @patch("controller.controller_server.csi_controller_server.get_agent")
+    def test_expand_volume_with_size_too_large_fail(self, storage_agent):
```
oriyarde, line 1317: what about size too small?

controller/tests/controller_server/csi_controller_server_test.py
oriyarde: I only see you moved two lines. what about most of the other lines in this function?

controller/array_action/messages.py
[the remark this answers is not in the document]
ArbelNathan: controller gets it in the parameters var in the request.
mediators dont have it and they are not getting the volume during the expansion so they cant query it.

controller/tests/controller_server/csi_controller_server_test.py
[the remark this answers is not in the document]
ArbelNathan: why?

controller/tests/controller_server/csi_controller_server_test.py
ArbelNathan: there is a massive change for it in [CSI-1601](https://github.com/IBM/ibm-block-csi-driver/blob/e3755778032f718e079b0392554acdb19836b9ba/controller/tests/controller_server/csi_controller_server_test.py#L19)

controller/tests/controller_server/csi_controller_server_test.py
ArbelNathan: we have a default for it

controller/controller_server/csi_controller_server.py
ArbelNathan: the mediators never got them.. at least not all of them

controller/array_action/array_mediator_ds8k.py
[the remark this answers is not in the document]

controller/controller_server/csi_controller_server.py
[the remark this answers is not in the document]
ArbelNathan: yesyes, my bad

controller/controller_server/utils.py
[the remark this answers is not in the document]
ArbelNathan: right, done

controller/tests/controller_server/csi_controller_server_test.py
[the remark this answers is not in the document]
ArbelNathan: done, but no mock as its a classproperty

controller/array_action/messages.py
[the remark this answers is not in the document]
oriyarde: then could you please link + quote the spec where it says the controller gets a "parameters" var in the request?
which mediators are not getting the volume during expansion and why can't they query it?

controller/controller_server/utils.py
ArbelNathan: removed it, didnt used capabilities

controller/controller_server/csi_controller_server.py
oriyarde: then wouldn't it be more efficient for those who did get them?

controller/tests/controller_server/csi_controller_server_test.py
oriyarde: where? and how does it justify not unit testing such scenario?

controller/tests/controller_server/csi_controller_server_test.py
oriyarde: ok

controller/controller_server/utils.py
oriyarde: ok, please note that it is still required for this feature, just easier to be done in a [separate PR](https://github.com/IBM/ibm-block-csi-driver/pull/212#pullrequestreview-528927603)

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -1264,6 +1260,131 @@ def test_controller_get_capabilities(self):
...
+        self._prepare_expand_volume_mocks(storage_agent)
+        self.mediator.maximal_volume_size_in_bytes = 1 * 1024 * 1024 * 1024 * 1024 * 1024
+        self.request.capacity_range.required_bytes = 1 * 1024 * 1024 * 1024 * 1024 * 1024 * 1024
```
oriyarde, line 1320: nice. mind using simpler values for debug?
the point here is to test simple logic (e.g. `2 > 1`), not live system tests with real values.

controller/tests/controller_server/csi_controller_server_test.py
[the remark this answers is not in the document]
oriyarde: ok, hope it won't affect other tests

controller/array_action/array_mediator_ds8k.py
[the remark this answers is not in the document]
oriyarde: same for `PB`...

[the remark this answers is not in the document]
oriyarde: this relatively long conversation is an example to why comments maintenance is not cost-affective

controller/array_action/messages.py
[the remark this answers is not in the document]
ArbelNathan: I looked again and the controller dont get it.
ds8k dont get it.
why should it get it if it dosent need it?

controller/tests/controller_server/csi_controller_server_test.py
ArbelNathan: 1. k8s wont let you expand less then current size
1 its in [CreateVolume](https://github.com/IBM/ibm-block-csi-driver/blob/ca8cf8c3f7eb8caec6a0337bda44941c13982549/controller/controller_server/csi_controller_server.py#L97)

controller/array_action/messages.py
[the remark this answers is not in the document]
ArbelNathan: ok, I will put it in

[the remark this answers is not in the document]
oriyarde: `it` might be confusing.
ds8k mediator should get the pool value because the controller server needs it, that's the purpose of return values.
so the ds8k mediator will have to get the pool value somehow eventually because that's what the `ControllerExpandVolume` flow requires (no matter how we arrange the code).

controller/tests/controller_server/csi_controller_server_test.py
oriyarde: we shouldn't rely on current k8s implementation details - the spec allows k8s to change that behavior, and it shouldn't break our code

controller/array_action/messages.py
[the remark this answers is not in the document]
ArbelNathan: done

controller/controller_server/csi_controller_server.py
ArbelNathan: done (working on mediators now)

controller/array_action/array_mediator_interface.py
```diff
@@ -131,6 +131,23 @@ def get_volume_name(self, volume_id):
...
 
+    @abstractmethod
+    def expand_volume(self, volume_id, required_bytes):
+        """
+        Expand the volume size on storage.
+        Args:
+           volume_id        : volume id
+           required_bytes   : capacity of the volume after expansion
+        Returns:
+           volume
```
oriyarde, line 142: Capital `V`

controller/array_action/array_mediator_ds8k.py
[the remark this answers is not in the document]
ArbelNathan: yes. ok removed.

controller/tests/controller_server/csi_controller_server_test.py
ArbelNathan: added validation and UT

controller/controller_server/csi_controller_server.py
```diff
@@ -534,6 +539,76 @@ def GetCapacity(self, request, context):
...
+                context.set_code(grpc.StatusCode.INVALID_ARGUMENT)
+                return csi_pb2.ControllerExpandVolumeResponse()
+
+            with get_agent(user, password, array_addresses, array_type).get_mediator() as array_mediator:
+                logger.debug(array_mediator)
+
+                size = request.capacity_range.required_bytes
+
+                if size > array_mediator.maximal_volume_size_in_bytes or \
+                        size < array_mediator.minimal_volume_size_in_bytes:
```
oriyarde, line 565: style suggestion:
```
min_size = array_mediator.minimal_volume_size_in_bytes
max_size = array_mediator.maximal_volume_size_in_bytes
if not min_size <= size <= max_size:
   ...
```
but only if you think it's better

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -1264,6 +1260,132 @@ def test_controller_get_capabilities(self):
...
+        self.mediator.expand_volume.return_value = utils.get_mock_mediator_response_volume(self.capacity_bytes,
+                                                                                           volume_name,
+                                                                                           self.volume_id,
+                                                                                           "a9k")
+
+    @patch("controller.controller_server.csi_controller_server.get_agent")
+    def test_expand_volume_with_size_too_small_fail(self, storage_agent):
+        self._prepare_expand_volume_mocks(storage_agent)
+
+        self.request.capacity_range.required_bytes = 0
```
oriyarde, line 1313: `A value of 0 is equal to an unspecified field value.`
I think we should return OK without expanding in this case.
do you?

ArbelNathan: Its only in create, I didnt saw that in expand too

oriyarde: ok, so do we interpret that `required_bytes` is mandatory for expand just because `CapacityRange` is `REQUIRED` there?
if so, does that mean we should interpret that `limit_bytes` is also mandatory?

controller/controller_server/csi_controller_server.py
ArbelNathan: it is looks better
done

controller/array_action/array_mediator_interface.py
ArbelNathan: now its None again

controller/tests/controller_server/csi_controller_server_test.py
ArbelNathan: good question, I ignored it like done in create, its not very informative.

oriyarde: 1. do you think it is correct for create to ignore it?
2. if we align expand and create behavior regarding `limit_bytes`, why shouldn't we do the same for `required_bytes` (i.e. `A value of 0 is equal to an unspecified field value.`)?

controller/controller_server/csi_controller_server.py
```diff
@@ -534,6 +539,76 @@ def GetCapacity(self, request, context):
...
+
+                logger.debug("expanding volume {0}".format(volume_id))
+                array_mediator.expand_volume(
+                    volume_id=volume_id,
+                    required_bytes=size)
+
+                volume = array_mediator.get_object_by_id(volume_id, config.VOLUME_TYPE_NAME)
+                res = utils.generate_csi_expand_volume_response(volume)
+                logger.info("finished expanding volume")
+                return res
```
oriyarde, line 580: these lines could and therefore should be outside the `with` (indented one level to the left)

controller/tests/controller_server/csi_controller_server_test.py

controller/controller_server/csi_controller_server.py
```diff
@@ -534,6 +539,91 @@ def GetCapacity(self, request, context):
...
+
+                volume_before_expand = array_mediator.get_object_by_id(volume_id, config.VOLUME_TYPE_NAME)
+                if not volume_before_expand:
+                    raise controller_errors.ObjectNotFoundError(volume_id)
+
+                if required_bytes == 0:
+                    context.set_code(grpc.StatusCode.OK)
+                    return csi_pb2.ControllerExpandVolumeResponse()
+
+                if not min_size <= required_bytes <= max_size:
```
oriyarde, line 574: now that I think about it - I think it's ok for `required_bytes` to be smaller than `min_size`.
what do you think?

controller/controller_server/csi_controller_server.py
```diff
@@ -534,6 +539,91 @@ def GetCapacity(self, request, context):
...
+                min_size = array_mediator.minimal_volume_size_in_bytes
+                max_size = array_mediator.maximal_volume_size_in_bytes
+
+                volume_before_expand = array_mediator.get_object_by_id(volume_id, config.VOLUME_TYPE_NAME)
+                if not volume_before_expand:
+                    raise controller_errors.ObjectNotFoundError(volume_id)
+
+                if required_bytes == 0:
+                    context.set_code(grpc.StatusCode.OK)
+                    return csi_pb2.ControllerExpandVolumeResponse()
```
oriyarde, line 572: shouldn't we include `volume_before_expand.capacity_bytes` and `node_expansion_required=False` in idempotent cases?

controller/controller_server/csi_controller_server.py
ArbelNathan: I think that the [spec](https://github.com/container-storage-interface/spec/blob/v1.3.0/spec.md#controllerexpandvolume-errors) expect in out_of_range that the capacity_range will be inside our limits

controller/controller_server/csi_controller_server.py
ArbelNathan: I think its good idea. lets not relay on what k8s might do

controller/controller_server/csi_controller_server.py
oriyarde: could you elaborate?

here are the details of my interpretation:
`If a volume corresponding to the specified volume ID is already larger than or equal to the target capacity of the expansion request, the plugin SHOULD reply 0 OK.`

`OUT_OF_RANGE Indicates that the capacity` **range** `is not allowed by the Plugin.`

`CapacityRange.required_bytes`: `Volume MUST be` **at least** `this big.`

controller/controller_server/csi_controller_server.py

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -1264,6 +1260,177 @@ def test_controller_get_capabilities(self):
...
+
+        self.servicer.ControllerExpandVolume(self.request, self.context)
+
+        self.assertEqual(self.context.code, grpc.StatusCode.NOT_FOUND)
+
+        self.mediator.get_object_by_id.side_effect = [self.volume_before_expand, None]
+
+        self.servicer.ControllerExpandVolume(self.request, self.context)
+
+        self.assertEqual(self.context.code, grpc.StatusCode.NOT_FOUND)
```
oriyarde, line 1389: please split it into two separate unit tests

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -1264,6 +1260,177 @@ def test_controller_get_capabilities(self):
...
+        self.mediator.expand_volume.assert_not_called()
+
+    @patch("controller.controller_server.csi_controller_server.get_agent")
+    def test_expand_volume_with_size_zero(self, storage_agent):
+        self._prepare_expand_volume_mocks(storage_agent)
+        self.request.capacity_range.required_bytes = 0
+        self._test_no_expand_needed()
+
+    @patch("controller.controller_server.csi_controller_server.get_agent")
+    def test_expand_volume_with_source_volume_already_in_range(self, storage_agent):
```
oriyarde, line 1349: source?

controller/tests/controller_server/csi_controller_server_test.py
ArbelNathan: done

controller/tests/controller_server/csi_controller_server_test.py

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -1264,6 +1260,180 @@ def test_controller_get_capabilities(self):
...
+    def test_expand_volume_with_first_object_not_found(self, storage_agent):
+        self._prepare_expand_volume_mocks(storage_agent)
+        self.mediator.get_object_by_id.side_effect = [None, None]
+
+        self.servicer.ControllerExpandVolume(self.request, self.context)
+
+        self.assertEqual(self.context.code, grpc.StatusCode.NOT_FOUND)
+
+    @patch("controller.controller_server.csi_controller_server.get_agent")
+    def test_expand_volume_with_second_object_not_found(self, storage_agent):
```
oriyarde, line 1386: I wouldn't consider the expanded volume a "second object".
perhaps `test_expand_volume_not_found_before_expansion` and `test_expand_volume_not_found_after_expansion`

ArbelNathan: done

controller/controller_server/csi_controller_server.py
ArbelNathan: as I see it - `CapacityRange` contains the range of the new volume size , `required_bytes` is min and `limit_bytes` is max. in `OUT_OF_RANGE` we need to check the range given by the request- the min and max. so  if `required_bytes `(or "the min") is lower then our min limitation, we suppose to return `OUT_OF_RANGE `..

oriyarde: but why would they expect us to enforce the *entire* requested range?
if someone asks you for any value that is larger than 1, and you can provide values only between 3 and 5, why wouldn't you just give them e.g. 3?

oriyarde: similar question - if they only specify `limit_bytes` of 10, and our `maximal_volume_size_in_bytes` is only 9, do you think we should fail with `OUT_OF_RANGE` as well?
I think it just means that the request is more "flexible" than us, which should increase the likelihood for a success.

ArbelNathan: well, yes, but in what size the costumer actually wanted the volume?

oriyarde: `To specify an exact size, required_bytes and limit_bytes SHALL be set to the same value.`
(otherwise, the wanted size is a range, not an exact value)

ArbelNathan: @erantzabari can we have a third opinion?

erantzabari: Starting from the bottom line - We don't need to fire out OUT_OF_RANGE in case required_bytes is smaller than min_size.
The main goal (also as far as I read in the spec) is that we want to expend volume which is size is smaller than the requested size. If the volume size is larger than the requested size, we shall return OK neglected to boundaries.

controller/controller_server/csi_controller_server.py
```diff
@@ -534,6 +539,88 @@ def GetCapacity(self, request, context):
...
+                volume_before_expand = array_mediator.get_object_by_id(volume_id, config.VOLUME_TYPE_NAME)
+                if not volume_before_expand:
+                    raise controller_errors.ObjectNotFoundError(volume_id)
+
+                if required_bytes == 0 or volume_before_expand.capacity_bytes >= required_bytes:
+                    context.set_code(grpc.StatusCode.OK)
+                    return utils.generate_csi_expand_volume_response(volume_before_expand.capacity_bytes,
+                                                                     node_expansion_required=False)
+
+                if not required_bytes <= max_size:
```
oriyarde, line 575: IMO, `if required_bytes > max_size:` is more readable

controller/controller_server/utils.py
```diff
@@ -158,6 +158,24 @@ def validate_delete_snapshot_request(request):
...
+def validate_expand_volume_request(request):
+    logger.debug("validating expand volume request")
+
+    if not request.volume_id:
+        raise ValidationException(messages.id_should_not_be_empty_message)
+
+    logger.debug("validating volume capacity")
+    if request.capacity_range:
+        if request.capacity_range.required_bytes < 0:
+            raise ValidationException(messages.size_bigger_than_0_message)
```
oriyarde, line 170: I know it is an existing message, but now that I think about it, size **can** be 0, so it should be changed to something like `size can/should not be negative/lower than 0`

controller/controller_server/csi_controller_server.py

controller/controller_server/utils.py

oriyarde: regarding the variable name, it is a good opportunity to align it with the others to not sound "ok".
i.e. `size_not_negative_message` sounds like the size was found to be not negative/required to be negative.
consider something like `size_should_not_be_negative_message`

controller/array_action/messages.py
```diff
@@ -62,3 +62,8 @@
 ObjectIsStillInUseError_message = "Object {0} is still in use by {1}"
 
 InvalidCliResponseError_message = "Invalid CLI response. Details : {0}"
+
+NotEnoughSpaceInPoolError_message = "Not enough space in pool {0}"
+
+SizeOutOfRangeError_message = "requested size is out of limits. requested: {0}," \
+                              " limits are: min_in_byte: {1}, max_in_byte: {2}"
```
oriyarde, line 69: consider removing the confusing `min_size` details until we add support for `limit_bytes`

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -1264,6 +1260,176 @@ def test_controller_get_capabilities(self):
         self.servicer.ControllerGetCapabilities(request, context)
 
 
+class TestControllerServerExpandVolume(AbstractControllerTest):
+
+    def get_create_object_method(self):
```
oriyarde, line 1265: why `create?`

controller/controller_server/csi_controller_server.py
```diff
@@ -534,6 +539,88 @@ def GetCapacity(self, request, context):
...
+
+                required_bytes = request.capacity_range.required_bytes
+                min_size = array_mediator.minimal_volume_size_in_bytes
+                max_size = array_mediator.maximal_volume_size_in_bytes
+
+                volume_before_expand = array_mediator.get_object_by_id(volume_id, config.VOLUME_TYPE_NAME)
+                if not volume_before_expand:
+                    raise controller_errors.ObjectNotFoundError(volume_id)
+
+                if required_bytes == 0 or volume_before_expand.capacity_bytes >= required_bytes:
```
oriyarde, line 570: the `required_bytes == 0` check could theoretically save us an unnecessary call to `expand_volume`, e.g. if the sidecar would actually send 0 **and** if a negative capacity would return from a mediator (i.e. mediator bug).
but I'm not sure if this reason is good enough to keep this seemingly unnecessary check.
what do you think?

controller/tests/controller_server/csi_controller_server_test.py
ArbelNathan: its changing in [here](https://github.com/IBM/ibm-block-csi-driver/pull/212#discussion_r460808027)..

oriyarde: ok

controller/controller_server/csi_controller_server.py
ArbelNathan: the idea here is that if  `required_bytes == 0` then we dont care about the volume size.. but if you think that its redundant I can remove that check.

oriyarde: I agree that 0 means that the volume size doesn't matter, so the code is more robust with this check, but we do call `get_object_by_id` (for `NOT_FOUND` error) anyway, so it doesn't add any other value.
and IMO, in this case, simplicity is more important than the insignificant additional "robustness".

ArbelNathan: ok, done

controller/array_action/messages.py
ArbelNathan: ok, done

oriyarde (dismissed):
GJ

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -24,6 +24,8 @@ def setUp(self):
         self.detect_array_type = detect_array_type_patcher.start()
         self.detect_array_type.return_value = "a9k"
         self.addCleanup(detect_array_type_patcher.stop)
+        self.request = Mock()
```
erantzabari, line 27: I am not sure it is accurate to put it on the global setup since several tests change the request (an not all the information in the request).
For example: self.request.name, self.request.source_volume_id, self.request.parameters, self.request.volume_content_source, etc.
When Mock called each setup it doesn't contain leftovers which might affect the test.
(Not all setUp methods call to super().setUp)

controller/controller_server/csi_controller_server.py
```diff
@@ -614,20 +708,24 @@ def _join_object_prefix_with_name(self, prefix, name):
...
-                    service=csi_pb2.PluginCapability.Service(type=types.Value(cap))
-                )
-            )
-
+                    service=csi_pb2.PluginCapability.Service(type=service_type.Value(service_capability))))
+        if volume_expansion_capability:
+            capability_list.append(
+                csi_pb2.PluginCapability(
+                    volume_expansion=csi_pb2.PluginCapability.VolumeExpansion(
+                        type=volume_expansion_type.Value(volume_expansion_capability))))
```
erantzabari, line 719: Isn't the result of it "ONLINE" and not expansion?

controller/controller_server/csi_controller_server.py
```diff
@@ -534,6 +539,93 @@ def GetCapacity(self, request, context):
...
+                    context.set_code(grpc.StatusCode.OK)
+                    return utils.generate_csi_expand_volume_response(volume_before_expand.capacity_bytes,
+                                                                     node_expansion_required=False)
+
+                logger.debug("expanding volume {0}".format(volume_id))
+                array_mediator.expand_volume(
+                    volume_id=volume_id,
+                    required_bytes=required_bytes)
+
+                volume_after_expand = array_mediator.get_object_by_id(volume_id, config.VOLUME_TYPE_NAME)
```
erantzabari, line 585: Why we are not checking the new volume size after expansion?

controller/tests/controller_server/csi_controller_server_test.py
ArbelNathan: a. all `setUp` methods that inherit from `AbstractControllerTest` calls to `super().setUp`.
b. we are not using a `self.request` parameter that we are not config to our needs or at least knows what its value is..

controller/controller_server/csi_controller_server.py
ArbelNathan: that is how the response built

controller/controller_server/csi_controller_server.py
ArbelNathan: cause it can be slightly other values and we consider the storage OK as enough

controller/controller_server/csi_controller_server.py
erantzabari: OK. I see the format is "volume_expansion=ONLINE"

erantzabari (approved):

oriyarde (approved):

PR #257, opened 2020-11-19
CSI-2350 finish allowing openssl 1.0.2

oriyarde:
a continuation of https://github.com/IBM/ibm-block-csi-driver/pull/249

### DIFF
--- a/Dockerfile-csi-controller
+++ b/Dockerfile-csi-controller
@@ -33,6 +33,8 @@ COPY controller/requirements.txt /driver/controller/
 RUN pip3 install --default-timeout=100 --upgrade pip==19.1.1
 # avoid default boringssl lib, since it does not support z systems
 ENV GRPC_PYTHON_BUILD_SYSTEM_OPENSSL=True
+# TODO: remove CRYPTOGRAPHY_ALLOW_OPENSSL_102 when upgrading to ubi8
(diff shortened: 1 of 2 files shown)


Dockerfile-csi-controller
```diff
@@ -33,6 +33,8 @@ COPY controller/requirements.txt /driver/controller/
 RUN pip3 install --default-timeout=100 --upgrade pip==19.1.1
 # avoid default boringssl lib, since it does not support z systems
 ENV GRPC_PYTHON_BUILD_SYSTEM_OPENSSL=True
+# TODO: remove CRYPTOGRAPHY_ALLOW_OPENSSL_102 when upgrading to ubi8
```
ArbelNathan, line 36: is there any ticket for it that you can link?

oriyarde: no, and AFAIK it might take years before we want to do that, so I think it's too early to open a ticket

ArbelNathan: so it will be a legacy TODO?

oriyarde: you could call it that. it's just something that needs to be done when changing the base image to ubi8 in this file

roysahar-ibm (approved):

ArbelNathan (approved):

PR #278, opened 2020-12-28
Add a comment to the yaml demo-pvc-file-system for volumeMode

lisyukanna:


deploy/kubernetes/examples/demo-pvc-file-system.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: v1
 metadata:
   name: demo-pvc-file-system
 spec:
-  volumeMode: Filesystem
+  volumeMode: Filesystem  #Optional. Values Filesystem\Block. The default is Filesystem.
```
oriyarde, line 6: add 1 space between `#` and `Optional`

oriyarde (approved):
LGTM

deploy/kubernetes/examples/demo-pvc-file-system.yaml
lisyukanna: done

ArbelNathan:
so why we need demo-pvc-raw-block.yaml? and should we change the name of the modified file?

oriyarde:
> so why we need demo-pvc-raw-block.yaml? and should we change the name of the modified file?

it was added with raw block support and it will simplify the example tests.
change to what?

lisyukanna:
> so why we need demo-pvc-raw-block.yaml? and should we change the name of the modified file?

what do you mean?

ArbelNathan:
> 
> 
> > so why we need demo-pvc-raw-block.yaml? and should we change the name of the modified file?
> 
> it was added with raw block support and it will simplify the example tests.
> change to what?

to demo-pvc.yaml.. so why the comment if we have both?

ArbelNathan:
> 
> 
> > so why we need demo-pvc-raw-block.yaml? and should we change the name of the modified file?
> 
> what do you mean?

we have similar pvc but with `volumeMode: Block` so if we have comment that say that `volumeMode` can be `Block `or `FileSystem` so do we still need both files?

oriyarde:
> > > so why we need demo-pvc-raw-block.yaml? and should we change the name of the modified file?
> > 
> > 
> > what do you mean?
> 
> we have similar pvc but with `volumeMode: Block` so if we have comment that say that `volumeMode` can be `Block `or `FileSystem` so do we still need both files?

as I mentioned in CSI-1855, the separate yaml doesn't tell that the field is optional and which value is the default.
and as I mentioned here, it will also simplify the example tests.

deploy/kubernetes/examples/demo-pvc-file-system.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: v1
 metadata:
   name: demo-pvc-file-system
 spec:
-  volumeMode: Filesystem
+  volumeMode: Filesystem   # Optional. Values Filesystem\Block. The default is Filesystem.
```
ArbelNathan, line 6: please remove `Values Filesystem\Block.`:
`# Optional. The default is Filesystem.`

oriyarde: LGTM as well

lisyukanna: done

oriyarde (approved):

ArbelNathan (approved):

PR #282, opened 2021-01-18
CSI-2471 fix arg for xfs_growfs

oriyarde:
in [contrast to ext4](https://linux.die.net/man/8/resize2fs), a mount-point directory [should be used](https://linux.die.net/man/8/xfs_growfs) to expand xfs.
and since `NodeExpandVolume` [may be called before NodePublishVolume](https://github.com/container-storage-interface/spec/blob/v1.3.0/spec.md#nodeexpandvolume), it cannot rely on the publish target path to be mounted, so `NodeStageVolume` must mount the staging path in advance.

please note:
1. this means we cannot create the (unnecessary) stage info file in the staging directory anymore
2. this should not affect raw block

### DIFF
--- a/node/pkg/driver/node.go
+++ b/node/pkg/driver/node.go
@@ -27,9 +27,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"k8s.io/utils/mount"
-	"os"
 	"path"
-	"path/filepath"
 	"strings"
 )
 
@@ -122,7 +120,7 @@ func (d *NodeService) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 	volId := req.VolumeId
 	err = d.VolumeIdLocksMap.AddVolumeLock(volId, "NodeStageVolume")
 	if err != nil {
-		logger.Errorf("Another operation is being perfomed on volume : {%s}.", volId)
+		logger.Errorf("Another operation is being performed on volume : {%s}.", volId)
 		return nil, status.Error(codes.Aborted, err.Error())
 	}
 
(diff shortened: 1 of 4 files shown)


node/pkg/driver/node.go
```diff
@@ -641,7 +643,12 @@ func (d *NodeService) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	err = d.NodeUtils.ExpandFilesystem(device, existingFormat)
+	mountPointToExpand := req.GetStagingTargetPath()
```
ArbelNathan, line 646: `staging_target_path`  equal to `mount-point`?

oriyarde: it is now.
alternatively, we could just ignore `staging_target_path` and always use `VolumePath`, whether it contains the publish or the stage path

roysahar-ibm (approved):
Approved

node/pkg/driver/node.go
ArbelNathan: it will be the same for ext and xfs?

oriyarde: yes, but it is not used for ext

node/pkg/driver/node.go
```diff
@@ -264,6 +263,36 @@ func (d *NodeService) nodeStageVolumeRequestValidation(req *csi.NodeStageVolumeR
...
+	fsTypeForMount := requestedFsType
+	if requestedFsType == "" {
+		if existingFormat == "" {
+			fsTypeForMount = defaultFSType
+		} else {
+			fsTypeForMount = existingFormat
+		}
+	} else if existingFormat != "" {
+		if requestedFsType != existingFormat {
+			return "", status.Errorf(codes.AlreadyExists, "Requested fs_type {%v} but found {%v}", requestedFsType, existingFormat)
```
ArbelNathan, line 276: no way of placing this in the validation part?

oriyarde: sure, we could return `existingFormat` from the validation function, or just call `GetDiskFormat` twice.
though it would make us call `GetAccessType()` twice (at least it doesn't require an external i/o).
note that currently all validation functions only perform "static" checks (i.e. without external i/o, such as `GetDiskFormat`)

ArbelNathan: ok, you right, better that  external i/o will be reduce as much as can be..

ArbelNathan (approved):
LGTM
great job

PR #293, opened 2021-02-14
Bug/csi 2597 community tests failures svc

ArbelNathan:
resolutions for problem 1 in ticket and minor fix for 3 which not resolve the problem but fix the ex bug.


controller/controller_server/csi_controller_server.py
```diff
@@ -706,9 +706,7 @@ def _get_object_final_name(self, request, max_name_prefix_length, max_name_lengt
         return full_name[:max_name_length]
 
     def _join_object_prefix_with_name(self, prefix, name):
-        if prefix:
-            return settings.NAME_PREFIX_SEPARATOR.join((prefix, name))
-        return name
+        return settings.NAME_PREFIX_SEPARATOR.join((prefix, name))
```
oriyarde, line 709: does this mean a prefix-less volume will always begin with the separator, for every storage system?

controller/controller_server/csi_controller_server.py
```diff
@@ -706,9 +706,7 @@ def _get_object_final_name(self, request, max_name_prefix_length, max_name_lengt
         return full_name[:max_name_length]
 
     def _join_object_prefix_with_name(self, prefix, name):
```
oriyarde, line 708: since the function is now a one liner - do we still need it?

controller/controller_server/csi_controller_server.py
ArbelNathan: yes, do you think its a problem?

oriyarde: I am not sure if we want to begin with a separator for systems which can work without it.
@zingero what do you think?

zingero: As far as I understand from the variable name, the separator is related to the prefix and not to the suffix. Thus, If there is nothing to separate, we don't need to use a separator

controller/controller_server/csi_controller_server.py
```diff
@@ -512,14 +512,14 @@ def DeleteSnapshot(self, request, context):
                 except controller_errors.ObjectNotFoundError as ex:
                     logger.debug("Snapshot was not found during deletion: {0}".format(ex))
 
-        except controller_errors.ObjectNotFoundError:
-            logger.debug("snapshot was not found during deletion: {0}".format(ex))
+        except controller_errors.ObjectNotFoundError as ex:
+            logger.debug("snapshot was not found during deletion: {0}".format(ex.message))
```
zingero, line 516: Why do we format the `ex.message` instead of `ex`? Some exceptions don't have meaningful messages

controller/controller_server/csi_controller_server.py
oriyarde: perhaps this change lacks a context - it makes a test pass, for more info please see the ticket

controller/controller_server/csi_controller_server.py
oriyarde: could you please [point to an exception](https://github.com/IBM/ibm-block-csi-driver/blob/73e76506dee47c5fcc8f4069931db994af3087b7/controller/array_action/errors.py) which doesn't have a meaningful message?
I think meaningful messages are important

zingero: I meant non-IBM exceptions (Pythonic exceptions, 3rd Parties exceptions, etc.).
Personally, in IBMic exceptions, I prefer to override `__str__` and/or `__repr__` and return the relevant information because it saves the developer checking if the exception has a meaningful exception or not.

controller/controller_server/csi_controller_server.py
zingero: The prefix-related text in the ticket is what you wrote:
`when a user prefix is not specified, the controller should guarantee (at least for svc) that the final name doesn't start with a number.`
I'm not sure how it's related though. The prefix separator should separate the prefix and the suffix. No prefix means no separator, no?

controller/controller_server/csi_controller_server.py
oriyarde: I am not sure I follow - AFAIK these formattings of `ex.message` are done only for our own exceptions, so I don't see an issue with this change.
regarding the refactor you proposed - it sounds possible, but not related to this PR, feel free to open a ticket for it.

controller/controller_server/csi_controller_server.py
oriyarde: this change uses the separator to ensure, for all storage systems (not just svc), that the final name doesn't start with a number

oriyarde: IMO, if this failure is svc-only, then we can consider adding some info in the mediator interface that would be useful only for svc.
e.g.
```
if not prefix:
    prefix = mediator.default_volume_prefix
```

controller/controller_server/csi_controller_server.py
zingero: A [ticket](https://jira.xiv.ibm.com/browse/CSI-2785) has been created.
Let's continue discussing there.

oriyarde: please avoid using internal links in public repos

controller/controller_server/csi_controller_server.py
zingero: I prefer adding a default prefix when it's necessary than having a single character as an assurance for non-digit first character

oriyarde: by "when it's necessary" - do you mean only for certain storage systems, as in my suggestion?
or some cross-storage default prefix (that isn't required for some storage systems)?

zingero: As in your suggestion.

ArbelNathan: what are suggesting for a default prefix?

zingero: I suggest making it a configurable value and use the storage type as a default value.

ArbelNathan: so volume will start svc_...?

zingero: In this case, yes. Where it's not mandatory, no prefix/separator should be included.

ArbelNathan: @oriyarde agreed?

oriyarde: [here](https://github.com/IBM/ibm-block-csi-driver/pull/293#discussion_r575818570) is how I think it should be implemented, and I would use a prefix that will convey more info to the storage admin regarding the origin of this prefix, e.g. "CSI" (and not "svc")

ArbelNathan: done, plz check

controller/controller_server/csi_controller_server.py
```diff
@@ -676,18 +676,20 @@ def _get_volume_final_name(self, request, array_mediator):
...
                                            config.SNAPSHOT_TYPE_NAME,
-                                           config.PARAMETERS_SNAPSHOT_NAME_PREFIX)
+                                           config.PARAMETERS_SNAPSHOT_NAME_PREFIX,
+                                           array_mediator.default_object_prefix)
         return name
 
     def _get_object_final_name(self, request, max_name_prefix_length, max_name_length, object_type,
-                               prefix_param_name):
-        full_name = name = request.name
+                               prefix_param_name, default_prefix):
```
oriyarde, line 691: I'm starting to think we should pass the mediator, instead of many of its properties

ArbelNathan: the mediator is very big, should we?

controller/array_action/array_mediator_svc.py
```diff
@@ -140,6 +140,10 @@ def maximal_volume_size_in_bytes(self):
     def max_lun_retries(self):
         return 10
 
+    @classproperty
+    def default_object_prefix(self):
```
oriyarde, line 144: please add a unit test that asserts `default_object_prefix` is not larger than `max_volume_prefix_length` and `max_snapshot_prefix_length`

controller/controller_server/csi_controller_server.py
oriyarde: it's only 1 reference, compared to current 3

zingero (dismissed):

controller/controller_server/csi_controller_server.py
oriyarde: LGTM

controller/controller_server/csi_controller_server.py
ArbelNathan: still relevance?

controller/controller_server/csi_controller_server.py
ArbelNathan: 6, snapshot and volume, how do you want to unite them? (and we really want to have size for each volume type?)

oriyarde: perhaps I wasn't clear - I only suggest to pass `mediator` instead of `max_name_prefix_length`, `max_name_length` and `default_prefix`

controller/controller_server/csi_controller_server.py
oriyarde: no

controller/controller_server/csi_controller_server.py
ArbelNathan: look at the parameters source.. the calling functions not using the same values, how can we know what values to use?

controller/controller_server/csi_controller_server.py
ArbelNathan: @zingero plz check and resolve

controller/controller_server/csi_controller_server.py
oriyarde: current calls:
```
self._get_object_final_name(request, array_mediator.max_volume_prefix_length,
                            array_mediator.max_volume_name_length,
                            config.VOLUME_TYPE_NAME,
                            config.PARAMETERS_VOLUME_NAME_PREFIX,
                            array_mediator.default_object_prefix)

self._get_object_final_name(request, array_mediator.max_snapshot_prefix_length,
                            array_mediator.max_snapshot_name_length,
                            config.SNAPSHOT_TYPE_NAME,
                            config.PARAMETERS_SNAPSHOT_NAME_PREFIX,
                            array_mediator.default_object_prefix)
```
my suggestion:
```
self._get_object_final_name(request, array_mediator,
                            config.VOLUME_TYPE_NAME,
                            config.PARAMETERS_VOLUME_NAME_PREFIX)

self._get_object_final_name(request, array_mediator,
                            config.SNAPSHOT_TYPE_NAME,
                            config.PARAMETERS_SNAPSHOT_NAME_PREFIX)
```
hope it's clear

zingero: @ArbelNathan , what do you mean by "the mediator is very big"? Is it a "god object"? 
Maybe it's another thing we can improve.
I'm not sure it's relevant for this task though.

controller/controller_server/csi_controller_server.py
zingero: LGTM indeed

controller/controller_server/csi_controller_server.py
ArbelNathan: @oriyarde and inside `_get_object_final_name` should we use `max_volume_name_length` or `max_snapshot_name_length` or add if for every call and choose by `<>_TYPE_NAME`? cause its not sound pretty

ArbelNathan: @zingero Im not sure I understand the term "god object", can you elaborate?

zingero: God object is an object that knows too much or does too much. The God object is an example of an anti-pattern
https://en.wikipedia.org/wiki/God_object

oriyarde: @ArbelNathan I see what you mean.
if `max_volume_name_length` and `max_snapshot_name_length` are always the same, then perhaps we can unify them to one `max_object_name_length`.
if not, then I agree that choosing by `<>_TYPE_NAME` isn't pretty.
feel free to unify them, open a ticket to unify them, or just resolve this conversation without any change.

@zingero regarding the god object - we can move some of the mediator properties values to its abstract class or other objects, such as `Volume` and `Snapshot`, etc...
but it's definitely not for this ticket.

controller/controller_server/csi_controller_server.py
ArbelNathan: great, @oriyarde please resolve, you opened

controller/controller_server/csi_controller_server.py
zingero: Of course, we'll improve it in the future.

oriyarde: I opened CSI-2794

controller/array_action/array_mediator_svc.py
ArbelNathan: done

controller/tests/array_action/svc/array_mediator_svc_test.py
```diff
@@ -63,6 +63,11 @@ def test_close(self):
         self.svc.disconnect()
         self.svc.client.close.assert_called_with()
 
+    def test_default_object_prefix_size(self):
```
oriyarde, line 66: rename `size` to `length`

oriyarde, line 66: consider adding something like `_not_larger_than_max`

ArbelNathan: done

ArbelNathan: done

oriyarde (approved):
LGTM

zingero (dismissed):

oriyarde (approved):

zingero (approved):

PR #295, opened 2021-03-07
Feature/csi 2726 electronic license 1.5.0

ArbelNathan:
files moved to CSI-2741


oriyarde (approved):

zingero (approved):
For future versions, I suggest updating this file dynamically. i.e. we write the version in 1 place and generate this file using a template. WDYT? Should I open a ticket for that?

PR #301, opened 2021-03-30
CSI-1516 align illegal obejct id error gaps

oriyarde:
https://github.com/container-storage-interface/spec/blob/v1.4.0/spec.md#error-scheme

> INVALID_ARGUMENT - Indicates ... an invalid value.

### DIFF
--- a/controller/array_action/array_mediator_ds8k.py
+++ b/controller/array_action/array_mediator_ds8k.py
@@ -384,8 +384,14 @@ class DS8KArrayMediator(ArrayMediatorAbstract):
         logger.debug("Searching for volume with id: {0}".format(volume_id))
         try:
             api_volume = self.client.get_volume(volume_id)
-        except exceptions.NotFound:
-            raise array_errors.ObjectNotFoundError(volume_id)
+        except (exceptions.NotFound, exceptions.InternalServerError) as ex:
+            uppercase_message = str(ex.message).upper()
+            if ERROR_CODE_RESOURCE_NOT_EXISTS in uppercase_message:
+                raise array_errors.ObjectNotFoundError(volume_id)
+            elif INCORRECT_ID in uppercase_message:
+                raise array_errors.IllegalObjectID(volume_id)
+            else:
+                raise ex
 
         vol_name = api_volume.name
         logger.debug("found volume name : {0}".format(vol_name))
@@ -496,7 +502,7 @@ class DS8KArrayMediator(ArrayMediatorAbstract):
         except exceptions.NotFound:
             if not_exist_err:
                 raise array_errors.ObjectNotFoundError(volume_id)
-        except (exceptions.ClientError, exceptions.Unauthorized) as ex:
+        except (exceptions.ClientError, exceptions.InternalServerError) as ex:
             if INCORRECT_ID in str(ex.message).upper():
                 raise array_errors.IllegalObjectID(volume_id)
 
--- a/controller/array_action/array_mediator_svc.py
+++ b/controller/array_action/array_mediator_svc.py
@@ -20,7 +20,12 @@ array_connections_dict = {}
 logger = get_stdout_logger()
 
 OBJ_NOT_FOUND = 'CMMVC5753E'
-NAME_NOT_MEET = 'CMMVC5754E'
+NAME_NOT_EXIST_OR_MEET_RULES = 'CMMVC5754E'
+NON_ASCII_CHARS = 'CMMVC6017E'
+INVALID_NAME = 'CMMVC6527E'
+TOO_MANY_CHARS = 'CMMVC5738E'
+VALUE_TOO_LONG = 'CMMVC5703E'
+INVALID_FILTER_VALUE = 'CMMVC5741E'
 SPECIFIED_OBJ_NOT_EXIST = 'CMMVC5804E'
 VOL_ALREADY_MAPPED = 'CMMVC5878E'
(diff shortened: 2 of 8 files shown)


controller/array_action/array_mediator_svc.py
```diff
@@ -215,25 +220,27 @@ def _generate_snapshot_response_with_verification(self, cli_object):
             raise controller_errors.ExpectedSnapshotButFoundVolumeError(cli_object.name, self.endpoint)
         return self._generate_snapshot_response(cli_object, fcmap.source_vdisk_name)
 
-    def _get_cli_volume(self, volume_name_or_id, not_exist_err=True):
+    def _get_cli_volume(self, volume_name, not_exist_err=True):
```
ArbelNathan, line 223: why its so different then `_get_cli_volume_by_wwn`? can't we unite them with some `is_id` flag for `filter_value`?

controller/tests/array_action/xiv/array_mediator_xiv_tests.py
```diff
@@ -458,13 +505,13 @@ def test_map_volume_no_availabe_lun(self):
 
         self.mediator.client.cmd.mapping_list.return_value = Mock(as_list=[mapping1, mapping2, mapping3])
         with self.assertRaises(array_errors.NoAvailableLunError):
-            self.mediator.map_volume("vol", "host")
+            self.mediator.map_volume("vol-wwn", "host")
```
ArbelNathan, line 508: if you already here, `volume-wwn`?

controller/array_action/array_mediator_ds8k.py
```diff
@@ -384,8 +384,14 @@ def get_volume_name(self, volume_id):
         logger.debug("Searching for volume with id: {0}".format(volume_id))
         try:
             api_volume = self.client.get_volume(volume_id)
```
ArbelNathan, line 386: can you unify the `self.client.get_volume(volume_id)` calls?

controller/array_action/array_mediator_svc.py
oriyarde: they handle different errors, so most of their code is different (not just the `filter_value`).
I would leave them separated.

controller/tests/array_action/xiv/array_mediator_xiv_tests.py
oriyarde: this change is a small alignment to match most existing values, but your request is relevant to **all** existing values, and this PR diff is already large enough as it is.
I hope to make a separate renaming PR, if so, I can include it there

controller/array_action/array_mediator_ds8k.py
oriyarde: what do you mean?
should calling `get_volume_name` invoke `get_flashcopies_by_volume`?

oriyarde: anyway, I'm not sure if it's relevant to this PR, which is already large

controller/tests/array_action/xiv/array_mediator_xiv_tests.py
ArbelNathan: great

controller/controller_server/csi_controller_server.py
```diff
@@ -23,6 +22,7 @@
 from controller.csi_general import csi_pb2_grpc
 
 logger = None  # is set in ControllerServicer::__init__
+logger: logging.Logger
```
zingero, line 25: How this change is related to this PR?
I'm not sure we need it.

zingero (changes requested):
Almost :)

controller/controller_server/csi_controller_server.py
oriyarde: it's not, it only allows auto-completion for `logger`.
do you think it should be undone?

ArbelNathan: I think its good that it no longer shows warnings

zingero: I'm not sure why we use the `logging.Logger` class instead of using `logging`.
We still see a warning in case of:
`logger = None  # is set in ControllerServicer::__init__`
`logger: logging.Logger`
since `None` is not an instance of `logging.Logger`

[This](https://github.com/zingero/smart-fridge/blob/ef530fd0a64e7fd88f570a519ac505d2788039c7/main.py) is an example (from a different repository) of using the `logging` module directly.

oriyarde: I wouldn't worry about this warning (a lot of them don't mean much to us), but feel free to open a ticket for it

zingero: CSI-2894

zingero (approved):

controller/array_action/array_mediator_ds8k.py
ArbelNathan: `get_flashcopies_by_volume ` is not an actual operation, it's just a workaround to a  bug with pyds8k so It should be called only where necessary and not in the common method. I can see that it will cause another `try` block and eventually won't save lines.. forget about it for now..

ArbelNathan (approved):
LGTM

controller/controller_server/csi_controller_server.py
oriyarde: @zingero can this conversation be resolved?

PR #305, opened 2021-04-01
CSI-1358 expose exception from with blocks

zingero:


oriyarde (dismissed):
LGTM

ArbelNathan (approved):
LGTM

controller/controller_server/exception_handler.py
```diff
@@ -0,0 +1,42 @@
...
+    def inner(servicer, request, context):
+        try:
+            return call(servicer, request, context)
+        except Exception as ex:
+            logger.exception(ex)
+            context.set_details(str(ex))
+            status_code = exceptions_to_handlers_map.get(type(ex), grpc.StatusCode.INTERNAL)
+            context.set_code(status_code)
+            return response_types[call.__name__]()
+    return inner
```
oriyarde, line 42: consider using [`from decorator import decorator`](https://stackoverflow.com/a/18577188)

zingero: Done

controller/controller_server/exception_handler.py
```diff
@@ -0,0 +1,43 @@
...
+    controller_errors.IllegalObjectName: grpc.StatusCode.INVALID_ARGUMENT,
+    controller_errors.SpaceEfficiencyNotSupported: grpc.StatusCode.INVALID_ARGUMENT,
+    controller_errors.PoolDoesNotExist: grpc.StatusCode.INVALID_ARGUMENT,
+    controller_errors.PoolDoesNotMatchCapabilities: grpc.StatusCode.INVALID_ARGUMENT,
+    controller_errors.PoolParameterIsMissing: grpc.StatusCode.INVALID_ARGUMENT,
+    controller_errors.ExpectedSnapshotButFoundVolumeError: grpc.StatusCode.INVALID_ARGUMENT,
+    controller_errors.ObjectNotFoundError: grpc.StatusCode.NOT_FOUND,
+    controller_errors.PermissionDeniedError: grpc.StatusCode.PERMISSION_DENIED,
+    controller_errors.VolumeAlreadyExists: grpc.StatusCode.ALREADY_EXISTS,
+    controller_errors.NotEnoughSpaceInPool: grpc.StatusCode.RESOURCE_EXHAUSTED
```
ArbelNathan, line 25: can you order them by `grpc.StatusCode` type plz?

controller/controller_server/exception_handler.py
```diff
@@ -0,0 +1,43 @@
...
+}
+
+
+@decorator
+def handle_requests_safely(call, servicer, request, context):
+    try:
+        return call(servicer, request, context)
+    except Exception as ex:
+        logger.exception(ex)
+        context.set_details(str(ex))
```
ArbelNathan, line 40: Is that the same as ex.message?

ArbelNathan:
Great! now do it for all other methods! :) (or open a ticket for it..)

controller/controller_server/exception_handler.py
```diff
@@ -0,0 +1,43 @@
...
+
+import controller.array_action.errors as controller_errors
+from controller.common.csi_logger import get_stdout_logger
+from controller.controller_server.errors import ObjectIdError
+from controller.controller_server.errors import ValidationException
+from controller.csi_general import csi_pb2
+
+logger = get_stdout_logger()
+
+exceptions_to_status_codes_map = {
```
oriyarde, line 13: consider avoiding `map` by renaming to something like `status_codes_by_exception`

controller/controller_server/exception_handler.py
oriyarde, line 25: please leave here only the pairs in which the exception *always* leads to the status code.
e.g. `ObjectIdError` does not belong here, since it doesn't always lead to `NOT_FOUND`

zingero: Done

controller/controller_server/exception_handler.py
zingero: It's a little better since sometimes (in Pythonic exceptions), we don't have a `message` field or in other cases, an empty one.
All the relevant IBMic exceptions have `__str__` method that returns the message.

zingero:
> Great! now do it for all other methods! :) (or open a ticket for it..)

CSI-2778 :)

controller/controller_server/exception_handler.py
zingero: Done

oriyarde: I don't see any change

controller/controller_server/exception_handler.py
```diff
@@ -0,0 +1,43 @@
...
+}
+
+
+response_types = {
+    "CreateVolume": csi_pb2.CreateVolumeResponse
+}
+
+
+@decorator
+def handle_requests_safely(call, servicer, request, context):
```
oriyarde, line 35: consider renaming `handle_requests_safely` to `handle_common_exceptions`, since the decorator should handle the exceptions, and only the common ones

oriyarde, line 35: consider renaming `call` to something like `controller_method`

controller/controller_server/exception_handler.py
```diff
@@ -0,0 +1,43 @@
...
+    controller_errors.ObjectNotFoundError: grpc.StatusCode.NOT_FOUND,
+    controller_errors.PermissionDeniedError: grpc.StatusCode.PERMISSION_DENIED,
+    controller_errors.VolumeAlreadyExists: grpc.StatusCode.ALREADY_EXISTS,
+    controller_errors.NotEnoughSpaceInPool: grpc.StatusCode.RESOURCE_EXHAUSTED
+}
+
+
+response_types = {
+    "CreateVolume": csi_pb2.CreateVolumeResponse
+}
```
oriyarde, line 31: I just thought of a way to avoid this dict.
```
def handle_common_exceptions(response_type):
    @decorator
    def common_exceptions_handler(...):
        ...
        except ...:
            ...
            return response_type()
    return common_exceptions_handler
```

```
@handle_common_exceptions(csi_pb2.CreateVolumeResponse)
def CreateVolume(self, request, context):
    ....
```
what do you think?

controller/controller_server/exception_handler.py
zingero: Done

zingero: Done

controller/controller_server/exception_handler.py
zingero: I like it.
I changed the decorator name for clarity purposes.

controller/controller_server/exception_handler.py
zingero: I uploaded a commit recently.

controller/controller_server/exception_handler.py
zingero: Done

controller/controller_server/exception_handler.py
oriyarde: I'm not sure `decorated_handle_common_exceptions` is clearer than `common_exceptions_handler` (or even correct)

controller/controller_server/exception_handler.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def handle_common_exceptions(response_type):
+    @decorator
+    def decorated_handle_common_exceptions(controller_method, servicer, request, context):
+        try:
+            return controller_method(servicer, request, context)
+        except Exception as ex:
+            if type(ex) not in status_codes_by_exception.keys():
```
oriyarde, line 32: are you sure `.keys()` is needed?

controller/controller_server/exception_handler.py
```diff
@@ -0,0 +1,39 @@
...
+                raise
```
oriyarde, line 33: shouldn't we just return with `grpc.StatusCode.INTERNAL`?

controller/controller_server/exception_handler.py
oriyarde: please also remove uncommon exceptions, e.g. `SpaceEfficiencyNotSupported`, since it is method-specific

oriyarde: another way to reduce `except` clauses is to make same-status-codes exceptions inherit from the same intermediate class

controller/controller_server/csi_controller_server.py
```diff
@@ -136,32 +123,9 @@ def CreateVolume(self, request, context):  # pylint: disable=too-many-branches
...
-            logger.exception(ex)
-            context.set_details(ex.message)
-            context.set_code(grpc.StatusCode.RESOURCE_EXHAUSTED)
-            return csi_pb2.CreateVolumeResponse()
         except Exception as ex:
+            if type(ex) in exception_handler.status_codes_by_exception.keys():
+                raise
             logger.error("an internal exception occurred")
             logger.exception(ex)
             context.set_code(grpc.StatusCode.INTERNAL)
```
oriyarde, line 131: I think this should be handled by the decorator

controller/controller_server/exception_handler.py
zingero: The **common** handler shouldn't be familiar with other (uncommon) exceptions and their status codes.
It should transfer the responsibility to the original function.

controller/controller_server/exception_handler.py
zingero: It's not mandatory, but I like the Python zen:

> 	    Explicit is better than implicit.

Do you want me to remove it?

controller/controller_server/exception_handler.py
zingero: It's recommended (by naming convention) to name functions like verbs. i.e.:
do_something
start_program
load_configuration

It's recommended (by naming convention) to name classes like nouns. i.e.:
RequestHandler
ConfigurationLoader
NetworkManager

`common_exceptions_handler` is a noun. Thus, not recommended to be a function name.
Why `decorated_handle_common_exceptions` is incorrect?

controller/controller_server/csi_controller_server.py
zingero: This isn't a common exception.

controller/controller_server/exception_handler.py
zingero: What do you mean by "method-specific"?
I would like the program to handle exceptions as much as it can instead of handling them in the server.

controller/controller_server/csi_controller_server.py
oriyarde: if by `This` you mean the `Exception` type, then inheritance-wise it is very common

controller/controller_server/exception_handler.py
oriyarde: there is no original function, this is our last line of handling.
the way i see it, there are 3 kinds of exceptions:

1. method-specific
2. common
3. other

and they should be handled by:

1. the specific method
2. the decorator
3. the decorator also

since the handling logic of "other" exceptions is also common.

zingero: Then it's not a "common exceptions handler", but "not method-specific exceptions". I'm not a fan of this design.
How can I know which exception is a method-specific exception?
Some exceptions are used only in one method, but they can be used in other future functions as well.

controller/controller_server/exception_handler.py
oriyarde: yes.
I too like the zen of Python
> Beautiful is better than ugly.

> Simple is better than complex.

and the official docs agree

controller/controller_server/exception_handler.py
oriyarde: the point here is to move common logic, which is *mostly* about common exceptions.
if the exception in question appears in only one method, or has different handlings, then it's specific.
when new usage adds, its location should be determined by said reasoning.

controller/controller_server/exception_handler.py
zingero: Done

controller/controller_server/exception_handler.py
zingero: The original point here is removing some try/except clauses and a linter warning.
The point of CSI-2778 is breaking the god object and reduce the code length by breaking the file into a couple of files wrap some code with functions.
I think we should discuss how we can improve the class and redesign it in CSI-2778, but If you want, we can talk about it tomorrow.

controller/controller_server/exception_handler.py
oriyarde: I mean that the exception is relevant for only one method.
the purpose of this decorator is to have common handlings in one place, not to have as much handlings as it can.
the default place for exception handling should still be in the server, unless it already exists somewhere else there, in which case it can move to the common decorator.

controller/controller_server/exception_handler.py
oriyarde: 1. I don't see any function name in your examples that starts with an adjective
2. IMO it's more "decorating" than it is "decorated", but the outer function is also "decorating", so I don't see the value of having this term in the name

then how about:
```
def handle_common_exceptions_with(response_type):
    @decorator
    def handle_common_exceptions(...):
        ...
    return handle_common_exceptions
```

controller/controller_server/exception_handler.py
oriyarde: both of our suggestions serve the original point you described, so I'm not sure how this helps.
I agree that this decorator can easily solve both CSI-2778 and CSI-1358, and I don't see why not

controller/controller_server/csi_controller_server.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, array_endpoint):
         with open(path, 'r') as yamlfile:
             self.cfg = yaml.safe_load(yamlfile)  # TODO: add the following when possible : Loader=yaml.FullLoader)
 
-    @handle_common_exceptions(csi_pb2.CreateVolumeResponse)
+    @exception_handler.handle_common_exceptions(csi_pb2.CreateVolumeResponse)
```
oriyarde: I actually like the previous way better - is there a particular reason for this change?

controller/controller_server/exception_handler.py
zingero: 1. Usually it starts with a verb, not an adjective


`handle_common_exceptions_with` with what?

@ArbelNathan , we have a decorator inside a function and we're struggling with naming the function and the decorated function. Do you have any suggestions?

controller/controller_server/csi_controller_server.py
zingero: Fewer import statements. I changed it

controller/controller_server/exception_handler.py
oriyarde: with "the following" response type (not the most common style, but I think it's very readable, especially from the calling side)

controller/controller_server/csi_controller_server.py
zingero: Done

controller/controller_server/exception_handler.py
zingero: Done

controller/controller_server/exception_handler.py
zingero: Done

controller/controller_server/exception_handler.py
zingero: Done

controller/controller_server/exception_handler.py
```diff
@@ -0,0 +1,42 @@
...
+    controller_errors.PermissionDeniedError: grpc.StatusCode.PERMISSION_DENIED,
+    controller_errors.VolumeAlreadyExists: grpc.StatusCode.ALREADY_EXISTS,
+    controller_errors.NotEnoughSpaceInPool: grpc.StatusCode.RESOURCE_EXHAUSTED
+}
+
+
+def handle_exception(response_type, context, ex):
+    logger.exception(ex)
+    context.set_details(str(ex))
+    status_code = status_codes_by_exception.get(type(ex), grpc.StatusCode.INTERNAL)
```
oriyarde, line 29: I think this should come from a parameter

controller/controller_server/exception_handler.py
oriyarde: I still see some, e.g. [`SpaceEfficiencyNotSupported`](https://github.com/IBM/ibm-block-csi-driver/pull/305#discussion_r608638835)...

controller/controller_server/exception_handler.py
zingero: I don't like lots of parameters in functions, but I can't find something better than that at the moment. Done

controller/controller_server/exception_handler.py
zingero: Done

controller/array_action/errors.py
```diff
@@ -183,7 +187,7 @@ def __init__(self, connectivity_type):
         self.message = messages.UnsupportedConnectivityTypeError_message.format(connectivity_type)
 
 
-class ExpectedSnapshotButFoundVolumeError(BaseArrayActionException):
+class ExpectedSnapshotButFoundVolumeError(InvalidArgumentError):
```
oriyarde, line 190: what about `PoolParameterIsMissing` and `IllegalObjectName`?
I know they are common, but I think they are also a type of `InvalidArgumentError`

zingero: Done

oriyarde (approved):
LGTM

ArbelNathan (approved):
LGTM

controller/controller_server/exception_handler.py
```diff
@@ -0,0 +1,37 @@
...
+    return response_type()
+
+
+def handle_common_exceptions(response_type):
+    @decorator
+    def handle_common_exceptions_with_response(controller_method, servicer, request, context):
+        try:
+            return controller_method(servicer, request, context)
+        except Exception as ex:
+            return handle_exception(ex, context, status_codes_by_exception.get(type(ex), grpc.StatusCode.INTERNAL),
```
oriyarde, line 34: maybe just keep a separate a `status_code = status_codes_by_exception.get(type(ex), grpc.StatusCode.INTERNAL)` line?

zingero: Since we already merged this PR, we'll address your suggestion in a similar ticket (CSI-2778)

oriyarde: thanks

controller/controller_server/csi_controller_server.py
```diff
@@ -136,37 +123,10 @@ def CreateVolume(self, request, context):  # pylint: disable=too-many-branches
...
-            context.set_code(grpc.StatusCode.PERMISSION_DENIED)
-            context.set_details(ex.message)
-            return csi_pb2.CreateVolumeResponse()
+        except controller_errors.InvalidArgumentError as ex:
+            handle_exception(ex, context, grpc.StatusCode.INVALID_ARGUMENT, csi_pb2.CreateVolumeResponse)
         except controller_errors.VolumeAlreadyExists as ex:
-            context.set_details(ex.message)
-            context.set_code(grpc.StatusCode.ALREADY_EXISTS)
-            return csi_pb2.CreateVolumeResponse()
-        except controller_errors.NotEnoughSpaceInPool as ex:
```
oriyarde, line 159: I think this `except` block [was added](https://github.com/IBM/ibm-block-csi-driver/pull/251/files#diff-1d72325bc9ed2e0861808fe985ff7e74edfbba4533a5a71f12b8fd498bb2647c) here [by mistake](https://github.com/container-storage-interface/spec/blob/v1.5.0/spec.md#createvolume-errors) - so it shouldn't be in the common exceptions.
the status code for this exception is only mandatory in `CreateSnapshot`, optional in `ControllerExpandVolume`, and is a bug in `CreateVolume`.
it can be fixed as part of https://github.com/IBM/ibm-block-csi-driver/pull/349

zingero: [Done](https://github.com/IBM/ibm-block-csi-driver/pull/349/commits/648349dc35b82ec9070fd807ef16b06628df1d40)

PR #309, opened 2021-04-12
CSI-2883 remove useless inheritance

oriyarde:


controller/array_action/array_connection_manager.py
```diff
@@ -39,7 +39,7 @@ def _socket_connect_test(host, port, timeout=1):
         return -1
 
 
-class ArrayConnectionManager(object):
+class ArrayConnectionManager():
```
zingero, line 42: Why do we need an empty paranthesis?

oriyarde: removed

zingero (approved):

ArbelNathan (approved):
LGTM

PR #311, opened 2021-04-14
Printing the exit codes of the scripts we run when we test

zingero:
Now, we don't need to search the exit codes. After running the tests, it'll print a summary that looks like this:

> -------- Summary of exit codes --------
pycodestyle: 0
pylint: 0
unitests: 0


controller/scripts/entrypoint-test.sh
```diff
@@ -1,5 +1,17 @@
 #!/bin/bash
 set -x
```
oriyarde, line 2: could you remove all `set` lines in this file? IMO prints like `+ PYCODESTYLE=0` are just noise

zingero: Done

oriyarde (approved):
LGTM

ArbelNathan (approved):
NICE

PR #316, opened 2021-04-22
CSI-2906 enabling global statement linting

zingero:
This is still WIP. I recommend leaving it for now.
Current status: unit tests pass, but we can't see some of the controller logs after running:
`kubectl logs pod_name -c container_name`


zingero:
I'll create another soon

PR #318, opened 2021-04-26
csi-2783 fix ds8k and svc community tests

barakda:


scripts/ci/run_csi_test_client.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash -xe
 
-[ $# -ne 2 ] && { echo "Usage $0 : container_name folder_for_junit"  ; exit 1; }
+[ $# -ne 3 ] && { echo "Usage $0 : container_name folder_for_junit"  ; exit 1; }
 
 #/tmp/k8s_dir is the directory of the csi grpc\unix socket that shared between csi server and csi-test docker 
-docker build -f Dockerfile-csi-test -t csi-sanity-test .  && docker run --user=root -e STORAGE_ARRAYS=${STORAGE_ARRAYS} -e USERNAME=${USERNAME} -e PASSWORD=${PASSWORD}  -e POOL_NAME=${POOL_NAME} -v /tmp/k8s_dir:/tmp/k8s_dir:rw  -v$2:/tmp/test_results:rw --rm  --name $1 csi-sanity-test
+if [ $3 = 'community_svc' ] ; then
+    docker build -f Dockerfile-csi-svc-test -t csi-sanity-test .  && docker run --user=root -e STORAGE_ARRAYS=${STORAGE_ARRAYS} -e USERNAME=${USERNAME} -e PASSWORD=${PASSWORD}  -e POOL_NAME=${POOL_NAME} -v /tmp/k8s_dir:/tmp/k8s_dir:rw  -v$2:/tmp/test_results:rw --rm  --name $1 csi-sanity-test
```
oriyarde, line 7: instead of building from the separate `Dockerfile-csi-svc-test`, simply pass a different `PARAM_FILE` for `docker run` here

scripts/ci/community_csi_test_cleanup.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash -xe
 
-if [ `docker ps | grep $1 | wc -l` != 0 ] ; then
-    docker stop $1
+if [ `docker ps -a | grep $1 | wc -l` != 0 ] ; then
+    docker rm -f $1
 fi
```
oriyarde, line 5: is this because of some new gap?
could you add a newline at EOF while you're at it?

ArbelNathan (approved):
LGTM

Dockerfile-csi-svc-test
```diff
@@ -0,0 +1,36 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM golang:1.11
+
+RUN git clone https://github.com/kubernetes-csi/csi-test /usr/local/go/src/github.com/kubernetes-csi/csi-test
+RUN make -C /usr/local/go/src/github.com/kubernetes-csi/csi-test
```
matancarmeli7, line 18: Why not running it in th
```suggestion
RUN git clone https://github.com/kubernetes-csi/csi-test /usr/local/go/src/github.com/kubernetes-csi/csi-test
RUN make -C /usr/local/go/src/github.com/kubernetes-csi/csi-test
```

Dockerfile-csi-test
```diff
@@ -14,6 +14,8 @@
 
 FROM golang:1.11
 
+ARG CSI_PARAMS
```
oriyarde, line 17: consider adding a default value

scripts/ci/community_csi_test_cleanup.sh
oriyarde: and did the problem occur with `-a` but without `-f`? because it would be better to avoid `-f` if possible

scripts/ci/run_csi_test_client.sh
```diff
@@ -1,6 +1,12 @@
 #!/bin/bash -xe
 
-[ $# -ne 2 ] && { echo "Usage $0 : container_name folder_for_junit"  ; exit 1; }
+[ $# -ne 3 ] && { echo "Usage $0 : container_name folder_for_junit"  ; exit 1; }
+
+#/tmp/k8s_dir is the directory of the csi grpc\unix socket that shared between csi server and csi-test docker
```
oriyarde, line 5: this should be right above the docker commands line

scripts/ci/community_csi_test_cleanup.sh
barakda: didn't check it, @matancarmeli7 what do you think ?

Dockerfile-csi-test
barakda: already thought about it, no reason to add default since every storage array has its own param

oriyarde: how would the ds8k param look? must it be different from both existing param files?

barakda: as a9k AFAIK, no issues on that as was on svc

scripts/ci/community_csi_test_cleanup.sh
barakda: @matancarmeli7 answer through slack: 
"we are using -a in order to get also containers with the wanted name that are also stopped, because they are interrupt when you are trying to create a container with the same name as they do.
About the -f it's important because you want to delete running containers with the wanted name that can fail your run"

Dockerfile-csi-test
oriyarde: then I think having a non-svc default is better than nothing

scripts/ci/community_csi_test_cleanup.sh
oriyarde: I now consulted with @matancarmeli7 as well, we agree on `docker stop` + `docker rm` instead of `-f`, to avoid an error-prone SIGKILL, while not relying on existing `--rm`

oriyarde (approved):
LGTM

PR #321, opened 2021-05-12
CSI-2938 add node topology support

oriyarde:


Dockerfile-csi-node
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # Build stage
-FROM golang:1.12.6 as builder
+FROM golang:1.13 as builder
```
zingero, line 16: Will it affect the entire node?
If so, I suggest bumping the version in a different PR.

oriyarde: it will, but the reason to do it is coupled with this PR, and it's not a large diff, so I still think it is correct to include it here

oriyarde: @ArbelNathan what do you think?

zingero: I know we need it for this feature, but it can affect the entire node. Splitting it into two PRs can help us debug future issues.

oriyarde: it will also make the relation between the changes less obvious and add a merge blocker overhead

ArbelNathan (approved):
LGTM

Dockerfile-csi-node
ArbelNathan: I agree with both but because our operator is already on 1.13 I'm not really worried and the connection here is important for context.

zingero (dismissed):

node/pkg/driver/node_utils.go
```diff
@@ -425,7 +425,8 @@ func (n NodeUtils) GenerateNodeID(hostName string, fcWWNs []string, iscsiIQN str
 func (n NodeUtils) GetTopologyLabels(ctx context.Context, nodeName string) (map[string]string, error) {
 	kubeConfig, err := rest.InClusterConfig()
 	if err != nil {
-		return nil, err
+		logger.Infof("unable to load in-cluster configuration: %v", err)
+		return nil, nil
```
ArbelNathan, line 429: why is that OK? (no err returned)

oriyarde: for e.g. community tests

zingero: What's special with the community tests in this case? They don't need to get the error?

oriyarde: they fail if we error, since they check our CSI spec compliance, not our k8s compatibility.
even if we currently officially support only k8s-based orchestrators, our code should still be spec compliant, i.e. not fail just because it's outside k8s

zingero: I see.
The code doesn't explain this situation very well. I'm not a big fan of creating a variable called `csi_spec_expects_no error_in_this_case = nil;` and use it where needed, but I don't like the current code. WDYT?
Also, should the log severity be higher? Elaborate on the relevant errors, please.

oriyarde: I think the log severity tells that this external error is not considered an error on our side, which saves the need for the suggested variable

zingero: @ArbelNathan, WDYT?

oriyarde: I can add another log message `we are probably not in k8s, skipping topology retrieval`

zingero: Assuming things in logs can lead to questions when it happens in prod. env.
May I suggest adding `skipping topology retrieval. Are we in K8s?`?

oriyarde: I don't see a point for logs to ask rhetorical questions, could we settle for an additional `skipping topology retrieval` message?

zingero: sure

oriyarde: done

zingero (approved):

node/pkg/driver/node_utils.go
oriyarde: @ArbelNathan could this conversation be resolved?

ArbelNathan (approved):
LGTM

PR #324, opened 2021-05-16
Feature/csi 2899 Topology controller support

ArbelNathan:
adding support for secret config and by_system_id parameters in storageclass


controller/controller_server/csi_controller_server.py
```diff
@@ -60,14 +61,21 @@ def CreateVolume(self, request, context):
...
 
-            space_efficiency = request.parameters.get(config.PARAMETERS_SPACE_EFFICIENCY)
+        secrets = request.secrets
 
+        try:
+
+            user, password, array_addresses, secret_uid = utils.get_array_connection_info_from_secrets(
+                secrets=secrets,
+                topologies=topologies)
+            logger.info("chosen array_addresses: {}".format(array_addresses))
```
zingero, line 74: Usually, we don't write the parameter name (`array_addresses`) in the logs, but a description of the formatted parameter.
In this case:
`array addresses`

controller/controller_server/utils.py
```diff
@@ -14,36 +16,140 @@
...
+def _get_array_connection_info_from_secret(secret):
+    user = secret[config.SECRET_USERNAME_PARAMETER]
+    password = secret[config.SECRET_PASSWORD_PARAMETER]
+    array_addresses = secret[config.SECRET_ARRAY_PARAMETER].split(",")
     return user, password, array_addresses
 
 
-def get_volume_id(new_volume):
-    return _get_object_id(new_volume)
+def get_array_connection_info_from_secrets(secrets, topologies=None, secret_uid=None):
```
zingero, line 78: I'm not a big fan of functions that return tuples in Python because whenever we want to change the tuple, we need to change lots of places and QA should test it.
I suggest returning a dictionary and save the keys in "constant" variables (CAP_LETTER_VARIABLE) or creating a data class for that.

controller/controller_server/utils.py
```diff
@@ -14,36 +16,140 @@
...
-    password = secrets[config.SECRET_PASSWORD_PARAMETER]
-    array_addresses = secrets[config.SECRET_ARRAY_PARAMETER].split(",")
+def _is_topology_match(secret_topologies, node_topologies):
+    is_match = False
+    for topologies in secret_topologies:
+        logger.debug(
+            "Comparing topologies: object topologies: {},"
+            " node topologies: {}".format(topologies, node_topologies))
+        if all(item in node_topologies.items() for item in topologies.items()):
+            is_match = True
```
zingero, line 26: Can we return True and next line return False instead of having a local boolean variable?

controller/controller_server/utils.py
```diff
@@ -14,36 +16,140 @@
...
-    user = secrets[config.SECRET_USERNAME_PARAMETER]
-    password = secrets[config.SECRET_PASSWORD_PARAMETER]
-    array_addresses = secrets[config.SECRET_ARRAY_PARAMETER].split(",")
+def _is_topology_match(secret_topologies, node_topologies):
+    is_match = False
+    for topologies in secret_topologies:
+        logger.debug(
+            "Comparing topologies: object topologies: {},"
+            " node topologies: {}".format(topologies, node_topologies))
+        if all(item in node_topologies.items() for item in topologies.items()):
```
zingero, line 25: We compare only if we have an item in secret_topologies which is ⊆ node_topologies. We don't compare the other way: node_topologies ⊆ secret_topologies.
Does it bother us? If no, Why?

controller/controller_server/utils.py
```diff
@@ -14,36 +16,140 @@
...
+                                                          node_topologies=topologies)
+        else:
+            raise ValidationException(messages.invalid_secret_message)
+    return secret, secret_uid
+
+
+def _get_array_connection_info_from_secret(secret):
+    user = secret[config.SECRET_USERNAME_PARAMETER]
+    password = secret[config.SECRET_PASSWORD_PARAMETER]
+    array_addresses = secret[config.SECRET_ARRAY_PARAMETER].split(",")
```
zingero, line 74: I suggest putting the delimiter in `controller/controller_server/config.py`

controller/controller_server/utils.py
```diff
@@ -14,36 +16,140 @@
...
 
-def validate_secret(secret):
-    logger.debug("validating secrets")
-    if secret:
-        if not (config.SECRET_USERNAME_PARAMETER in secret and
-                config.SECRET_PASSWORD_PARAMETER in secret and
-                config.SECRET_ARRAY_PARAMETER in secret):
-            raise ValidationException(messages.invalid_secret_message)
+def _validate_secret_uid(uid):
+    if not re.match(config.SECRET_VALIDATION_REGEX, uid.lstrip()):
```
zingero, line 106: Why not stripping from both sides?

controller/controller_server/utils.py
```diff
@@ -14,36 +16,140 @@
...
+                raise ValidationException(messages.invalid_topology_value_message.format(topology_value))
 
-    else:
-        raise ValidationException(messages.secret_missing_message)
 
+def _validate_secret_config(secret_config_string):
+    secret_config = ast.literal_eval(secret_config_string)
+    for uid, secret_info in secret_config.items():
+        _validate_secret_uid(uid)
+        _validate_secret(secret_info)
```
zingero, line 139: I suggest merging these functions into one. If impossible, please elaborate.

controller/controller_server/utils.py
ArbelNathan: It's supposed to be this way.
`node_topologies` contains all of the worker topology keys the define it. the `secret_topologies` contain the bare minimum to identify the system by it, so if several workers have the same topologies or a common one, it's enough to only mention them to identify the system. the other way is not working cause we don't need to identify the worker..

controller/controller_server/utils.py
```diff
@@ -14,36 +16,140 @@
...
+
+
+def _validate_topologies(secret_topologies):
+    for topologies in secret_topologies:
+        for topology_key, topology_value in topologies.items():
+            split_topology = topology_key.split(config.PARAMETERS_TOPOLOGY_DELIMITER)
+            if len(split_topology) != 2:
+                raise ValidationException(messages.invalid_topology_key_message.format(topology_key))
+            prefix, suffix = split_topology
+            if not re.match(config.SECRET_VALIDATION_REGEX, suffix.lstrip()):
```
zingero, line 127: Same

controller/controller_server/utils.py
```diff
@@ -14,36 +16,140 @@
...
+                raise ValidationException(messages.invalid_topology_key_message.format(suffix))
+            if prefix not in config.TOPOLOGY_PREFIXES:
+                raise ValidationException(messages.invalid_topology_key_message.format(prefix))
+            if not re.match(config.SECRET_VALIDATION_REGEX, topology_value.lstrip()):
```
zingero, line 131: Same (regarding stripping)

controller/controller_server/utils.py
ArbelNathan: `secret` is not bound by `uid`, they are 2 different things, in the topology feature they are connected, without it, I'd have no meaning and does not exist

controller/controller_server/utils.py
```diff
@@ -339,12 +448,12 @@ def validate_unpublish_volume_request(request):
     logger.debug("validating unpublish volume request")
 
     logger.debug("validating volume id")
-    if len(request.volume_id.split(config.PARAMETERS_OBJECT_ID_DELIMITER)) != 2:
+    if len(request.volume_id.split(config.PARAMETERS_OBJECT_ID_DELIMITER)) not in (2, 3):
```
zingero, line 451: (2, 3) is similar to using a magic number. I suggest wrapping it with a CAP_LETTER_VARIABLE.
Also, why using a tuple and not a set?

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -1198,9 +1249,16 @@ def test_map_volume_by_initiators_exceptions(self, storage_agent, map_volume_by_
...
+class TestControllerServerUnPublishVolume(AbstractControllerTest):
+
+    def get_create_object_method(self):
+        return self.servicer.ControllerUnpublishVolume
+
+    def get_create_object_response_method(self):
+        return csi_pb2.ControllerUnpublishVolumeResponse
 
     def setUp(self):
+        super().setUp()
```
zingero, line 1261: Nice!

controller/tests/controller_server/utils_test.py
```diff
@@ -22,32 +23,116 @@ def setUp(self):
...
+
+    def test_validate_secrets_with_config_invalid_parameters(self):
+        self._test_validate_secrets_with_config_invalid_parameters(uid="u-1")
+        self._test_validate_secrets_with_config_invalid_parameters(uid="u:1")
+        self._test_validate_secrets_with_config_invalid_parameters(uid="u1+")
+        self._test_validate_secrets_with_config_invalid_parameters(uid="u1*")
+        self._test_validate_secrets_with_config_invalid_parameters(uid="u-1(")
+        self._test_validate_secrets_with_config_invalid_parameters(uid="u/1")
+        self._test_validate_secrets_with_config_invalid_parameters(uid="u=1")
+        self._test_validate_secrets_with_config_invalid_parameters(uid=" ")
```
zingero, line 92: I suggest adding an empty string test and a None test if it can theoretically happen.

controller/controller_server/csi_controller_server.py
ArbelNathan: removed

controller/controller_server/utils.py
ArbelNathan: done

controller/controller_server/utils.py

controller/controller_server/utils.py

controller/controller_server/utils.py

controller/controller_server/utils.py

controller/controller_server/utils.py

controller/controller_server/utils.py

controller/tests/controller_server/utils_test.py
ArbelNathan: nice, Im not sure that None is possible but added another check for both

zingero (approved):

controller/controller_server/utils.py
```diff
@@ -273,12 +386,15 @@ def get_snapshot_id_info(snapshot_id):
...
+        array_type, object_id = splitted_object_id
+    elif len(splitted_object_id) == 3:
+        array_type, secret_uid, object_id = splitted_object_id
+    else:
         raise ObjectIdError(object_type, full_object_id)
-
-    array_type, object_id = splitted_object_id
     logger.debug("volume id : {0}, array type :{1}".format(object_id, array_type))
-    return array_type, object_id
+    return array_type, object_id, secret_uid
```
oriyarde, line 397: returning `secret_uid` in the middle would be more aligned with the order of appearances in the string

controller/controller_server/utils.py
```diff
@@ -8,42 +10,153 @@
...
+        if _is_topology_match(secret_topologies, node_topologies):
+            return secret_info, secret_uid
+
+
+def get_secret_by_uid(secret_config_string, secret_uid):
+    secret_config = ast.literal_eval(secret_config_string)
+    return secret_config.get(secret_uid)
+
+
+def get_pool_by_uid(pools, uid):
```
oriyarde, line 52: consider renaming `pools` to something like `raw_pools_by_<final_name_of_secret_uid>`

controller/controller_server/utils.py
```diff
@@ -8,42 +10,153 @@
...
+
+def _validate_secret_config(secret_config_string):
+    secret_config = ast.literal_eval(secret_config_string)
+    for uid, secret_info in secret_config.items():
+        if uid and secret_info:
+            _validate_secret_uid(uid)
+            _validate_secret(secret_info)
+            topologies = secret_info.get(config.SECRET_SUPPORTED_TOPOLOGIES_PARAMETER)
+            if topologies:
+                _validate_topologies(topologies)
```
oriyarde, line 147: shouldn't these last 3 lines be inside the 1st?

controller/controller_server/utils.py
```diff
@@ -8,42 +10,153 @@
...
+            split_topology = topology_key.split(config.PARAMETERS_TOPOLOGY_DELIMITER)
+            if len(split_topology) != 2:
+                raise ValidationException(messages.invalid_topology_key_message.format(topology_key))
+            prefix, suffix = split_topology
+            if not _is_imported_string_valid(suffix):
+                raise ValidationException(messages.invalid_topology_key_message.format(suffix))
+            if prefix not in config.TOPOLOGY_PREFIXES:
+                raise ValidationException(messages.invalid_topology_key_message.format(prefix))
+            if not _is_imported_string_valid(topology_value):
+                raise ValidationException(messages.invalid_topology_value_message.format(topology_value))
```
oriyarde, line 136: I think these validations should be aligned to only enforce the CSI spec, i.e. without any SP-specific validations

controller/controller_server/utils.py
```diff
@@ -8,42 +10,153 @@
...
+
+def _validate_secret_uid(uid):
+    if not _is_imported_string_valid(uid):
+        raise ValidationException(messages.invalid_uid_value_message.format(uid))
+    if len(uid) > config.SECRET_UID_MAX_LENGTH:
+        raise ValidationException(
+            messages.parameter_length_is_too_long.format("secret_uid", uid, config.SECRET_UID_MAX_LENGTH))
+
+
+def _validate_secret(secret):
```
oriyarde, line 117: I know this is existing code, but we should prefer spec terms over k8s terms, so this dict should be `secrets`, and not `secret`

controller/controller_server/utils.py
```diff
@@ -8,42 +10,153 @@
...
 
-    else:
-        raise ValidationException(messages.secret_missing_message)
 
+def validate_secrets(secrets):
+    logger.debug("validating secrets")
+    if not secrets:
+        raise ValidationException(messages.secret_missing_message)
+    if secrets.get(config.SECRET_CONFIG_PARAMETER):
+        _validate_secret_config(secrets.get(config.SECRET_CONFIG_PARAMETER))
```
oriyarde, line 157:
```suggestion
    secret_config = secrets.get(config.SECRET_CONFIG_PARAMETER, None)
    if secret_config:
        _validate_secret_config(secret_config)
```

controller/controller_server/utils.py
```diff
@@ -8,42 +10,153 @@
...
+        else:
             raise ValidationException(messages.invalid_secret_message)
 
-    else:
-        raise ValidationException(messages.secret_missing_message)
 
+def validate_secrets(secrets):
+    logger.debug("validating secrets")
+    if not secrets:
+        raise ValidationException(messages.secret_missing_message)
```
oriyarde, line 155: then why do we usually do `if request.secrets:` before calling this function?

controller/tests/controller_server/utils_test.py
```diff
@@ -22,32 +23,118 @@ def setUp(self):
...
+        self._test_validate_secrets_with_config_invalid_parameters(uid="")
+        self._test_validate_secrets_with_config_invalid_parameters(uid=None)
+        self._test_validate_secrets_with_config_invalid_parameters(uid="a" * 91)
+        self._test_validate_secrets_with_config_invalid_parameters(topology="topology.kubernetes.io")
+        self._test_validate_secrets_with_config_invalid_parameters(topology="topology.kubernetes/test")
+        self._test_validate_secrets_with_config_invalid_parameters(topology="topology.kubernetes.io/t-est")
+        self._test_validate_secrets_with_config_invalid_parameters(value="a*")
```
oriyarde, line 99: consider using a `for` loop

controller/tests/controller_server/utils_test.py
```diff
@@ -22,32 +23,118 @@ def setUp(self):
...
+
+    def _test_get_pool_from_parameters(self, parameters, pool_expected=pool, uid=None):
+        pool_respond = utils.get_pool_from_parameters(parameters, uid)
+        self.assertEqual(pool_respond, pool_expected)
```
oriyarde, line 103: consider renaming to `actual_pool` and `expected_pool` where applicable

controller/controller_server/utils.py
```diff
@@ -14,36 +16,140 @@
...
+        if all(item in node_topologies.items() for item in topologies.items()):
+            is_match = True
+    return is_match
+
+
+def get_volume_topologies(request):
+    if request.HasField(config.REQUEST_ACCESSIBILITY_REQUIREMENTS_FIELD):
+        accessibility_requirements = request.accessibility_requirements
+        if accessibility_requirements.preferred:
+            topologies = accessibility_requirements.preferred[0].segments
```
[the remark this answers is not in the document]
ArbelNathan: I dont care about them

controller/controller_server/config.py
```diff
@@ -12,8 +12,13 @@
 SECRET_USERNAME_PARAMETER = "username"
 SECRET_PASSWORD_PARAMETER = "password"
 SECRET_ARRAY_PARAMETER = "management_address"
+SECRET_CONFIG_PARAMETER = "config"
+SECRET_SUPPORTED_TOPOLOGIES_PARAMETER = "supported_topologies"
+SECRET_UID_MAX_LENGTH = 90
```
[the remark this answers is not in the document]
ArbelNathan: this is not Array ID, it got nothing in common to the actual storage

controller/controller_server/csi_controller_server.py
```diff
@@ -60,14 +61,21 @@ def CreateVolume(self, request, context):
...
+            pool = utils.get_pool_from_parameters(parameters=request.parameters, secret_uid=secret_uid)
+            if not pool:
+                raise ValidationException(controller_messages.wrong_pool_passed_message)
+            space_efficiency = request.parameters.get(config.PARAMETERS_SPACE_EFFICIENCY)
```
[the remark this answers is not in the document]
ArbelNathan: moved? when was it outside?

controller/controller_server/utils.py
```diff
@@ -8,42 +10,153 @@
...
+            logger.info("Chosen volume topologies: {}".format(topologies))
+            return topologies
+
+
+def get_secret_by_topologies(secret_config_string, node_topologies):
+    secret_config = ast.literal_eval(secret_config_string)
```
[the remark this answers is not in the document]
ArbelNathan: cause its sensitive to " and '

controller/controller_server/utils.py
ArbelNathan: raw? can you explain?

controller/controller_server/utils.py
ArbelNathan: what validations? what part of the spec?

controller/controller_server/csi_controller_server.py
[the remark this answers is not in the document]
oriyarde: actually it [was done](https://github.com/IBM/ibm-block-csi-driver/pull/305/files#diff-1d72325bc9ed2e0861808fe985ff7e74edfbba4533a5a71f12b8fd498bb2647c) before this PR, mind moving back outside whatever is possible?

controller/controller_server/utils.py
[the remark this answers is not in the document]
oriyarde: why not?

controller/controller_server/utils.py
ArbelNathan: I guess the `if request.secrets:` is redundant and should be removed. (as it fixed in [validate volume capabilities](https://github.com/IBM/ibm-block-csi-driver/pull/212/files#diff-cfa9702e443362434f95936cee6df91703ec484838eb64262231cab9fd03466bR156))

controller/controller_server/config.py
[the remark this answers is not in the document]
oriyarde: well, it has something to do with the storage, since it's meant to distinguish between different `connection info`s.
perhaps connection info id/key?

controller/tests/controller_server/utils_test.py
ArbelNathan: yes, but it will be only for the first uid checks not for the `"a" * 91` , `None` and empty strings..

controller/controller_server/utils.py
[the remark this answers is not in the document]
oriyarde: do we want it not to be?
json syntax is more common than python syntax, in configuration files

controller/controller_server/utils.py
oriyarde: i.e. the raw, unprocessed, unparsed string

controller/controller_server/utils.py
oriyarde: the validations in these lines, the part of the spec which states the string value constraints

controller/tests/controller_server/utils_test.py
oriyarde: better than nothing, but why is that?

controller/controller_server/controller_types.py
```diff
@@ -0,0 +1,9 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class Secret:
```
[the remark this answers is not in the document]
ArbelNathan: for every secret reference in the code?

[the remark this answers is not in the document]
oriyarde: at least in this PR, yes.
I think there is currently a great mix-up in the terms here

controller/controller_server/utils.py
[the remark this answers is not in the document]
oriyarde: in addition, `literal_eval` probably has greater security risks comparing to `json.loads()`.
@zingero what do you think?

controller/controller_server/config.py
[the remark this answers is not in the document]
oriyarde: user can also write the same files to volumes with different wwpns - that doesn't mean a wwpn is not a volume "id"

controller/controller_server/utils.py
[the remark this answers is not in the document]
ArbelNathan: in every "dict" value from the storageclas, we would have to replace all " and ' in order to make it work with json.loads() to end with the same result.

controller/tests/controller_server/utils_test.py
ArbelNathan: it depends on how do you want to do the for.. to iterate on invalid symbols or to iterate on a list of all the values

controller/controller_server/config.py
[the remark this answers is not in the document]
ArbelNathan: but it is for different volumes. one wwpn for one volume, this is not one key per storage system.

controller/controller_server/csi_controller_server.py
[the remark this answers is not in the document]
ArbelNathan: how is it relate to this scope? the diff is big enough in my opinion

controller/controller_server/config.py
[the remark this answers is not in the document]
oriyarde: my point is that both examples describe identical instances with different ids.
from the secret config perspective, different ids *could* point to identical info, but each unique info content would still represent a different system

controller/controller_server/csi_controller_server.py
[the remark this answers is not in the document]
oriyarde: I agree, resolving

controller/controller_server/utils.py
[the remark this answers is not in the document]
oriyarde: what if a user specifies allowed topologies? shouldn't we also ensure the system we choose supports the other required topologies?
what about CSI spec requirements (not relying on k8s details)?

controller/controller_server/utils.py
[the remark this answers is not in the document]
oriyarde: sounds good to me

controller/tests/controller_server/utils_test.py
oriyarde: why not all values?
```suggestion
        uids = ["u-1", "u:1", "u1+", "u1*", "u-1(", "u/1", "u=1", " ", "", None, "a" * 91]
        for uid in uids:
            self._test_validate_secrets_with_config_invalid_parameters(uid=uid)
```

controller/controller_server/utils.py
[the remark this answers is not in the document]
ArbelNathan: allowed topologies do nothing that we care for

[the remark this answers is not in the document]
oriyarde: what if a user has:
config: `{system1: suppotedTopologies: [{R1: Z1}], system2: suppotedTopologies: [{R1: Z1}, {R1: Z2}]}`
sc: `allowedTopologes: [{R1: Z1}, {R1: Z2}]`
worker1: `R1: Z1`
worker2: `R1: Z2`

shouldn't we ensure the volume is created in `system2` and not `system1`?

controller/controller_server/config.py
```diff
@@ -12,8 +12,13 @@
...
+SECRET_VALIDATION_REGEX = '[a-zA-Z0-9 _.]+$'
```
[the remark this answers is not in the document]
ArbelNathan: done

controller/controller_server/controller_types.py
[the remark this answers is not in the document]
ArbelNathan: done

controller/controller_server/csi_controller_server.py
```diff
@@ -60,17 +61,24 @@ def CreateVolume(self, request, context):
...
+                topologies=topologies)
+            pool = utils.get_pool_from_parameters(parameters=request.parameters, secret_uid=secret.uid)
+            if not pool:
+                raise ValidationException(controller_messages.wrong_pool_passed_message)
+            space_efficiency = request.parameters.get(config.PARAMETERS_SPACE_EFFICIENCY)
             # TODO : pass multiple array addresses
-            array_type = detect_array_type(array_addresses)
-            with get_agent(user, password, array_addresses, array_type).get_mediator() as array_mediator:
+            array_type = detect_array_type(secret.array_addresses)
+            with get_agent(secret.user, secret.password, secret.array_addresses,
```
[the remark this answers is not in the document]

controller/controller_server/csi_controller_server.py
```diff
@@ -198,15 +208,16 @@ def DeleteVolume(self, request, context):
         try:
             utils.validate_delete_volume_request(request)
 
-            user, password, array_addresses = utils.get_array_connection_info_from_secret(secrets)
-
             try:
-                array_type, vol_id = utils.get_volume_id_info(request.volume_id)
+                array_type, vol_id, secret_uid = utils.get_volume_id_info(request.volume_id)
```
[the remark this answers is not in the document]

controller/controller_server/csi_controller_server.py
```diff
@@ -386,15 +401,18 @@ def CreateSnapshot(self, request, context):
...
         try:
-            _, volume_id = utils.get_volume_id_info(source_volume_id)
-            array_type = detect_array_type(array_addresses)
-            with get_agent(user, password, array_addresses, array_type).get_mediator() as array_mediator:
+            _, volume_id, secret_uid = utils.get_volume_id_info(source_volume_id)
+            secret = utils.get_array_connection_info_from_secrets(
+                secrets,
+                secret_uid=secret_uid)
+            pool = utils.get_pool_from_parameters(parameters=request.parameters, secret_uid=secret_uid)
+            array_type = detect_array_type(secret.array_addresses)
```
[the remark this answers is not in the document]

controller/controller_server/csi_controller_server.py
```diff
@@ -60,17 +61,24 @@ def CreateVolume(self, request, context):
...
-            user, password, array_addresses = utils.get_array_connection_info_from_secret(secrets)
 
-            pool = request.parameters[config.PARAMETERS_POOL]
+        topologies = utils.get_volume_topologies(request)
 
-            space_efficiency = request.parameters.get(config.PARAMETERS_SPACE_EFFICIENCY)
+        secrets = request.secrets
 
+        try:
+
```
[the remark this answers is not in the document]

controller/controller_server/utils.py
```diff
@@ -8,42 +10,153 @@
...
-def get_array_connection_info_from_secret(secrets):
-    user = secrets[config.SECRET_USERNAME_PARAMETER]
-    password = secrets[config.SECRET_PASSWORD_PARAMETER]
-    array_addresses = secrets[config.SECRET_ARRAY_PARAMETER].split(",")
+def _is_topology_match(secret_topologies, node_topologies):
+    for topologies in secret_topologies:
+        logger.debug(
+            "Comparing topologies: object topologies: {},"
+            " node topologies: {}".format(topologies, node_topologies))
+        if all(item in node_topologies.items() for item in topologies.items()):
```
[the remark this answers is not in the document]
ArbelNathan: nice

controller/controller_server/utils.py
```diff
@@ -8,42 +10,153 @@
...
 
 
-def get_array_connection_info_from_secret(secrets):
-    user = secrets[config.SECRET_USERNAME_PARAMETER]
-    password = secrets[config.SECRET_PASSWORD_PARAMETER]
-    array_addresses = secrets[config.SECRET_ARRAY_PARAMETER].split(",")
+def _is_topology_match(secret_topologies, node_topologies):
+    for topologies in secret_topologies:
+        logger.debug(
+            "Comparing topologies: object topologies: {},"
```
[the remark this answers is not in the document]

controller/controller_server/utils.py

controller/controller_server/utils.py
ArbelNathan: no need anymore

controller/controller_server/utils.py

controller/controller_server/utils.py
ArbelNathan: its aligned with spec

controller/controller_server/utils.py

controller/controller_server/utils.py

controller/tests/controller_server/utils_test.py
ArbelNathan: ok, nice, done

controller/tests/controller_server/utils_test.py
ArbelNathan: done

controller/controller_server/utils.py
oriyarde: I mean that they should also allow everything that the spec allows, e.g. not specific prefixes

controller/controller_server/config.py
```diff
@@ -12,8 +12,14 @@
 SECRET_USERNAME_PARAMETER = "username"
 SECRET_PASSWORD_PARAMETER = "password"
 SECRET_ARRAY_PARAMETER = "management_address"
+SECRET_CONFIG_PARAMETER = "config"
+SECRET_SUPPORTED_TOPOLOGIES_PARAMETER = "supported_topologies"
+# SECRET_UID_MAX_LENGTH = max string response - length of volume wwn - max length of array type key - delimiters
```
oriyarde, line 17: use consts instead of comment

controller/controller_server/config.py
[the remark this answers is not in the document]
oriyarde: I take that as a "no"

controller/controller_server/csi_controller_server.py
```diff
@@ -454,15 +476,16 @@ def DeleteSnapshot(self, request, context):
         secrets = request.secrets
         try:
             utils.validate_delete_snapshot_request(request)
-            user, password, array_addresses = utils.get_array_connection_info_from_secret(secrets)
             try:
-                array_type, snapshot_id = utils.get_snapshot_id_info(request.snapshot_id)
+                volume_id_info = utils.get_snapshot_id_info(request.snapshot_id)
```
oriyarde, line 480: perhaps this class name and instances should instead be named "*object* id info", sorry for the double work.
it can be in a separate ticket, if it helps

controller/controller_server/controller_types.py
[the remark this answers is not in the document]
oriyarde: I still see this mix-up, e.g. in `_get_secret_from_secrets`.
AFAIK, in spec terminology, `secret` is e.g. `username`, so e.g. `secret[config.SECRET_USERNAME_PARAMETER]` should be `secrets[config.SECRET_USERNAME_PARAMETER]`.
hope my suggestion is clear

controller/controller_server/utils.py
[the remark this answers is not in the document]
oriyarde: why `volume`? shouldn't it be `system` or something?

controller/controller_server/utils.py
oriyarde: same for the order of the dataclass members (i.e. constructor parameters order here)

controller/controller_server/utils.py
oriyarde: I meant that `_validate_topologies` be called from inside `_validate_secret` (not call 2 functions in the same line)

controller/controller_server/utils.py
oriyarde: I still see the parameter name is `secret` (singular)

controller/tests/controller_server/utils_test.py
```diff
@@ -22,32 +23,112 @@ def setUp(self):
...
+            self._test_validate_secrets_with_config_invalid_parameters(topology=topology)
+
+        self._test_validate_secrets_with_config_invalid_parameters(value="a*")
+
+    def _test_get_array_connection_info_from_secrets(self, secrets, topologies=None, secret_uid=None):
+        response = utils.get_array_connection_info_from_secrets(
+            secrets=secrets,
+            topologies=topologies,
+            secret_uid=secret_uid)
+        secret = response
```
oriyarde, line 100: consider using something like `actual_connection_info` to begin with

controller/controller_server/controller_types.py
```diff
@@ -0,0 +1,23 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class ArrayConnectionInfo:
+    user: str
+    password: str
+    array_addresses: list
```
oriyarde, line 8: I think `array_addresses` should be first, as in the original secret of our yaml examples - it's more aligned with their logical hierarchy

controller/controller_server/utils.py
```diff
@@ -8,42 +10,163 @@
...
+def get_volume_parameters(parameters, secret_uid):
+    return get_object_parameters(parameters, config.PARAMETERS_VOLUME_NAME_PREFIX, secret_uid)
+
+
+def get_snapshot_parameters(parameters, secret_uid):
+    return get_object_parameters(parameters, config.PARAMETERS_SNAPSHOT_NAME_PREFIX, secret_uid)
+
+
+def get_object_parameters(parameters, prefix_param_name, secret_uid):
+    raw_capabilities_by_system = parameters.get(config.PARAMETERS_BY_SYSTEM)
```
oriyarde, line 88: I don't think e.g. `prefix` is a capability

controller/controller_server/utils.py
```diff
@@ -8,42 +10,163 @@
...
+    system_capabilities = {}
+    if raw_capabilities_by_system and secret_uid:
+        capabilities_by_system = ast.literal_eval(raw_capabilities_by_system)
+        system_capabilities = capabilities_by_system.get(secret_uid, {})
+    return VolumeParameters(
+        pool=system_capabilities.get(config.PARAMETERS_POOL, parameters.get(config.PARAMETERS_POOL)),
+        space_efficiency=system_capabilities.get(config.PARAMETERS_SPACE_EFFICIENCY,
+                                                 parameters.get(config.PARAMETERS_SPACE_EFFICIENCY)),
+        prefix=system_capabilities.get(prefix_param_name,
+                                       parameters.get(prefix_param_name)))
```
oriyarde, line 98: please split this train wreck into meaningful var names, e.g. `default_` prefix for var names of default values

oriyarde (commented):
don't forget to update the PR description

controller/controller_server/config.py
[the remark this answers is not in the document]
ArbelNathan: done - `system_id`

controller/controller_server/controller_types.py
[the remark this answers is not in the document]
ArbelNathan: done

controller/controller_server/utils.py
ArbelNathan: no more checking that

controller/controller_server/utils.py

controller/controller_server/controller_types.py
ArbelNathan: done

controller/tests/controller_server/utils_test.py
ArbelNathan: done

controller/controller_server/csi_controller_server.py
ArbelNathan: done

controller/controller_server/config.py
ArbelNathan: done

controller/controller_server/utils.py

controller/controller_server/utils.py

controller/controller_server/config.py
```diff
@@ -9,11 +9,23 @@
 VOLUME_CAPABILITIES_FIELD_ACCESS_TYPE_MOUNT = 'mount'
 VOLUME_CAPABILITIES_FIELD_ACCESS_TYPE_BLOCK = 'block'
 
+MAX_STRING_RESPONSE_LENGTH = 128
```
oriyarde, line 12: `MAX_RESPONSE_STRING_LENGTH`?

controller/controller_server/config.py
```diff
@@ -9,11 +9,23 @@
...
+
+VOLUME_WWN_LENGTH = 32
+MAX_ARRAY_TYPE_KEY_LENGTH = 4
```
oriyarde, line 15: I think `KEY` can be omitted

controller/controller_server/config.py
```diff
@@ -9,11 +9,23 @@
...
+DELIMITERS_IN_VOLUME_ID = 2
```
oriyarde, line 16: `NUMBER_OF_...`

controller/controller_server/utils.py
```diff
@@ -8,43 +10,147 @@
...
+
+def get_secret_by_topologies(raw_secret_config, node_topologies):
+    secret_config = ast.literal_eval(raw_secret_config)
+    for system_id, secret_info in secret_config.items():
+        secret_topologies = secret_info.get(config.SECRET_SUPPORTED_TOPOLOGIES_PARAMETER)
+        if _is_topology_match(secret_topologies, node_topologies):
+            return secret_info, system_id
+
+
+def get_secrets_by_system_id(secret_config_string, system_id):
```
oriyarde, line 47: consider renaming `secret_config_string` to `raw_secret_config` here as well

controller/controller_server/utils.py
```diff
@@ -8,43 +10,147 @@
...
+    if request.HasField(config.REQUEST_ACCESSIBILITY_REQUIREMENTS_FIELD):
+        accessibility_requirements = request.accessibility_requirements
+        if accessibility_requirements.preferred:
+            topologies = accessibility_requirements.preferred[0].segments
+            logger.info("Chosen volume topologies: {}".format(topologies))
+            return topologies
+
+
+def get_secret_by_topologies(raw_secret_config, node_topologies):
+    secret_config = ast.literal_eval(raw_secret_config)
```
oriyarde, line 40: consider wrapping this line with try-except, for a user friendly parsing error message, and moving it all to a separate function

controller/controller_server/controller_types.py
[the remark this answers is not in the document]
oriyarde: I still see `secret` (singular) in `get_secret_by_topologies`

controller/controller_server/utils.py
```diff
@@ -8,43 +10,147 @@
...
+
+def _get_system_info_from_secrets(secrets, topologies=None, system_id=None):
+    secret_config = secrets.get(config.SECRET_CONFIG_PARAMETER)
+    system_secrets = secrets
+    if secret_config:
+        if system_id:
+            system_secrets = get_secrets_by_system_id(secret_config_string=secret_config, system_id=system_id)
+        elif topologies:
+            system_secrets, system_id = get_secret_by_topologies(raw_secret_config=secret_config,
+                                                                 node_topologies=topologies)
```
oriyarde, line 60: `raw_secret_config` can be parsed right after `if secret_config:`, so none of the other two functions would need to worry about parsing it

controller/controller_server/controller_types.py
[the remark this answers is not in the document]
ArbelNathan: sorry, I didn't meant to right "done" there, still in work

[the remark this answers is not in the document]
ArbelNathan: done now

controller/controller_server/utils.py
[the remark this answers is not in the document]

controller/controller_server/utils.py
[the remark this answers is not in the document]

controller/controller_server/utils.py
ArbelNathan: it won't do here cause secret can be the B.C one and not the config's so topology won't be necessarily related in `_validate_secret`

controller/controller_server/utils.py

controller/controller_server/config.py
ArbelNathan: done

controller/controller_server/config.py
ArbelNathan: done

controller/controller_server/config.py
ArbelNathan: done

controller/controller_server/utils.py

controller/controller_server/utils.py

controller/controller_server/utils.py

controller/controller_server/utils.py
```diff
@@ -8,43 +10,164 @@
...
 from controller.controller_server.errors import ObjectIdError, ValidationException
 from controller.csi_general import csi_pb2
 
 logger = get_stdout_logger()
 
 
-def get_array_connection_info_from_secret(secrets):
+def _parse_raw_input(raw_input):
+    try:
+        parsed_input = json.loads(raw_input)
```
oriyarde, line 22: consider replacing `input` with `json` wherever applicable

controller/controller_server/messages.py
```diff
@@ -3,8 +3,12 @@
 ObjectIdError_message = "Wrong {0} id format : {1}"
 
 # validation error messages
-invalid_secret_message = "invalid secret was passed"
-secret_missing_message = 'secret is missing'
+invalid_secrets_message = "invalid secrets was passed"
+invalid_topology_key_message = "invalid topology key was passed: {}"
+invalid_topology_value_message = "invalid topology value was passed: {}"
+invalid_system_id_value_message = "invalid system id value was passed: {}"
+invalid_non_json_parameter_message = "invalid non json parameter was passed: {}."
```
oriyarde, line 10: consider omitting `non`

controller/controller_server/utils.py
```diff
@@ -8,43 +10,164 @@
...
+
+def get_system_info_by_topologies(secrets_config, node_topologies):
+    for system_id, system_info in secrets_config.items():
+        system_topologies = system_info.get(config.SECRET_SUPPORTED_TOPOLOGIES_PARAMETER)
+        if _is_topology_match(system_topologies, node_topologies):
+            return system_info, system_id
+
+
+def _get_system_info_by_system_id(secrets_config, system_id):
+    return secrets_config.get(system_id)
```
oriyarde, line 55: consider removing this function

controller/controller_server/messages.py
```diff
@@ -3,8 +3,12 @@
 ObjectIdError_message = "Wrong {0} id format : {1}"
 
 # validation error messages
-invalid_secret_message = "invalid secret was passed"
-secret_missing_message = 'secret is missing'
+invalid_secrets_message = "invalid secrets was passed"
```
oriyarde, line 6: please replace the usages of this generic message with more informative messages, according to each specific issue with the secret, and try to keep it singular

controller/controller_server/utils.py
```diff
@@ -8,43 +10,164 @@
...
 
 
-def validate_secret(secret):
-    logger.debug("validating secrets")
-    if secret:
-        if not (config.SECRET_USERNAME_PARAMETER in secret and
-                config.SECRET_PASSWORD_PARAMETER in secret and
-                config.SECRET_ARRAY_PARAMETER in secret):
-            raise ValidationException(messages.invalid_secret_message)
+def _is_imported_string_valid(imported_string):
```
oriyarde, line 122: rename to something more specific, e.g. `_is_system_id_valid`/`_is_valid_system_id`

controller/controller_server/messages.py
```diff
@@ -3,8 +3,12 @@
...
+invalid_topology_key_message = "invalid topology key was passed: {}"
+invalid_topology_value_message = "invalid topology value was passed: {}"
+invalid_system_id_value_message = "invalid system id value was passed: {}"
```
oriyarde, line 9: consider including the enforcing regex

controller/controller_server/utils.py
```diff
@@ -8,43 +10,164 @@
...
+    if not (config.SECRET_USERNAME_PARAMETER in secrets and
+            config.SECRET_PASSWORD_PARAMETER in secrets and
+            config.SECRET_ARRAY_PARAMETER in secrets):
+        raise ValidationException(messages.invalid_secrets_message)
+
 
+def _validate_topologies(topologies):
+    if topologies:
+        for topology in topologies:
+            if not topology:
```
oriyarde, line 144: would `if not all(topologies)` work?

controller/controller_server/utils.py
```diff
@@ -8,43 +10,164 @@
...
     else:
-        raise ValidationException(messages.secret_missing_message)
+        raise ValidationException(messages.invalid_secrets_message)
 
-    logger.debug("secret validation finished")
+
+def _validate_secrets_config(secrets_config):
+    for system_id, system_info in secrets_config.items():
+        if system_id and system_info:
+            _validate_topologies(system_info.get(config.SECRET_SUPPORTED_TOPOLOGIES_PARAMETER))
```
oriyarde, line 153: please avoid multiple calls in the same line

controller/controller_server/utils.py
```diff
@@ -8,43 +10,164 @@
...
+            _validate_system_id(system_id)
+            _validate_secrets(system_info)
```
oriyarde, line 155: I would order them by order of appearance in the example, i.e. system id, system info "secrets", then topologies

controller/controller_server/utils.py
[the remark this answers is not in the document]
oriyarde: according to [the spec]([url](https://github.com/container-storage-interface/spec/blob/v1.4.0/spec.md)):
>   // If requisite is specified, the provisioned volume MUST be
  // accessible from at least one of the requisite topologies.

`Z1` is at least one, so the answer is "no", but still:
>   // If the list of requisite topologies is specified and the SP is
  // unable to to make the provisioned volume available from any of the
  // requisite topologies it MUST fail the CreateVolume call. 

which means we are expected to verify this, and not rely on k8s implementation details to take care of it.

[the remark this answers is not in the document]
oriyarde: what about the rest of the preferred? did you manage to test without strict topology?

controller/controller_server/csi_controller_server.py
```diff
@@ -729,7 +751,8 @@ def _get_source_type_and_id(self, request):
                 source_type = config.VOLUME_TYPE_NAME
             else:
                 return None, None
-            _, object_id = utils.get_object_id_info(source_id, source_type)
+            volume_id_info = utils.get_object_id_info(source_id, source_type)
```
oriyarde, line 754: `object_id_info`?

controller/controller_server/messages.py
oriyarde, line 9: `passed` has a positive connotation, consider using `got an invalid ...:`.
and for system id - no need for `value` (nor `key`)

controller/controller_server/messages.py
```diff
@@ -3,8 +3,12 @@
...
+invalid_non_json_parameter_message = "invalid non json parameter was passed: {}."
+secrets_missing_message = 'secrets is missing'
```
oriyarde, line 11: from the user perspective (and grammatically), this should be singular

controller/controller_server/utils.py
```diff
@@ -8,43 +10,164 @@
...
+def get_volume_topologies(request):
+    if request.HasField(config.REQUEST_ACCESSIBILITY_REQUIREMENTS_FIELD):
+        accessibility_requirements = request.accessibility_requirements
+        if accessibility_requirements.preferred:
+            topologies = accessibility_requirements.preferred[0].segments
+            logger.info("Chosen volume topologies: {}".format(topologies))
+            return topologies
+
+
+def get_system_info_by_topologies(secrets_config, node_topologies):
```
oriyarde, line 47: `by` sounds like a mapping to me, consider using `for` instead

controller/controller_server/utils.py
```diff
@@ -8,43 +10,164 @@
...
     password = secrets[config.SECRET_PASSWORD_PARAMETER]
-    array_addresses = secrets[config.SECRET_ARRAY_PARAMETER].split(",")
-    return user, password, array_addresses
+    array_addresses = secrets[config.SECRET_ARRAY_PARAMETER].split(config.PARAMETERS_ARRAY_ADDRESSES_DELIMITER)
+    return ArrayConnectionInfo(array_addresses=array_addresses, user=user, password=password, system_id=system_id)
+
+
+def get_array_connection_info_from_secrets(secrets, topologies=None, system_id=None):
+    system_info, system_id = _get_system_info_from_secrets(secrets, topologies, system_id)
+    return _get_array_connection_info_from_system_info(system_info, system_id)
```
oriyarde, line 82: so if there is no topology match, we get a non-indicative key error in `secrets[config.SECRET_USERNAME_PARAMETER]`?

controller/controller_server/utils.py
```diff
@@ -116,15 +239,15 @@ def validate_create_volume_request(request):
...
     if request.secrets:
-        validate_secret(request.secrets)
+        validate_secrets(request.secrets)
 
     logger.debug("validating storage class parameters")
     if request.parameters:
-        if config.PARAMETERS_POOL not in request.parameters:
+        if config.PARAMETERS_POOL in request.parameters:
+            if not request.parameters[config.PARAMETERS_POOL]:
+                raise ValidationException(messages.wrong_pool_passed_message)
```
oriyarde, line 248: wrong, or more specifically, empty?

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -118,21 +123,42 @@ def test_create_snapshot_with_empty_name(self, a_enter):
...
+    def test_create_snapshot_succeeds(self, storage_agent):
+        self._test_create_snapshot_succeeds(storage_agent)
+
+    @patch("controller.controller_server.csi_controller_server.get_agent")
+    def test_create_snapshot_with_pool_parameter_succeeds(self, storage_agent):
+        self.request.parameters = {config.PARAMETERS_POOL: pool}
+        self._test_create_snapshot_succeeds(storage_agent, expected_pool=pool)
+
+    def _test_create_snapshot_with_pools_by_system_parameter(self, storage_agent, system_id, expected_pool):
+        self.request.source_volume_id = "{}:{}:{}".format("A9000", system_id, snapshot_volume_wwn)
```
oriyarde, line 152: I thought the max system type length is 4

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -422,11 +449,18 @@ def test_create_volume_idempotent_no_source_succeeds(self, storage_agent):
     def test_create_volume_with_wrong_secrets(self, a_enter):
         self._test_create_object_with_wrong_secrets(a_enter)
 
+    @patch("controller.controller_server.csi_controller_server.get_agent")
+    def test_create_volume_no_pool(self, storage_agent):
+        self._prepare_create_volume_mocks(storage_agent)
+        self.request.parameters = {"pools_by_system": str({"u1": pool, "u2": "other_pool"})}
```
oriyarde, line 455: is the `pools_by_system` name purposely invalid?

controller/tests/utils.py
```diff
@@ -28,6 +33,29 @@ def get_mock_mediator_response_snapshot(capacity, name, wwn, volume_name, array_
     return snapshot
 
 
+def get_mock_array_connection_info(user="user", password="pass", array_addresses=None, system_id="u1"):
+    if array_addresses is None:
+        array_addresses = ["arr1"]
+    return ArrayConnectionInfo(array_addresses=array_addresses, user=user, password=password, system_id=system_id)
```
oriyarde, line 39: is it really a mock?

controller/controller_server/utils.py
ArbelNathan: yes, but if no topologies, we don't care about the rest..

controller/controller_server/utils.py
ArbelNathan: I think it's correct, there is a map for a system by topologies and that what it getting

controller/tests/utils.py
ArbelNathan: in the sense that its not the real return value from utils

controller/tests/controller_server/csi_controller_server_test.py
ArbelNathan: it's a leftover, I removed all of them

controller/tests/controller_server/csi_controller_server_test.py
ArbelNathan: right, done (thought of XIV)

controller/controller_server/utils.py
ArbelNathan: error

controller/controller_server/messages.py
ArbelNathan: done

controller/controller_server/messages.py
ArbelNathan: done

controller/controller_server/csi_controller_server.py

controller/controller_server/utils.py
ArbelNathan: sure

controller/controller_server/messages.py
ArbelNathan: done

controller/controller_server/utils.py

controller/controller_server/messages.py
ArbelNathan: done

controller/controller_server/utils.py

controller/controller_server/messages.py
ArbelNathan: done

controller/controller_server/utils.py

controller/tests/controller_server/csi_controller_server_test.py
oriyarde: then perhaps this also should be computed (e.g. `max(map(len, ALL_ARRAY_TYPES)))`)

controller/tests/utils.py
oriyarde: then rename `mock` to something like `fake`/`test`

controller/controller_server/utils.py
oriyarde: no, just a more specific message

controller/controller_server/utils.py
oriyarde: does it return a dictionary where the topologies are the keys?

controller/controller_server/utils.py
oriyarde: isn't this true for each of them?

controller/controller_server/messages.py
```diff
@@ -3,7 +3,12 @@
 ObjectIdError_message = "Wrong {0} id format : {1}"
 
 # validation error messages
-invalid_secret_message = "invalid secret was passed"
+invalid_secrets_config_message = "invalid secrets config was passed"
+no_way_to_identify_system_from_secrets_config_message = "no way to identify system from secrets config"
+secrets_missing_connectivity_info_message = "secrets is missing connectivity info"
+secrets_missing_topologies_message = "secrets is missing topologies"
+invalid_system_id_message = "got an invalid system id: {}, valid regex: {}"
```
oriyarde, line 10: consider replacing `valid` with something like `validating` (the regex itself is obviously valid)

controller/controller_server/messages.py
oriyarde: I still see some messages with `was passed`

controller/controller_server/utils.py
oriyarde: which error?

controller/controller_server/utils.py
```diff
@@ -154,7 +274,7 @@ def validate_delete_snapshot_request(request):
         raise ValidationException(messages.name_should_not_be_empty_message)
     logger.debug("validating secrets")
     if request.secrets:
-        validate_secret(request.secrets)
+        validate_secrets(request.secrets)
```
oriyarde, line 277: is this `if` still needed? if not, could you remove it while you're at it?

controller/controller_server/messages.py
```diff
@@ -3,7 +3,12 @@
 ObjectIdError_message = "Wrong {0} id format : {1}"
 
 # validation error messages
-invalid_secret_message = "invalid secret was passed"
+invalid_secrets_config_message = "invalid secrets config was passed"
+no_way_to_identify_system_from_secrets_config_message = "no way to identify system from secrets config"
```
oriyarde, line 7: I wouldn't call it `identify`, perhaps something like `insufficient data to choose a storage system`?

controller/controller_server/utils.py
```diff
@@ -8,43 +10,161 @@
...
+            logger.info("Chosen volume topologies: {}".format(topologies))
+            return topologies
+
+
+def get_system_info_by_topologies(secrets_config, node_topologies):
+    for system_id, system_info in secrets_config.items():
+        system_topologies = system_info.get(config.SECRET_SUPPORTED_TOPOLOGIES_PARAMETER)
+        if _is_topology_match(system_topologies, node_topologies):
+            return system_info, system_id
+    raise ValidationException(messages.no_way_to_identify_system_from_secrets_config_message)
```
oriyarde, line 52: this could be more specific, e.g. `no system match requested topology`

controller/controller_server/messages.py
```diff
@@ -3,7 +3,12 @@
...
+secrets_missing_connectivity_info_message = "secrets is missing connectivity info"
```
oriyarde, line 8: 1. secret*
2. connection info?

controller/controller_server/messages.py
```diff
@@ -3,7 +3,12 @@
 ObjectIdError_message = "Wrong {0} id format : {1}"
 
 # validation error messages
-invalid_secret_message = "invalid secret was passed"
+invalid_secrets_config_message = "invalid secrets config was passed"
```
oriyarde, line 6: secret*?

controller/tests/utils.py
```diff
@@ -1,6 +1,11 @@
+import json
+
 from mock import Mock
 import grpc
 
+from controller.controller_server.controller_types import ArrayConnectionInfo
+from controller.controller_server.test_settings import user as _user, password as _password, array as _array
```
oriyarde, line 7: use a more indicative prefix than `_`, e.g. `test`

controller/controller_server/utils.py
ArbelNathan: no we can be fine without system_id and even if the secrets are valid it doesn't matter if we don't have topologies..

controller/controller_server/utils.py
ArbelNathan: its not returning map, but the system info that returns are mapped by topologies. (and AFAIK map and dictionary are equivalent.)

controller/controller_server/utils.py
ArbelNathan: in [get_system_info_by_topologies ](https://github.com/IBM/ibm-block-csi-driver/blob/7d4d4442ddbc18cfe786657eabd85122356ad71d/controller/controller_server/utils.py#L52)

controller/controller_server/utils.py
ArbelNathan: the message is 'pool parameter must provide value for it'

ArbelNathan: I can change the name of the message

controller/controller_server/utils.py
oriyarde: not sure I follow, could you write an example of the structure here?

controller/controller_server/utils.py
oriyarde: right, so could you change it to something like `pool parameter should not be empty` while you're at it?

controller/tests/utils.py
ArbelNathan: done

controller/controller_server/messages.py
ArbelNathan: done

controller/controller_server/messages.py
ArbelNathan: done

controller/controller_server/utils.py

controller/controller_server/messages.py
ArbelNathan: done

controller/controller_server/utils.py

controller/controller_server/messages.py
ArbelNathan: done

controller/tests/utils.py
ArbelNathan: done

controller/tests/controller_server/csi_controller_server_test.py
ArbelNathan: where?

controller/controller_server/messages.py
ArbelNathan: done

controller/controller_server/utils.py
ArbelNathan: how?

controller/controller_server/utils.py
ArbelNathan: the structure is the secret config.json

controller/controller_server/utils.py

controller/controller_server/utils.py
oriyarde:
```suggestion
            supported_topologies = system_info.get(config.SECRET_SUPPORTED_TOPOLOGIES_PARAMETER)
            _validate_topologies(supported_topologies)
```

controller/tests/controller_server/csi_controller_server_test.py
oriyarde: [here](https://github.com/IBM/ibm-block-csi-driver/pull/324/files#diff-805936c9666f06d2a0664c47d089a0d84d457d1f07cda4d80713d8e5bfb3eefaR15)

controller/controller_server/utils.py
oriyarde: great! why is it public btw?

controller/controller_server/utils.py
oriyarde: I thought that in the config.json the system info is "mapped by" system id, not topologies

controller/controller_server/utils.py
```diff
@@ -8,43 +10,161 @@
...
 
 logger = get_stdout_logger()
 
 
-def get_array_connection_info_from_secret(secrets):
+def _parse_raw_json(raw_json):
+    try:
+        parsed_json = json.loads(raw_json)
+    except json.decoder.JSONDecodeError:
+        raise ValidationException(messages.invalid_json_parameter_message.format(raw_json))
```
oriyarde, line 24: could you include the error message in our exception message here?

controller/controller_server/utils.py
oriyarde: 1. if I rearrange the second clause into `if topologies are valid it doesn't matter if we don't have the secrets` - isn't it also true?
2. 

> we can be fine without system_id

the scope of this code is under `if system_id`, and even if topologies are valid it doesn't matter if the system_id is invalid

controller/controller_server/messages.py
oriyarde: same for `secrets_missing_topologies_message` var name

controller/controller_server/messages.py
oriyarde: 2. same for var name

controller/controller_server/messages.py
ArbelNathan: it is changed in the mesasage

oriyarde: I know, which is good, but I think the var name should be aligned accordingly

controller/controller_server/utils.py
ArbelNathan: sure, done

controller/controller_server/messages.py
ArbelNathan: done

controller/controller_server/messages.py
ArbelNathan: done

controller/tests/controller_server/csi_controller_server_test.py
ArbelNathan: ok, done

controller/controller_server/utils.py
ArbelNathan: right, now it's private

controller/controller_server/utils.py
ArbelNathan: ok, done. it's `for` now

controller/controller_server/utils.py
ArbelNathan: ok, it in the `appearance in the example` order now

controller/controller_server/utils.py
ArbelNathan: tnx, done

controller/controller_server/utils.py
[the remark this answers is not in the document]
ArbelNathan: CSI-3062

controller/tests/controller_server/csi_controller_server_test.py
oriyarde: I think it would be easier if we maintain a `ALL_ARRAY_TYPES` const in the same file the types are defined

controller/controller_server/messages.py
oriyarde: I still see `invalid_secrets_config_message` (plural)

ArbelNathan: wired, I can't.. can you link the occurrences?

controller/tests/controller_server/csi_controller_server_test.py
ArbelNathan: done

controller/controller_server/messages.py
oriyarde: now I can't see it either, must have been on my side. resolving

oriyarde (approved):
nice job!

PR #336, opened 2021-06-16
Feature/csi 2953 volume clone pool gap

ArbelNathan:


controller/array_action/array_mediator_ds8k.py
```diff
@@ -64,10 +64,12 @@ def scsi_id_to_volume_id(scsi_id):
     return scsi_id[-4:]
 
 
-def try_convert_first_arg(converter, args):
+def try_convert_firsts_arg(converter, args, num_of_args):
```
oriyarde, line 67: `try_convert_first_args`, `number_of_args`/`args_amount`

controller/array_action/array_mediator_ds8k.py
```diff
@@ -64,10 +64,12 @@ def scsi_id_to_volume_id(scsi_id):
...
     if args:
-        converted = converter(args[0])
-        return (converted,) + args[1:]
+        converted = []
+        for index in range(num_of_args):
+            converted.append(converter(args[index]))
+        return tuple(converted) + args[num_of_args:]
```
oriyarde, line 72:
```suggestion
        args_to_convert = args[:num_of_args]
        converted = map(converter, args_to_convert)
        return tuple(converted) + args[num_of_args:]
```

controller/array_action/array_mediator_ds8k.py
```diff
@@ -292,19 +297,19 @@ def _extend_volume(self, api_volume, new_size_in_bytes):
             if ERROR_CODE_EXPAND_VOLUME_NOT_ENOUGH_EXTENTS in str(ex.message).upper():
                 raise array_errors.NotEnoughSpaceInPool(api_volume.pool)
 
-    def copy_to_existing_volume_from_source(self, name, source_name, source_capacity_in_bytes,
-                                            minimum_volume_size_in_bytes, pool=None):
+    @convert_scsi_ids_to_array_id(2)
```
oriyarde, line 300: please use a named argument

controller/array_action/array_mediator_ds8k.py
```diff
@@ -292,19 +297,19 @@ def _extend_volume(self, api_volume, new_size_in_bytes):
...
+    def copy_to_existing_volume_from_source(self, volume_id, source_id, source_capacity_in_bytes,
+                                            minimum_volume_size_in_bytes):
         logger.debug(
             "Copy source {0} data to volume {1}. source capacity {2}. Minimal requested volume capacity {3}".format(
-                name, source_name, source_capacity_in_bytes,
+                volume_id, source_id, source_capacity_in_bytes,
```
oriyarde, line 305: I think the original format here mixed up the target and the source

controller/array_action/array_mediator_ds8k.py
```diff
@@ -292,19 +297,19 @@ def _extend_volume(self, api_volume, new_size_in_bytes):
...
                 minimum_volume_size_in_bytes))
-        api_new_volume = self._get_api_volume_by_name(name, pool_id=pool)
-        api_source_object = self._get_api_volume_by_name(source_name, pool_id=pool)
         if minimum_volume_size_in_bytes < source_capacity_in_bytes:
+            api_new_volume = self._get_api_volume_by_id(volume_id)
```
oriyarde, line 308: `new_api_volume`

controller/array_action/array_mediator_ds8k.py
```diff
@@ -79,10 +81,13 @@ def is_snapshot(api_volume):
     return False
 
 
-@decorator
-def convert_scsi_id_to_array_id(mediator_method, self, *args):
-    args = try_convert_first_arg(scsi_id_to_volume_id, args)
-    return mediator_method(self, *args)
+def convert_scsi_ids_to_array_id(num_of_args=1):
```
oriyarde, line 84: `to_array_ids`?

controller/array_action/array_mediator_svc.py
```diff
@@ -407,9 +407,11 @@ def _copy_to_target_volume(self, target_volume_name, source_volume_name):
...
             raise ex
 
-    def copy_to_existing_volume_from_source(self, name, source_name, source_capacity_in_bytes,
-                                            minimum_volume_size_in_bytes, pool=None):
-        self._copy_to_target_volume(name, source_name)
+    def copy_to_existing_volume_from_source(self, volume_id, source_id, source_capacity_in_bytes,
+                                            minimum_volume_size_in_bytes):
+        source_name = self._get_volume_name_by_wwn(source_id)
+        target_volume_name = self._get_volume_name_by_wwn(volume_id)
+        self._copy_to_target_volume(target_volume_name, source_name)
```
oriyarde, line 414: now that we get the source *id* (instead of name) as a parameter, could you please change `_copy_to_target_volume` (and `_create_and_start_fcmap`) to require the source *id* accordingly?

controller/array_action/array_mediator_interface.py
```diff
@@ -54,18 +54,17 @@ def create_volume(self, volume_name, size_in_bytes, space_efficiency, pool):
...
-            source_name                     : name of source to create from
+            volume_id                       : id of the volume to be created in the storage system
+            source_id                       : id of source to create from
             source_capacity_in_bytes        : capacity of source to create from
             minimum_volume_size_in_bytes    : if source capacity is lower than this value volume will
                                             be increased to this value
-            pool                            : pool of the volume and source object to find them more efficiently.
 
         Returns:
             Volume
```
oriyarde, line 70: `Raises` (and the mediators exception handling) should be updated accordingly

controller/array_action/array_mediator_xiv.py
```diff
@@ -233,13 +236,13 @@ def copy_to_existing_volume_from_source(self, name, source_name, source_capacity
...
-            raise array_errors.ObjectNotFoundError(source_name)
+            raise array_errors.ObjectNotFoundError(source_id)
         except (xcli_errors.VolumeBadNameError, xcli_errors.TargetVolumeBadNameError) as ex:
             logger.exception(ex)
-            raise array_errors.ObjectNotFoundError(name)
+            raise array_errors.ObjectNotFoundError(volume_id)
         except xcli_errors.OperationForbiddenForUserCategoryError as ex:
             logger.exception(ex)
-            raise array_errors.PermissionDeniedError("create volume : {0}".format(name))
+            raise array_errors.PermissionDeniedError("create volume : {0}".format(volume_id))
```
oriyarde, line 245: I think the original message here is misleading, since it's not really "create"

controller/array_action/array_mediator_svc.py
ArbelNathan: in this stage its not gonna save any code

controller/array_action/array_mediator_xiv.py
ArbelNathan: done

controller/array_action/array_mediator_interface.py
ArbelNathan: ok, I think I cover the "new" possible errors

controller/array_action/array_mediator_ds8k.py
ArbelNathan: sure

controller/array_action/array_mediator_ds8k.py

controller/array_action/array_mediator_ds8k.py

controller/array_action/array_mediator_ds8k.py
ArbelNathan: sure

controller/array_action/array_mediator_ds8k.py

oriyarde: `first_arg`**s** (plural), please avoid half words like `num`

controller/array_action/array_mediator_xiv.py
```diff
@@ -233,13 +236,13 @@ def copy_to_existing_volume_from_source(self, name, source_name, source_capacity
...
-            raise array_errors.ObjectNotFoundError(source_name)
+            raise array_errors.ObjectNotFoundError(source_id)
         except (xcli_errors.VolumeBadNameError, xcli_errors.TargetVolumeBadNameError) as ex:
             logger.exception(ex)
-            raise array_errors.ObjectNotFoundError(name)
+            raise array_errors.ObjectNotFoundError(volume_id)
         except xcli_errors.OperationForbiddenForUserCategoryError as ex:
             logger.exception(ex)
-            raise array_errors.PermissionDeniedError("create volume : {0}".format(name))
+            raise array_errors.PermissionDeniedError("copy to {0} from source".format(volume_id))
```
oriyarde, line 245: could you add the source while you're at it?

controller/array_action/array_mediator_interface.py
```diff
@@ -54,27 +54,26 @@ def create_volume(self, volume_name, size_in_bytes, space_efficiency, pool):
...
 
         Raises:
             ObjectNotFoundError
             IllegalObjectName
```
oriyarde, line 74: which object name?

ArbelNathan: the [same one](https://github.com/IBM/ibm-block-csi-driver/blob/cf76881f4aa478cd1fc6676f9b70363952cee78a/controller/array_action/array_mediator_svc.py#L230) there was before.. (but now we are getting it in the mediator, the rest of the calls didn't change so the possible errors returned stayed the same)

controller/controller_server/exception_handler.py
```diff
@@ -11,6 +11,7 @@
 status_codes_by_exception = {
     ValidationException: grpc.StatusCode.INVALID_ARGUMENT,
     array_errors.IllegalObjectName: grpc.StatusCode.INVALID_ARGUMENT,
+    array_errors.IllegalObjectID: grpc.StatusCode.INVALID_ARGUMENT,
```
zingero, line 14: How this is related to this PR?

controller/array_action/array_mediator_ds8k.py
zingero: `arg` is a half-word as well

zingero (commented):
almost :)

controller/array_action/array_mediator_ds8k.py
oriyarde: but `args` can be found in official [PEP](https://www.python.org/dev/peps/pep-0448/#specification)s and docs, so I think it's legit

controller/controller_server/exception_handler.py
ArbelNathan: it an error that was added to the possible error that can be returned..

controller/array_action/array_mediator_interface.py
oriyarde: since the server no longer sends a name, I don't think it should except/expect this error, it sounds to me like an internal mediator implementation detail (not relevant to e.g. xiv).

and even if for some reason it gets raised, it should not lead to `INVALID_ARGUMENT`, since at this point we know the request values (specifically the source **id**) are valid

controller/array_action/array_mediator_xiv.py
ArbelNathan: done

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -743,10 +743,9 @@ def test_create_volume_from_snapshot_success(self, storage_agent):
                                                                                                 "a9k")
         response_volume = self.servicer.CreateVolume(self.request, self.context)
         self.assertEqual(self.context.code, grpc.StatusCode.OK)
-        self.mediator.copy_to_existing_volume_from_source.assert_called_once_with(volume_name, snapshot_name,
+        self.mediator.copy_to_existing_volume_from_source.assert_called_once_with('wwn2', snapshot_id,
                                                                                   snapshot_capacity_bytes,
-                                                                                  self.capacity_bytes,
-                                                                                  pool)
+                                                                                  self.capacity_bytes)
```
oriyarde, line 748: please add test coverage for the interface error change, on both sides of the interface (server and mediators)

controller/array_action/array_mediator_ds8k.py
zingero: In the official docs, they refer to `*args` / `**kwargs`, not to a "regular" parameter called `args`

oriyarde: I don't think it matters if it's "regular" or not - it always means `arguments`, it's a common enough name in the language, so it should be easy for the reader to read

controller/array_action/array_mediator_interface.py
ArbelNathan: then its a bug, cause nothing actually changed

oriyarde: I agree that according to my suggested interface, some mediators will be "supposed" to internally except `IllegalObjectName` accordingly, but I'm OK if they will "blindly" rely on the name to be valid, since they now get it from **the storage**.

controller/controller_server/exception_handler.py
zingero: Is it a common exception?

ArbelNathan: just like IllegalObjectName

controller/array_action/array_mediator_ds8k.py
ArbelNathan: I changed to args.. plz resolved it between yourself

controller/array_action/array_mediator_interface.py
ArbelNathan: ok, done

controller/tests/controller_server/csi_controller_server_test.py
ArbelNathan: done

controller/array_action/array_mediator_ds8k.py
```diff
@@ -79,10 +80,13 @@ def is_snapshot(api_volume):
     return False
 
 
-@decorator
-def convert_scsi_id_to_array_id(mediator_method, self, *args):
-    args = try_convert_first_arg(scsi_id_to_volume_id, args)
-    return mediator_method(self, *args)
+def convert_scsi_ids_to_array_ids(args_amount=1):
+    @decorator
+    def convert_scsi_id_to_array_id(mediator_method, self, *args):
```
oriyarde, line 85: consider renaming to something like `try_convert_first_args`

controller/tests/array_action/xiv/array_mediator_xiv_tests.py
```diff
@@ -116,58 +116,67 @@ def test_create_volume__generate_volume_response_raise_exception(self, response)
...
-        self.mediator.copy_to_existing_volume_from_source(volume_name, src_snapshot_name,
+        target_volume = self._get_cli_volume(name=volume_name)
+        source_volume = self._get_cli_volume(name=src_snapshot_name)
+        self.mediator.client.cmd.vol_list.side_effect = [Mock(as_single_element=target_volume),
+                                                         Mock(as_single_element=source_volume)]
+        self.mediator.copy_to_existing_volume_from_source(volume_id, source_id,
                                                           src_snapshot_capacity_in_bytes, min_volume_size_in_bytes)
-        volume_size_in_blocks = 1
+        calls = [call(wwn=volume_id), call(wwn=source_id)]
+        self.mediator.client.cmd.vol_list.assert_has_calls(calls, any_order=False)
```
oriyarde, line 135: do we really want to enforce the order?

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -768,7 +767,7 @@ def test_create_volume_from_source_source_snapshot_invalid(self):
 
     @patch("controller.controller_server.csi_controller_server.get_agent")
     def test_create_volume_from_source_illegal_object_name(self, storage_agent):
-        array_exception = array_errors.IllegalObjectName("")
+        array_exception = array_errors.IllegalObjectID("")
```
oriyarde, line 770: please rename the test accordingly

controller/array_action/array_mediator_ds8k.py
zingero: I think naming a parameter `args` without marking it with `*` is confusing.

oriyarde: even with `*`, it just means `arguments`, like `argparse`. I don't see a reason to confuse it with the unrelated `*` operator.

controller/tests/array_action/xiv/array_mediator_xiv_tests.py
ArbelNathan: yes, if it's in any other order so it means that something went wrong. that is the order of the callings in the [method](https://github.com/IBM/ibm-block-csi-driver/blob/57287fe734bc3c22f09e59cc890b34d306b41861/controller/array_action/array_mediator_xiv.py#L223-L224)

oriyarde: why is it wrong for the method to call them in reversed order?

ArbelNathan: how can you be sure that the right value returns to each parameter? (e.g. the call for the `source_name ` value should be called with `source_id`)

controller/tests/controller_server/csi_controller_server_test.py

controller/array_action/array_mediator_ds8k.py

controller/tests/array_action/xiv/array_mediator_xiv_tests.py
oriyarde: got it

oriyarde (approved):
LGTM

controller/array_action/array_mediator_ds8k.py
zingero: In Python `args` is kind of a saved keyword.
It's similar to have a function like:
`def foo(a, b, self): pass`
Usually, we don't use special names like that.

oriyarde: `args` is not special as `self`.
`self` is important because otherwise the reader needs out-of-function context to know if the first arg is the class instance.
but the `*` (or lack of it) is visible in the same place, without extra context.
in addition, the general word "arguments" (or "args", for short) is more prevalent than the word "self", in coding.

zingero: > the general word "arguments" (or "args", for short) is more prevalent than the word "self", in coding.
It's not in Python.
Not sure why you're against my suggestion. Do the extra five characters bother you?

oriyarde: yes, and I think it's always more natural in python to use `args` instead of `arguments`

ArbelNathan: as I see it, even with the fact that it's a `special name`, the `args` that are used here are the `args` of the special meaning. (unlike the `self` that clearly is not the context of the function)

zingero: > the args that are used here are the args of the special meaning

I disagree. As long as there is no `*`, it's a regular parameter with confusing type and name.
Since both of you prefer to use `args`, I retreat.

zingero (approved):

controller/array_action/array_mediator_ds8k.py
oriyarde: why confusing type? it's a tuple, just like with `*`

zingero: How do you its type? You need to search for all the places we call this function first. With `*` you don't need to do that.

oriyarde: with this naming convention, you don't need to search for all the places, just like `self`

PR #341, opened 2021-06-23
Bug/csi 3102 gen3 unsupported parameters

ArbelNathan:


controller/array_action/array_mediator_xiv.py
```diff
@@ -114,6 +114,13 @@ def is_active(self):
     def _convert_size_blocks_to_bytes(self, size_in_blocks):
         return int(size_in_blocks) * self.BLOCK_SIZE_IN_BYTES
 
+    def _get_copy_source_object_wwn(self, cli_object):
```
oriyarde, line 117: consider removing the `copy_`

controller/array_action/array_mediator_xiv.py
```diff
@@ -114,6 +114,13 @@ def is_active(self):
...
+        if hasattr(cli_object, "copy_master_wwn"):
```
zingero, line 118: When it doesn't have this attribute?

controller/array_action/array_mediator_xiv.py
```diff
@@ -131,18 +138,21 @@ def _generate_snapshot_response(self, cli_snapshot):
...
                         cli_snapshot.name,
                         self.endpoint,
-                        volume_id=cli_snapshot.copy_master_wwn,
+                        volume_id=self._get_copy_source_object_wwn(cli_snapshot),
                         is_ready=True,
                         array_type=self.array_type)
 
-    def get_volume(self, volume_name, pool=None):
-        logger.debug("Get volume : {}".format(volume_name))
+    def _get_cli_object_by_name(self, volume_name):
```
zingero, line 145: Why is it called CLI? Is it related to Command Line Interface?

controller/array_action/array_mediator_xiv.py
oriyarde: on gen3...

controller/array_action/array_mediator_xiv.py
```diff
@@ -131,18 +138,21 @@ def _generate_snapshot_response(self, cli_snapshot):
...
         try:
-            cli_volume = self.client.cmd.vol_list(vol=volume_name).as_single_element
+            return self.client.cmd.vol_list(vol=volume_name).as_single_element
         except xcli_errors.IllegalNameForObjectError as ex:
             logger.exception(ex)
             raise array_errors.IllegalObjectName(ex.status)
 
+    def get_volume(self, volume_name, pool=None):
+        logger.debug("Get volume : {}".format(volume_name))
+        cli_volume = self._get_cli_object_by_name(volume_name)
```
zingero, line 154: Does the function return a CLI object or CLI volume?

controller/array_action/array_mediator_xiv.py
oriyarde: yes, it's the data you see when using the xcli

controller/array_action/array_mediator_xiv.py
oriyarde: it's either a cli volume or a [cli snapshot](https://github.com/IBM/ibm-block-csi-driver/blob/ceb5060d18fa53856290428ec033e0d5e8a7a519/controller/array_action/array_mediator_xiv.py#L313), and the current common name for volume and snapshot, across this project, is "object"

controller/array_action/array_mediator_xiv.py
zingero: The code isn't clear enough. I suggest clarifying it by one of:
1. Create a gen3 mediator that inherits from this class and override the relevant function
2. Divide the function into two functions (one for gen3 and one for other gens)

WDYT?

oriyarde: a bit overkill IMO and the code looks pretty clear to me.
we don't even officially support gen3, it's mainly for our tests

zingero: If we don't officially support gen3, then it's even more important to explain why did we do so in the code.
Currently, nothing in this function implies gen3.

ArbelNathan: would you like me to copy [these ](https://github.com/IBM/ibm-block-csi-driver/blob/793cb5b37e1f1e24900b5fd84a4f3c4572bd9a51/controller/array_action/array_mediator_xiv.py#L118)messages to this method too?

oriyarde: I don't think we should maintain comments.
if we are certain that this scenario is relevant to (and only to) gen3, then we can convey this info by code, e.g. wrap the `hasattr` in a `is_gen3` method

ArbelNathan: or `is_not_gen3`..

oriyarde: not sure why `not` is better

ArbelNathan: cause if its `hasattr` is true so its not gen3..

zingero: The function can `return not hasattr...`

oriyarde: then it might be better for the `not` to be with `hasattr` - @zingero?

ArbelNathan: so what decided?

oriyarde: `return not hasattr` sounds good to me

controller/array_action/array_mediator_xiv.py
ArbelNathan: done

controller/array_action/array_mediator_xiv.py
ArbelNathan: done

zingero (dismissed):

oriyarde (dismissed):
LGTM

oriyarde (approved):
LGTM

controller/array_action/array_mediator_xiv.py
```diff
@@ -126,15 +126,13 @@ def _get_source_object_wwn(self, cli_object):
         return cli_object.copy_master_wwn
 
     def _generate_volume_response(self, cli_volume):
-        # vol_copy_type and copy_master_wwn were added in a9k. In xiv they didn't exist
-        is_copy = hasattr(cli_volume, "vol_copy_type") and cli_volume.vol_copy_type == "Copy"
-        copy_src_object_wwn = cli_volume.copy_master_wwn if is_copy else None
+        source_object_wwn = cli_volume.copy_master_wwn if not self._is_gen3(cli_volume) else None
```
zingero, line 129: What's your opinion regarding double negative?

My suggestion:
```suggestion
        source_object_wwn = None if self._is_gen3(cli_volume) else cli_volume.copy_master_wwn
```

oriyarde: I don't see a **double** negative here, and I think gen3+None should remain on the right, because they are (literally) marginal

zingero: not + else equals double.

@ArbelNathan WDYT?

ArbelNathan: I agree with ori

zingero: As you wish

zingero (approved):

PR #343, opened 2021-06-28
Improvment/csi 2625 allow node to work with empty fc path

matancarmeli7:


node/pkg/driver/node_utils.go
```diff
@@ -239,6 +243,25 @@ func (n NodeUtils) ParseFCPorts() ([]string, error) {
...
 }
 
+func (n NodeUtils) IsFCExists() bool {
+	if n.IsPathExists(FCPath) {
+		return n.IsDirHasContent(FCPath)
+	}
+	return false
+}
+
+func (n NodeUtils) IsDirHasContent(path string) bool {
```
oriyarde, line 253: this should be a private function (start with lower case letter).
also consider renaming to something like `isNotEmptyDir`, or even `isEmptyDir` with `!` before the call - @zingero what do you think?

node/pkg/driver/node_utils.go
```diff
@@ -239,6 +243,25 @@ func (n NodeUtils) ParseFCPorts() ([]string, error) {
...
+	f, _ := os.Open(path)
+	defer f.Close()
+
+	_, err := f.Readdir(1)
```
zingero, line 257: What `1` mean?
It looks like a [magic number](https://en.wikipedia.org/wiki/Magic_number_%28programming%29).
Can we create a constant with a descriptive name instead?

node/pkg/driver/node_utils.go
```diff
@@ -239,6 +243,25 @@ func (n NodeUtils) ParseFCPorts() ([]string, error) {
...
+
+	if err == io.EOF {
+		return false
+	}
+	return true
+}
```
zingero, line 263: We can simplify this by:
```suggestion
	return err != io.EOF
```

node/pkg/driver/node_utils.go
```diff
@@ -239,6 +243,25 @@ func (n NodeUtils) ParseFCPorts() ([]string, error) {
 	return fcPorts, nil
 }
 
+func (n NodeUtils) IsFCExists() bool {
+	if n.IsPathExists(FCPath) {
+		return n.IsDirHasContent(FCPath)
+	}
+	return false
```
zingero, line 250: We can simplify this by:
```suggestion
	return n.IsPathExists(FCPath) && return n.IsDirHasContent(FCPath)
```

node/pkg/driver/node_utils.go
zingero: `isEmptyDir` is preferred. Personally, I don't like adding "not" as part of the function name since it can create double negatives.

node/pkg/driver/node_utils.go
matancarmeli7: this means that it will try to read just one file in the directory.
Do you still think I need to add it as a const?

oriyarde: looks like a [well documented](https://golang.org/pkg/os/#File.Readdir) number to me

node/pkg/driver/node_utils.go
matancarmeli7: I used `IsDirHasContent` because I didn't want to use double negatives with `isNotEmptyDir` and with `return ! n.isEmptyDir(FCPath)`, so if you think that this `isEmptyDir` with the `!` is better so I will change it.
Anyway I will change the first character to a lower case letter.

oriyarde: I like `!n.isEmptyDir(FCPath)` a bit better than `n.IsDirHasContent(FCPath)`, but whatever you guys prefer

node/pkg/driver/node_utils.go
oriyarde: we could use a const like `MIN_FILES_IN_NON_EMPTY_DIR`, if you think it's better

node/pkg/driver/node_utils.go
zingero: No other errors are relevant here?

> If n > 0, Readdir returns at most n FileInfo structures. In this case, if Readdir returns an empty slice, it will return a non-nil error explaining why. At the end of a directory, the error is io.EOF.

node/pkg/driver/node_utils.go
zingero: @oriyarde there are lots of information online. We would like the code to be self-explanatory though.
Thus, I would avoid using magic numbers.

node/pkg/driver/node_utils.go
```diff
@@ -52,6 +54,7 @@ const (
 	resizeFsTimeoutMilliseconds = 30 * 1000
 	TimeOutMultipathdCmd        = 10 * 1000
 	multipathdCmd               = "multipathd"
+	MIN_FILES_IN_NON_EMPTY_DIR  = 1
```
oriyarde, line 57: I now see the const case convention is different, sorry for [misleading](https://github.com/IBM/ibm-block-csi-driver/pull/343#discussion_r659886887).
please align this new const according to existing convention

node/pkg/driver/node_utils.go
```diff
@@ -239,6 +243,27 @@ func (n NodeUtils) ParseFCPorts() ([]string, error) {
...
+func (n NodeUtils) isEmptyDir(path string) bool {
+	f, _ := os.Open(path)
+	defer f.Close()
+
+	_, err := f.Readdir(minFilesInNonEmptyDir)
+
+	if err != nil {
+		if err != io.EOF {
+			logger.Warningf("Check is directory %s empty returned error %s", path, err.Error())
+			return true
```
oriyarde, line 259: no need for this `return`

matancarmeli7: done

node/pkg/driver/node_utils.go
```diff
@@ -239,6 +243,26 @@ func (n NodeUtils) ParseFCPorts() ([]string, error) {
...
+
+func (n NodeUtils) isEmptyDir(path string) bool {
+	f, _ := os.Open(path)
+	defer f.Close()
+
+	_, err := f.Readdir(minFilesInNonEmptyDir)
+
+	if err != nil {
+		if err != io.EOF {
+			logger.Warningf("Check is directory %s empty returned error %s", path, err.Error())
```
zingero, line 258: Typo, but anyway, I would change the message to:
```suggestion
			logger.Warningf("Failed to read content of directory %s. Error %s", path, err.Error())
```

matancarmeli7: I wanted to do like [here](https://github.com/IBM/ibm-block-csi-driver/blob/259b3d40a7d44c476d7d232631c5646c6aebaf76/node/pkg/driver/node_utils.go#L246) but I agree that what you wrote is better, do you still want me to change it? or keep it like this that it will be similar to the other `logger.Warningf`

oriyarde (approved):
LGTM

node/pkg/driver/node_utils.go
zingero: We'll change it in a different PR. I created a new ticket for that: CSI-3149

zingero (approved):

PR #350, opened 2021-07-19
Feature/csi 3175 enforce system id

ArbelNathan:


controller/controller_server/csi_controller_server.py
```diff
@@ -90,6 +91,11 @@ def CreateVolume(self, request, context):
                     required_bytes = min_size
                     logger.debug("requested size is 0 so the default size will be used : {0} ".format(
                         required_bytes))
+
+                if system_id and system_id != array_mediator.identifier:
+                    raise ValidationException(
+                        controller_messages.system_id_mismatch_message.format(system_id, array_mediator.identifier))
```
oriyarde, line 97: shouldn't it be the first thing to check in the `with`?

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -404,22 +403,36 @@ def _test_create_volume_succeeds(self, storage_agent, expected_pool=pool):
...
+
+    @patch("controller.controller_server.csi_controller_server.get_agent")
+    def test_create_volume_with_topologies_succeeds(self, storage_agent):
+        self._prepare_create_volume_with_topologies()
         self._test_create_volume_succeeds(storage_agent, expected_pool="other_pool")
 
+    @patch("controller.controller_server.csi_controller_server.get_agent")
+    def test_create_volume_with_topologies_system_id_mismatch(self, storage_agent):
+        storage_agent.return_value = self.storage_agent
+        self._prepare_create_volume_with_topologies("wrong_id")
```
oriyarde, line 429: `"wrong_system_id"`

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -404,22 +403,36 @@ def _test_create_volume_succeeds(self, storage_agent, expected_pool=pool):
...
-        self.request.secrets = utils.get_fake_secret_config(system_id="u2", supported_topologies=[
+    def _prepare_create_volume_with_topologies(self, system_id="id123"):
+        self.request.secrets = utils.get_fake_secret_config(system_id=system_id, supported_topologies=[
             {"topology.kubernetes.io/test": "topology_value"}])
         self.request.accessibility_requirements.preferred = [
             ProtoBufMock(segments={"topology.kubernetes.io/test": "topology_value",
                                    "topology.kubernetes.io/test2": "topology_value2"})]
         self.request.parameters = {config.PARAMETERS_BY_SYSTEM: json.dumps(
-            {"u1": {config.PARAMETERS_POOL: pool}, "u2": {config.PARAMETERS_POOL: "other_pool"}})}
+            {"other_id": {config.PARAMETERS_POOL: pool}, system_id: {config.PARAMETERS_POOL: "other_pool"}})}
```
oriyarde, line 419: `"other_system_id"`

oriyarde: btw why is the other pool in the regular system and vice versa?

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -375,7 +374,7 @@ def setUp(self):
 
         self.mediator.maximal_volume_size_in_bytes = 10
         self.mediator.minimal_volume_size_in_bytes = 2
-
+        self.mediator._identifier = "id123"
```
oriyarde, line 377: could we use a more descriptive value? e.g. `"system_id_stub"`

controller/controller_server/csi_controller_server.py
ArbelNathan: is it a stronger problem than the size error (the part before)?

oriyarde: good question.
in a highly unlikely scenario, if a user has different storage types (xiv/svc/ds8k) and got the management address wrong, the size limitations may not be relevant

zingero:
I agree with Ori's comments

controller/controller_server/csi_controller_server.py
ArbelNathan: fair enough, done

controller/tests/controller_server/csi_controller_server_test.py

controller/tests/controller_server/csi_controller_server_test.py

controller/tests/controller_server/csi_controller_server_test.py

controller/controller_server/csi_controller_server.py
```diff
@@ -73,8 +73,14 @@ def CreateVolume(self, request, context):
             space_efficiency = volume_parameters.space_efficiency
             # TODO : pass multiple array addresses
             array_type = detect_array_type(array_connection_info.array_addresses)
+            system_id = array_connection_info.system_id
             with get_agent(array_connection_info, array_type).get_mediator() as array_mediator:
                 logger.debug(array_mediator)
+
+                if system_id and system_id != array_mediator.identifier:
+                    raise ValidationException(
+                        controller_messages.system_id_mismatch_message.format(system_id, array_mediator.identifier))
```
oriyarde, line 82: don't we need this in almost all csi methods (when the system_id is from the volume id)?

ArbelNathan: all other methods getting the `system_id` from the `volume_id`, so I think it's enough to validate only where it initiated.

oriyarde: what if the secret changes between creation and other operations?

oriyarde: or a user didn't get the volume id correctly when importing a volume?

ArbelNathan: and why should we check where we don't care about it? where should we check it? `deletevolume`? `createsnapshot`? can't see why

oriyarde: yes, we consider that as an option.
we care about it because this value has a meaning - if this value refers to an actual different storage system, it is no longer correct for us to let the user operate in the wrong system.

zingero (approved):

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -85,6 +89,16 @@ def _test_create_object_with_get_array_type_exception(self, storage_agent):
         msg = array_errors.FailedToFindStorageSystemType("endpoint").message
         self.assertIn(msg, context.details)
 
+    def _test_request_with_by_wrong_system_id_parameter(self, storage_agent):
```
zingero, line 92: Why do we need both "with" and "by" in the function name?
I suggest removing the "by".

controller/controller_server/csi_controller_server.py
oriyarde: unless we want to support migration, even beyond FRU changes

PR #352, opened 2021-07-25
CSI-3170 detecting remote copy relationships

zingero:


controller/tests/array_action/svc/array_mediator_svc_test.py
```diff
@@ -197,6 +199,16 @@ def test_delete_volume_still_copy_fcmaps_not_removed(self):
         with self.assertRaises(array_errors.ObjectIsStillInUseError):
             self.svc.delete_volume("volume")
 
+    def test_delete_volume_found_hyper_swap(self):
```
oriyarde, line 202: I think hyperswap is always one word

oriyarde, line 202: consider replacing `found` with something like `in`

controller/tests/array_action/svc/array_mediator_svc_test.py
```diff
@@ -197,6 +199,16 @@ def test_delete_volume_still_copy_fcmaps_not_removed(self):
...
+        fcmaps_as_target = Mock(as_list=[])
```
oriyarde, line 203: for a hyperswap volume, can there 0 fcmaps where said volume is the target?
if so, is said volume always created with just 1 "rc_controlled" fcmap, where said volume is the source?

controller/tests/array_action/svc/array_mediator_svc_test.py
```diff
@@ -917,6 +939,16 @@ def test_expand_volume_raise_object_in_use(self):
...
 
+    def test_expand_volume_found_hyper_swap(self):
+        self._prepare_mocks_for_expand_volume()
+        fcmaps = self.fcmaps_as_source
+        fcmaps[0].rc_controlled = 'yes'
+        self.svc.client.svcinfo.lsfcmap.side_effect = [Mock(as_list=[]), Mock(as_list=fcmaps)]
+        self.svc.expand_volume('vol_id', 1024)
+
+        self.svc.client.svctask.expandvdisksize.assert_called_once_with(vdisk_id='test_volume', unit='b', size=512)
+        self.svc.client.svctask.rmfcmap.assert_not_called()
```
oriyarde, line 950: is this really necessary?

controller/array_action/array_mediator_svc.py
```diff
@@ -517,14 +517,23 @@ def _stop_and_delete_fcmap(self, fcmap_id):
...
-                             if fcmap.status != FCMAP_STATUS_DONE or fcmap.copy_rate == "0"]
-        if unfinished_fcmaps:
-            raise array_errors.ObjectIsStillInUseError(id_or_name=object_name,
-                                                       used_by=unfinished_fcmaps)
+        fcmaps_to_delete = []
+        fcmaps_in_use = []
+
         for fcmap in fcmaps:
+            if not self._is_in_remote_copy_relationship(fcmap):
+                if fcmap.status != FCMAP_STATUS_DONE or fcmap.copy_rate == "0":
```
oriyarde, line 525: should we wrap the `fcmaps_in_use` checks with a function as well?

controller/tests/array_action/svc/array_mediator_svc_test.py
zingero: Done

controller/array_action/array_mediator_svc.py
zingero: I don't mind wrapping this code, but I think we should have something like:
```suggestion
                if self._is_fcmap_in_use(fcmap):
                    ...
            def _is_fcmap_in_use(self, fcmap):
                 return self._is_fcmap_active(fcmap) or self._is_fcmap_relate_to_snapshot(fcmap)
```
WDYT?

controller/tests/array_action/svc/array_mediator_svc_test.py
zingero: What do you mean by "0 fcmaps where said volume is the target"?
No fcmap means no source/target

controller/tests/array_action/svc/array_mediator_svc_test.py
zingero: Done

controller/tests/array_action/svc/array_mediator_svc_test.py
ArbelNathan: actually, `fcmaps_as_target` cannot be empty and hyperswap we always have a "map" of fcmaps that point to both directions.
therefore [fcmap_as_target](https://github.com/IBM/ibm-block-csi-driver/blob/0e33b1f43d99d50ade5b5656c0f63fc80acbd2ea/controller/array_action/array_mediator_svc.py#L539) will be a non-empty list and we will [try to delete it](https://github.com/IBM/ibm-block-csi-driver/blob/0e33b1f43d99d50ade5b5656c0f63fc80acbd2ea/controller/array_action/array_mediator_svc.py#L546) and get an [exception due to fcmap that is part of remote cope](https://www.ibm.com/docs/en/flashsystem-5x00/8.1.x?topic=messages-cmmvc7154e) (i.e. `rc_controlled`). we need to add some sort of "safe delete" for `fcmap_as_target` too.

controller/array_action/array_mediator_svc.py
oriyarde: I'm not sure `relate_to_snapshot` is accurate enough, since volume from snapshot can be active, but would get False if theoretically would be passed to `_is_fcmap_relate_to_snapshot`

oriyarde: and I assume you meant `related`

controller/tests/array_action/svc/array_mediator_svc_test.py
oriyarde: I mean "if you count the number of fcmaps that their target is said volume, you will get 0".
members of a set can have properties, even if the set is empty (i.e. vacuously)

controller/array_action/array_mediator_svc.py
zingero: yes, it's a typo.
What's your suggestion for wrapping the code?

controller/tests/array_action/svc/array_mediator_svc_test.py
zingero: no, it's impossible since each HS volume has at least one fcmap as a source and at least one fcmap as a target.

controller/array_action/array_mediator_svc.py
oriyarde: maybe `_is_fcmap_relate_to_snapshot` -> `_is_snapshot_fcmap`?

zingero: I'm not sure we do things like this (non-related code improvements) in PRs.
Do you want to open a ticket for that?

oriyarde: no, I suggested this because you wrote that you [don't mind wrapping this code](https://github.com/IBM/ibm-block-csi-driver/pull/352#discussion_r685223423). if you think it shouldn't be part of this PR, then I will resolve

zingero: I wrote that because I thought we want to improve the places we touch anyway.
Then I realized [we don't](https://github.com/IBM/ibm-block-csi-driver/pull/355#discussion_r685778214).

controller/tests/array_action/svc/array_mediator_svc_test.py
```diff
@@ -32,13 +32,22 @@ def setUp(self):
              'target_vdisk_name': 'target_name',
              'id': 'test_fc_id',
              'status': FCMAP_STATUS_DONE,
-             'copy_rate': "non_zero_value"})]
+             'copy_rate': 'non_zero_value',
+             'rc_controlled': 'no'})]
+        self.fcmaps_as_target = [Munch(
```
ArbelNathan, line 37: can you move all the fcmaps creations to a method so all the fcmap objects will be aligned and there will be less code duplication?

zingero: https://github.com/IBM/ibm-block-csi-driver/pull/352#discussion_r689307391

controller/array_action/array_mediator_svc.py
oriyarde: there is no "we" in this link - if both sides agree to include a change, it's allowed.
some suggestions are just not merge blockers, that's all

controller/tests/array_action/svc/array_mediator_svc_test.py
oriyarde: since this change adds duplications, IMO this suggestion is relevant to this PR

controller/tests/array_action/svc/array_mediator_svc_test.py
oriyarde: @ArbelNathan I'm not sure I understood your "safe delete" suggestion - should we open a new conversation for it, or do you want to continue here?

controller/array_action/array_mediator_svc.py
zingero: @ArbelNathan likes to keep the PR without unrelated code improvements. That's the reason I feel it's a team decision instead of an individual one.

oriyarde: then in my sentence replace `both sides` with `all sides`

oriyarde: @ArbelNathan would it be ok to include the suggested wrap in this PR? if not, I will resolve (even if most of us are ok with including it)

controller/tests/array_action/svc/array_mediator_svc_test.py
ArbelNathan: safe delete method that will prevent any rc_controlled fcmap to be deleted (or tried to be deleted..)

controller/array_action/array_mediator_svc.py
ArbelNathan: I believe that any improvement that related to the places that the diff is touching is worth doing. but as @zingero [linked](https://github.com/IBM/ibm-block-csi-driver/pull/352#discussion_r689361028), it seems that the implementation of that notion is very selective..

controller/tests/array_action/svc/array_mediator_svc_test.py
zingero: Done

ArbelNathan (dismissed):
LGTM

controller/tests/array_action/svc/array_mediator_svc_test.py
zingero: Done

ArbelNathan (dismissed):
LGTM

controller/tests/array_action/svc/array_mediator_svc_test.py
oriyarde: so is that what we have now? can I resolve this?

ArbelNathan: yes, its looks ok

controller/array_action/array_mediator_svc.py
oriyarde: great, so it seems in this case we are all ok with the wrapping suggestion

zingero: Actually, that's not the case.

oriyarde: I'm not asking if it's relevant, I'm asking if **in this case** we are ok with including an "irrelevant" change.
not all cases are the same

controller/array_action/array_mediator_svc.py
```diff
@@ -512,19 +512,29 @@ def _stop_fcmap(self, fcmap_id):
             if not is_warning_message(ex.my_message):
                 logger.warning("Failed to stop fcmap '{0}': {1}".format(fcmap_id, ex))
 
-    def _stop_and_delete_fcmap(self, fcmap_id):
-        self._stop_fcmap(fcmap_id)
-        self._delete_fcmap(fcmap_id, force=True)
+    def _stop_and_delete_fcmap(self, fcmap):
+        if not self._is_in_remote_copy_relationship(fcmap):
```
oriyarde, line 516: `_stop_and_delete_fcmap` -> `_safe_stop_and_delete_fcmap`?

controller/array_action/array_mediator_svc.py
zingero: I'm not ok with it

controller/array_action/array_mediator_svc.py
zingero: Done

controller/array_action/array_mediator_svc.py
oriyarde: I got a [different impression](https://github.com/IBM/ibm-block-csi-driver/pull/352#discussion_r689352298), but ok

controller/array_action/array_mediator_svc.py
```diff
@@ -512,19 +512,29 @@ def _stop_fcmap(self, fcmap_id):
...
+            raise array_errors.ObjectIsStillInUseError(id_or_name=object_name, used_by=fcmaps_in_use)
+        for fcmap in fcmaps_to_delete:
             self._delete_fcmap(fcmap.id, force=False)
 
+    def _is_in_remote_copy_relationship(self, fcmap):
+        return fcmap.rc_controlled == YES
+
     def _delete_object(self, cli_object, is_snapshot=False):
         object_name = cli_object.name
         fcmap_as_target = self._get_fcmap_as_target_if_exists(object_name)
```
oriyarde, line 540: could there be more than 1? now that we support both volume from snapshot and hyperswap?

zingero: At the moment we use the `_get_fcmap_as_target_if_exists` function in five scenarios:
1. Create volume
2. Create snapshot
3. Delete volume
4. Delete snapshot
5. Expand volume

Can't a volume have more than one snapshot? Why now it's different than the other cases?
Also, when do we expect to have more than one fcmap as target?

oriyarde: I think this will answer your 3 questions:
before hyperswap, a volume (that we created) could not have been a target in more than 1 fcmap, because that would cause data corruption/loss.
e.g. for
```
v1 --fcmap1-> s1 --fcmap2-->v2
 |
 `--fcmap3--> s2
 ```
 the fcmaps would have:
 ```
fcmap1.source = v1
fcmap1.target = s1
 
fcmap2.source = s1
fcmap2.target = v2
 
fcmap3.source = v1
fcmap3.target = s2
 ```
notice that no volume/snapshot has more than 1 fcmap "as input".


but now with hyperswap, I'm not sure if that is still the case.
@ArbelNathan ?

Comment (zingero) [reply controller/array_action/array_mediator_svc.py:540]: Thanks for the explanation.
In the [code](https://github.com/IBM/ibm-block-csi-driver/blob/a5bba0bf27b34a381f9ea1a538f1407ba6df0370/controller/array_action/array_mediator_svc.py#L244), I see that we can receive or expect to receive more than one volume as target since we use `!= 1` instead of `== 0`

Comment (oriyarde) [reply controller/array_action/array_mediator_svc.py:540]: because we didn't want to continue regularly in such an unexpected scenario (it was never supposed to happen)

Comment (zingero) [reply controller/array_action/array_mediator_svc.py:540]: I see. It confused me to think it's a legitimate scenario.

Comment (ArbelNathan) [reply controller/array_action/array_mediator_svc.py:540]: hyperswap does not support being a target of a fcmap, so besides its internal fcmaps, it won't have additional fcmaps as a target.

File: controller/tests/array_action/svc/array_mediator_svc_test.py [hunk 8]
```diff
@@ -469,6 +483,14 @@ def test_delete_snapshot_call_rmfcmap(self):
 
         self.svc.client.svctask.rmfcmap.assert_called_once_with(object_id="test_fc_id", force=True)
 
+    def test_delete_snapshot_in_hyperswap(self):
```
oriyarde, line 486: it's not completely clear from the tests name which feature should or should not work with hyperswap.
if the tests purpose is only to ensure that object (volume/snapshot) deletion doesn't delete hyperswap fcmaps, then it should be reflected in the tests name, e.g. `test_delete_snapshot_does_not_delete_hyperswap_fcmaps`

zingero: Done

oriyarde: thanks.
specifically for expand, since it also does `expandvdisksize.assert_called_once_with`, the previous name was ok

zingero: Done

ArbelNathan (approved):
LGTM

oriyarde (approved):
LGTM

PR #353, opened 2021-08-03
feature CSI-2939 topology example files

ArbelNathan:


deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: snapshot.storage.k8s.io/v1beta1
+kind: VolumeSnapshotClass
+metadata:
+  name: demo-snapshotclass
+driver: block.csi.ibm.com
+deletionPolicy: Delete
+parameters:
+  # pool, SpaceEfficiency and snapshot_name_prefix can be provided in by_system_id parameter for each system or by parameter name as default.
```
oriyarde, line 8: `pool, SpaceEfficiency and snapshot_name_prefix` could we just refer to them as something like "non-secret parameters"?

deploy/kubernetes/examples/demo-secret-configuration-file.json
```diff
@@ -0,0 +1,24 @@
+{
+  "demo-system_id1": {
+    "username": "demo-username",
+    "password": "demo-password",
+    "management_address": "demo-management-address",
+    "supported_topologies": [
+      {
+        "topology.block.csi.ibm.com/zone": "demo-zone",
+        "topology.block.csi.ibm.com/dev": "demo-dev"
```
oriyarde, line 9: I think the `region` is more common than `dev`

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
zingero: I prefer having the explicit way since we have more non-secret parameters to which the comment isn't related. @ArbelNathan WDYT?

deploy/kubernetes/examples/demo-secret-configuration-file.json
```diff
@@ -0,0 +1,24 @@
+{
+  "demo-system_id1": {
```
oriyarde, line 2: I know `by_system_id` has underscores (according to our controller convention), but I think our examples names convention should remain with hyphens, i.e. `demo-system-id1`

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
ArbelNathan: @rivkap what do you think?

deploy/kubernetes/examples/demo-secret-configuration-file.json
zingero: I don't mind having `_`/`-`, but I suggest not mixing them together in one string. In this case, `demo_system_id1` and `demo-system-id1` are ok IMO

deploy/kubernetes/examples/demo-secret-configuration-file.json
ArbelNathan: right, that why I wrote dev (which is common in other docs as well), to emphasize that we support more than the common cases.

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
oriyarde: @zingero which non-secret parameters are not relevant to the comment?

deploy/kubernetes/examples/demo-secret-configuration-file.json
ArbelNathan: ok, done

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
rivkap: @ArbelNathan @oriyarde 
It depends on what the other parameters are that are not relevant, as you asked above.
In any case - 
1. If we keep this note we should add a comma
      # pool, SpaceEfficiency**,** and snapshot_name_prefix

2. We can also write non-secret parameters in the example file itself and always add a note / text in the documentation itself with more specific information (ie. the list of what parameters can be provided)

deploy/kubernetes/examples/demo-secret-configuration-file.json
rivkap: agree with @zingero  just be sure to keep the hyphenation/underscores to be consistent as possible - especially within the same files

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
zingero: @oriyarde your suggestion doesn't refer to not relevant parameters, but all non-secret parameters.

oriyarde: @zingero what is the difference? why can't we refer to all possible `by_system_id` parameters as "non-secret parameters"?

deploy/kubernetes/examples/demo-secret-configuration-file.json
oriyarde: then how about `.../demo-region` and `.../demo-zone`?

deploy/kubernetes/examples/demo-secret-configuration-file.json
```diff
@@ -0,0 +1,24 @@
+{
```
oriyarde, line 1: is `file` really necessary in the filename?

deploy/kubernetes/examples/demo-secret-configuration-file.json
```diff
@@ -0,0 +1,24 @@
...
+      {
+        "topology.block.csi.ibm.com/zone": "demo-zone",
+        "topology.block.csi.ibm.com/dev": "demo-dev"
+      }
+    ]
+  },
+  "demo-system-id2": {
+    "username": "demo-username",
+    "password": "demo-password",
+    "management_address": "demo-management-address",
```
oriyarde, line 16: consider adding "2" suffix to the second system values

deploy/kubernetes/examples/demo-storageclass-using-secret-from-configuration.yaml
```diff
@@ -0,0 +1,25 @@
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: demo-storageclass
+provisioner: block.csi.ibm.com
+volumeBindingMode: WaitForFirstConsumer  # Optional. Required for Topology Awareness support.
+parameters:
+  # a pool value MUST be provided for all systems.
```
oriyarde, line 8: IMO this comment combined with `pool: demo-pool                # Optional.` is confusing.
I think we should leave `pool` without a comment, as was done for the regular storageclass

deploy/kubernetes/examples/demo-storageclass-using-secret-from-configuration.yaml
```diff
@@ -0,0 +1,25 @@
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: demo-storageclass
+provisioner: block.csi.ibm.com
+volumeBindingMode: WaitForFirstConsumer  # Optional. Required for Topology Awareness support.
```
oriyarde, line 6: consider removing `support`, since I don't think it adds any information

deploy/kubernetes/examples/demo-storageclass-using-secret-from-configuration.yaml
```diff
@@ -0,0 +1,25 @@
...
+  SpaceEfficiency: deduplicated  # Optional.
+
+  csi.storage.k8s.io/provisioner-secret-name: demo-secret-with-config
+  csi.storage.k8s.io/provisioner-secret-namespace: default
+  csi.storage.k8s.io/controller-publish-secret-name: demo-secret-with-config
+  csi.storage.k8s.io/controller-publish-secret-namespace: default
+  csi.storage.k8s.io/controller-expand-secret-name: demo-secret-with-config
+  csi.storage.k8s.io/controller-expand-secret-namespace: default
+
+  csi.storage.k8s.io/fstype: xfs   # Optional. Values ext4\xfs. The default is ext4.
```
oriyarde, line 23: any reason to have 3 spaces and not 2, before the number sign?

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: snapshot.storage.k8s.io/v1beta1
+kind: VolumeSnapshotClass
+metadata:
+  name: demo-snapshotclass
```
oriyarde, line 4: should we use a unique name for each class yaml?

deploy/kubernetes/examples/demo-secret-configuration-file.json
```diff
@@ -0,0 +1,24 @@
+{
+  "demo-system-id1": {
```
oriyarde, line 2: either add "1" suffix to all values of this storage system, or remove it from here

deploy/kubernetes/examples/demo-storageclass-using-secret-from-configuration.yaml
```diff
@@ -0,0 +1,25 @@
...
+                  "demo-system-id2":{"pool":"demo-pool2","volume_name_prefix":"demoPVC2"}}'  # Optional.
+  pool: demo-pool                # Optional.
+  SpaceEfficiency: deduplicated  # Optional.
+
+  csi.storage.k8s.io/provisioner-secret-name: demo-secret-with-config
+  csi.storage.k8s.io/provisioner-secret-namespace: default
+  csi.storage.k8s.io/controller-publish-secret-name: demo-secret-with-config
+  csi.storage.k8s.io/controller-publish-secret-namespace: default
+  csi.storage.k8s.io/controller-expand-secret-name: demo-secret-with-config
+  csi.storage.k8s.io/controller-expand-secret-namespace: default
```
oriyarde, line 21: could we follow the CSI-2885 approach for new class yamls, even before the rest of them are aligned?

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
zingero: @oriyarde because parameters like `csi.storage.k8s.io/snapshotter-secret-name` and `csi.storage.k8s.io/snapshotter-secret-namespace` are secret-related, but the comment doesn't refer to these parameters.

oriyarde: then how about "non-secret-related parameters"?

zingero: Sounds great

deploy/kubernetes/examples/demo-secret-configuration-file.json
ArbelNathan: how about more general? like `demo-suffix1` and `demo-suffix2`

deploy/kubernetes/examples/demo-secret-configuration-file.json
ArbelNathan: it is, keys and values were always different and I don't think we should change that..

deploy/kubernetes/examples/demo-secret-configuration-file.json
ArbelNathan: the command containing `from-file` so I thought to "connect" it..

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
ArbelNathan: what are you suggesting for the name?

deploy/kubernetes/examples/demo-secret-configuration-file.json
oriyarde: can said `from-file` command be used to create our regular secret?

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: snapshot.storage.k8s.io/v1beta1
```
oriyarde, line 1: would `/v1` work?

oriyarde, line 1: please align this filename to match the new storageclass filename (just with "snapshot" instead of "storage")

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
oriyarde: depends on its final filename

oriyarde: and regardless of this ticket:
I also noticed with have a somewhat similar issue with the example statefulset names.
I think we should consider removing the non-combined statefulset yamls, and remove the "combined" from its name, to have just one, comprehensive statefulset.
what do you think?

deploy/kubernetes/examples/demo-secret-configuration-file.json
oriyarde: could you give an example of an existing "mixed-convention" value?

ArbelNathan: within the same file?
in storageclass:
keys:
csi.storage.k8s.io/controller-expand-secret-name
volume_name_prefix
values:
demo-pool

oriyarde: no, not same file, same value.
there is a **key** difference between key and value.

regarding `demoPVC` - IMO it should be something like `demo-prefix`.

deploy/kubernetes/examples/demo-secret-configuration-file.json
ArbelNathan: I dont think so

deploy/kubernetes/examples/demo-storageclass-using-secret-from-configuration.yaml
ArbelNathan: it's aligned with the comment below

oriyarde: any reason to have 6 space and not 5, before the number sign, in the comment below?

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
ArbelNathan: IDK

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
ArbelNathan: what final name? is there any conversation about it?

ArbelNathan: and sure, sounds good, do you want me to change it in this ticket?

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
ArbelNathan: it is aligned

deploy/kubernetes/examples/demo-secret-configuration-file.json
ArbelNathan: what is the AI?

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
oriyarde: could you check? we don't want to add the old version if we don't have to

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
oriyarde: yes, there is [a conversation about it](https://github.com/IBM/ibm-block-csi-driver/pull/353#discussion_r682488910).
if @zingero is ok with it, it's fine by me

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
ArbelNathan: done

deploy/kubernetes/examples/demo-secret-configuration-file.json
ArbelNathan: ok, done

deploy/kubernetes/examples/demo-secret-configuration-file.json
ArbelNathan: changed

deploy/kubernetes/examples/demo-secret-configuration-file.json
ArbelNathan: done

deploy/kubernetes/examples/demo-storageclass-using-secret-from-configuration.yaml
ArbelNathan: aligned

deploy/kubernetes/examples/demo-storageclass-using-secret-from-configuration.yaml
ArbelNathan: done, you want me to align it?

deploy/kubernetes/examples/demo-secret-configuration-file.json
oriyarde: I [meant](https://github.com/IBM/ibm-block-csi-driver/pull/353#discussion_r682355801) that the keys suffix should contain `demo-` as well, to show that it's flexible as well

ArbelNathan: done

deploy/kubernetes/examples/demo-storageclass.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: demo-storageclass
 provisioner: block.csi.ibm.com
 parameters:
-  SpaceEfficiency: deduplicated   # Optional.
+  SpaceEfficiency: demo-space-efficiency  # Optional.
```
oriyarde, line 7: for non-user-defined values, our current examples convention is to use one of the possible values

deploy/kubernetes/examples/demo-storageclass.yaml
```diff
@@ -15,5 +15,5 @@ parameters:
   csi.storage.k8s.io/controller-expand-secret-namespace: default
 
   csi.storage.k8s.io/fstype: xfs   # Optional. Values ext4\xfs. The default is ext4.
-  volume_name_prefix: demoPVC      # Optional.
+  volume_name_prefix: demo-prefix  # Optional.
```
oriyarde, line 18: should we use a prefix value that also works "out-of-the-box" for ds8k a well? e.g. just `demo`?

deploy/kubernetes/examples/demo-storageclass-using-secret-from-configuration.yaml
oriyarde: I meant that this comment should be removed as well

deploy/kubernetes/examples/demo-storageclass.yaml
ArbelNathan: but there is no value that you can put there that is actually cross storage.

deploy/kubernetes/examples/demo-storageclass.yaml
ArbelNathan: I understand what you are saying but it's not self-explanatory, not 'demo' nor 'prefix', and the previous value was not "ds8k friendly" too

deploy/kubernetes/examples/demo-storageclass-using-secret-from-configuration.yaml
ArbelNathan: non `Optional` == required?

deploy/kubernetes/examples/demo-storageclass-using-secret-from-configuration.yaml
oriyarde: if @zingero is ok with it, sure

deploy/kubernetes/examples/demo-storageclass.yaml
oriyarde: true, but still, we use predefined values that work at least for svc.
this parameter is not even relevant for xiv, so it's not like `demo-space-efficiency` is more "cross storage"

deploy/kubernetes/examples/demo-storageclass.yaml
oriyarde: I agree.
@zingero what do you prefer?

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
zingero: I'm ok with it

deploy/kubernetes/examples/demo-storageclass.yaml
zingero: I don't understand how ds8k is different in this case.
`demo-prefix` looks good

deploy/kubernetes/examples/demo-storageclass-using-secret-from-config.yaml
```diff
@@ -0,0 +1,21 @@
...
+parameters:
+  # a pool value MUST be provided for all systems.
+  # non secret related parameters can be provided in by_system_id parameter for each system or by parameter name as default.
+
+  by_system_id: '{"demo-system-id-1":{"pool":"demo-pool-1","SpaceEfficiency":"demo-space-efficiency-1","volume_name_prefix":"demo-prefix-1"},
+                  "demo-system-id-2":{"pool":"demo-pool-2","volume_name_prefix":"demo-prefix-2"}}'  # Optional.
+  pool: demo-pool
+  SpaceEfficiency: demo-space-efficiency  # Optional.
+
+  csi.storage.k8s.io/secret-name: demo-secret-with-config
```
oriyarde, line 16: IMO `demo-secret-from-config` is better, but since we want to refer to this secret in other places, `demo-config-secret` is even better

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
oriyarde: oh, you're right.
could we at least shorten them?
how about e.g. `demo-snapshotclass-config-secret.yaml`?

deploy/kubernetes/examples/demo-storageclass.yaml
ArbelNathan: `thin` will be the most "cross storage"y

deploy/kubernetes/examples/demo-storageclass-using-secret-from-configuration.yaml
ArbelNathan: done

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
ArbelNathan: done

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
ArbelNathan: done

deploy/kubernetes/examples/demo-storageclass.yaml
ArbelNathan: I changed it to thin

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
oriyarde: why without hyphen? "non" as a separate word is [obsolete](https://en.wiktionary.org/wiki/non#English) and [used only in phrases from Latin](https://en.wiktionary.org/wiki/non-#Usage_notes)

oriyarde: @zingero is the "related" word really necessary?

deploy/kubernetes/examples/demo-storageclass.yaml
ArbelNathan: @zingero 
ds8k have a "strong" limit on name length so the prefix has 5 characters limit

deploy/kubernetes/examples/demo-storageclass-using-secret-from-config.yaml
ArbelNathan: done

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
oriyarde: suggestion:
`can be provided in by_system_id parameter for each system or by parameter name as default.` ->

1. `can be specified in by_system_id per system, or outside by_system_id as cross-system default.`
2. `can be specified in by_system_id per system, to override cross-system parameters.`
3. `can be specified in by_system_id per system, to override parameters outside by_system_id.``

deploy/kubernetes/examples/demo-storageclass.yaml
oriyarde: @zingero [ds8k max prefix length is 5](https://www.ibm.com/docs/en/stg-block-csi-driver/1.6.0?topic=configuration-creating-storageclass)

oriyarde: btw, for "config class" yamls, the only way I see to allow different ds8k prefixes is `demo1` and `demo2`

deploy/kubernetes/examples/demo-storageclass.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: demo-storageclass
 provisioner: block.csi.ibm.com
 parameters:
-  SpaceEfficiency: deduplicated   # Optional.
+  SpaceEfficiency: thin  # Optional.
```
oriyarde, line 7: ?

deploy/kubernetes/examples/demo-snapshotclass-config-secret.yaml
```diff
@@ -0,0 +1,17 @@
...
+metadata:
+  name: demo-snapshotclass
+driver: block.csi.ibm.com
+deletionPolicy: Delete
+parameters:
+  # non secret related parameters can be provided in by_system_id parameter for each system or by parameter name as default.
+
+  by_system_id: '{"demo-system-id-1":{"pool":"demo-pool-1","SpaceEfficiency":"thin","snapshot_name_prefix":"demo-prefix-1"},
+                  "demo-system-id-2":{"pool":"demo-pool-2","snapshot_name_prefix":"demo-prefix-2"}}'  # Optional.
+  pool: demo-pool                         # Optional. Use to create the snapshot on a different pool than the source.
```
oriyarde, line 12: why so much space?

deploy/kubernetes/examples/demo-snapshotclass.yaml
```diff
@@ -7,5 +7,6 @@ deletionPolicy: Delete
 parameters:
   csi.storage.k8s.io/snapshotter-secret-name: demo-secret
   csi.storage.k8s.io/snapshotter-secret-namespace: default
-  snapshot_name_prefix: demoSnapshot   # Optional.
-  pool: demo-pool                      # Optional. Use to create the snapshot on a different pool than the source.
+  snapshot_name_prefix: demo-prefix       # Optional.
```
oriyarde, line 10: why from 3 to 7 spaces and not to 2?

deploy/kubernetes/examples/demo-storageclass-config-secret.yaml
```diff
@@ -0,0 +1,20 @@
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: demo-storageclass
+provisioner: block.csi.ibm.com
+volumeBindingMode: WaitForFirstConsumer  # Optional. Required for Topology Awareness.
+parameters:
+  # non secret related parameters can be provided in by_system_id parameter for each system or by parameter name as default.
+
+  by_system_id: '{"demo-system-id-1":{"pool":"demo-pool-1","SpaceEfficiency":"thin","volume_name_prefix":"demo-prefix-1"},
```
oriyarde, line 10: wouldn't it be more realistic to use a space efficiency here that is different than the one below?

deploy/kubernetes/examples/demo-storageclass-config-secret.yaml
```diff
@@ -0,0 +1,20 @@
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: demo-storageclass
+provisioner: block.csi.ibm.com
+volumeBindingMode: WaitForFirstConsumer  # Optional. Required for Topology Awareness.
```
oriyarde, line 6: in the context of a config storageclass yaml, I think the `Optional` here is still TBD, since current replication design doesn't allow a storageclass with a config secret (without topology)

deploy/kubernetes/examples/demo-storageclass-config-secret.yaml
```diff
@@ -0,0 +1,20 @@
...
+  by_system_id: '{"demo-system-id-1":{"pool":"demo-pool-1","SpaceEfficiency":"thin","volume_name_prefix":"demo-prefix-1"},
+                  "demo-system-id-2":{"pool":"demo-pool-2","volume_name_prefix":"demo-prefix-2"}}'  # Optional.
+  pool: demo-pool
+  SpaceEfficiency: thin  # Optional.
+
+  csi.storage.k8s.io/secret-name: demo-config-secret
+  csi.storage.k8s.io/secret-namespace: default
+
+  csi.storage.k8s.io/fstype: xfs   # Optional. Values ext4\xfs. The default is ext4.
+  volume_name_prefix: demo-prefix  # Optional.
```
oriyarde, line 19: while we tidy things up here, could we also move the sidecar parameters to the bottom? so that the parameters are grouped by their recipient

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
oriyarde: I just realized that fstype is not secret related, but is also a sidecar parameter :/

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
oriyarde: so could we have a [unique name](https://github.com/IBM/ibm-block-csi-driver/pull/353#discussion_r682489291) for each class?

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
oriyarde: then I suggest: `non-csi.storage.k8s.io parameters`

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
oriyarde: a ticket number is also fine

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
ArbelNathan: what are you suggesting for the name?

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
zingero: no, it's not necessary

zingero: I agree with the new name

deploy/kubernetes/examples/demo-storageclass-using-secret-from-configuration.yaml
zingero: yes please

deploy/kubernetes/examples/demo-snapshotclass-config-secret.yaml
zingero: why                       not?
JK

deploy/kubernetes/examples/demo-storageclass.yaml
zingero: I still prefer having a readable and self-explanatory prefix like `demo-prefix`

oriyarde: correct me if I'm wrong, but:

1. this is the only value currently not allowing the examples to work "out-of-the-box" with ds8k
2. with my suggestion, the volume name on every storage would be something like `demo_pvc-123...` (which IMO is more natural looking than `demo-prefix_pvc-123`)
3. outside the storage, the value is always seen together with its field name (e.g. `volume_name_prefix: `), which clarifies it

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
oriyarde: same as the filename, minus the `.yaml`?

deploy/kubernetes/examples/demo-storageclass.yaml
oriyarde: if both of you still prefer `demo-prefix`, let me know and I will resolve this

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
oriyarde: so my current full suggestion:
`non-csi.storage.k8s.io parameters may be specified in by_system_id per system, or outside by_system_id as cross-system default.`

zingero: 👍🏻

deploy/kubernetes/examples/demo-storageclass.yaml
ArbelNathan: @zingero?

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
ArbelNathan: do you think that if I am already here I should change all `snapshotclass` to `volumesnapshotclass`? there is no such thing as snapshotclass..

deploy/kubernetes/examples/demo-storageclass-config-secret.yaml
ArbelNathan: so what are you suggesting?

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
oriyarde: I'm fine with either way.
@zingero?

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
ArbelNathan: blocked on CSI-2873.
v1 is for GA  external-snapshotter.

deploy/kubernetes/examples/demo-storageclass-config-secret.yaml
oriyarde: omit the `Optional` here, at least until (if) replication design changes in a way that would indeed make this optional

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
ArbelNathan: ok, done

deploy/kubernetes/examples/demo-storageclass-using-secret-from-configuration.yaml
ArbelNathan: done

deploy/kubernetes/examples/demo-snapshotclass-config-secret.yaml
ArbelNathan: done. aligned to comment under

deploy/kubernetes/examples/demo-snapshotclass.yaml
ArbelNathan: done

deploy/kubernetes/examples/demo-storageclass-config-secret.yaml
ArbelNathan: it's different from `demo-system-id-2` that has the default.

deploy/kubernetes/examples/demo-storageclass-config-secret.yaml
ArbelNathan: done

deploy/kubernetes/examples/demo-storageclass-config-secret.yaml
ArbelNathan: ok

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
oriyarde: perhaps the `or` should be `and/or` - I hope it clarifies which one overrides the other.
@rivkap what do you think?

deploy/kubernetes/examples/demo-storageclass-config-secret.yaml
ArbelNathan: done

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
oriyarde: if we want to rely on CSI-2873 for bumping to v1, it should be in the ticket **description**

deploy/kubernetes/examples/demo-storageclass-config-secret.yaml
oriyarde: should fstype come before the secret parameters (lexicographically and visually for the `# Optional.` comments)?

deploy/kubernetes/examples/demo-storageclass-config-secret.yaml
oriyarde: but if both systems eventually get 'thin', then is it really different?

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
oriyarde: `demo-volumesnapshotclass-config-secret` is more precise than:
`demo-snapshotclass-config-secret`, but very long

deploy/kubernetes/examples/demo-storageclass-config-secret.yaml
ArbelNathan: The second will get thick/none (that is the default..)

deploy/kubernetes/examples/demo-storageclass-config-secret.yaml
ArbelNathan: ok done

deploy/kubernetes/examples/demo-storageclass.yaml
zingero: It's ok to be non "out of the box" since it's a demo example. Our clients probably won't want to use this prefix even without the 5 characters limitation.

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
zingero: Long is better than short if we add precision or align the names or make it better to understand in lots of cases.
This is one of the cases IMO.

deploy/kubernetes/examples/demo-storageclass-config-secret.yaml
oriyarde: oh, then I misunderstood.
I thought it would get the value from [here](https://github.com/IBM/ibm-block-csi-driver/blob/6416b2e52d8e4465348fad54ab79664f0bd5b856/deploy/kubernetes/examples/demo-storageclass-config-secret.yaml#L13).
it would be nice if this nuance is clear from the yaml comment

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
ArbelNathan: done

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
ArbelNathan: done

deploy/kubernetes/examples/demo-storageclass-config-secret.yaml
ArbelNathan: oh, it will, I thought you only talked about  `demo-system-id-1` and `demo-system-id-2` as if there were no other optional parameters.
anyway, there is no other `"cross-storage" parameter to put in the second `SpaceEfficiency` parameter.

deploy/kubernetes/examples/demo-storageclass.yaml
oriyarde: IMO it is useful for users, especially new users, to be able to easily test the product with minimal configuration changes.
but I'm resolving this.

deploy/kubernetes/examples/demo-storageclass-config-secret.yaml
oriyarde: is cross-storage more important than for the example to make sense?
I thought the prefix value showed that (unfortunately) cross-storage is not so significant to us

zingero (dismissed):

deploy/kubernetes/examples/demo-storageclass-config-secret.yaml
ArbelNathan: there is a difference between "out-of-the-box" and "sense". "demo-prefix" is self-explanatory as it's a free string. `SpaceEfficiency` is not a free string parameter (and therefore we are not putting there demo-space-efficiency value) so it should contain one of the optional parameters that the user can put. no ds8k user can put `deduplication` so it's not a valid parameter to put there.

oriyarde: free string or not, I think it is more important for the example to make sense than to work on both systems.
@zingero what do you think?

ArbelNathan: it's not just "free string", to a non-SW user `deduplication` will not make any sense

zingero: It's an example. It doesn't have to work out of the box or have different values in the `by_system_id`/the parameter below.
From my point of view, we can put a dummy value like "demo-space-efficiency" or an actual valid value that applies to all of our storage types (svc, a9k, ds8k). Since a9k doesn't support any space efficiency value, I suggest not using an actual value, but "demo-space-efficiency". This will make the user pick a specific SE that works with their storage type.

ArbelNathan: @oriyarde what do you think?

oriyarde: I think we each hold a different opinion

oriyarde: perhaps the word "examples" is confusing.
the way I see it, ever since we moved to the "demo" approach, the goal was to provide a set of configurations that work with minimal change, to demonstrate the product flow

zingero: @ArbelNathan what do you think?

ArbelNathan: that we can use a third opinion

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
rivkap: @oriyarde 
I suggest:
`non-csi.storage.k8s.io parameters may be specified in by_system_id per system and/or outside by_system_id as the cross-system default. `

deploy/kubernetes/examples/demo-storageclass-config-secret.yaml
lisyukanna: IMO, the example yamls do not have to fit for each of the storages, as long as the documentation shows which `SpaceEfficiency` is supported for each of the storages

zingero: sounds great

oriyarde: > that we can use a third opinion

fourth*

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
oriyarde: thanks @rivkap

deploy/kubernetes/examples/demo-storageclass-config-secret.yaml
zingero: Thanks for another notification.

oriyarde: np

ArbelNathan: done

deploy/kubernetes/examples/demo-snapshotclass-using-secret-from-configuration.yaml
ArbelNathan: done

zingero (approved):

deploy/kubernetes/examples/demo-storageclass.yaml
```diff
@@ -4,16 +4,11 @@ metadata:
...
-  csi.storage.k8s.io/controller-publish-secret-name: demo-secret
-  csi.storage.k8s.io/controller-publish-secret-namespace: default
-  csi.storage.k8s.io/controller-expand-secret-name: demo-secret
-  csi.storage.k8s.io/controller-expand-secret-namespace: default
+  SpaceEfficiency: thin            # Optional.
+  volume_name_prefix: demo-prefix  # Optional.
 
   csi.storage.k8s.io/fstype: xfs   # Optional. Values ext4\xfs. The default is ext4.
-  volume_name_prefix: demoPVC      # Optional.
+  csi.storage.k8s.io/secret-name: demo-config-secret
```
oriyarde, line 12: config?

ArbelNathan: ok, bad copy paste 
done

zingero (dismissed):

deploy/kubernetes/examples/demo-volumesnapshot.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: snapshot.storage.k8s.io/v1beta1
 kind: VolumeSnapshot
 metadata:
-  name: demo-snapshot
+  name: demo-volumesnapshot
 spec:
   volumeSnapshotClassName: demo-snapshotclass
```
oriyarde, line 6: `demo-volumesnapshotclass`?

oriyarde: please use Ctrl+Shift+F when renaming in this dir

ArbelNathan: for what purpose? any change not in example yamls will be handled in aligning or documentation tickets

ArbelNathan: fixed the gap

oriyarde: > for what purpose? any change not in example yamls will be handled in aligning or documentation tickets

I meant while limiting the Ctrl+Shift+F to this dir

zingero (approved):

deploy/kubernetes/examples/demo-volumesnapshot.yaml
ArbelNathan: ok, any other AI?

oriyarde: nope

oriyarde (approved):
LGTM

PR #355, opened 2021-08-09
CSI-3103 move main from controller server

oriyarde:
prepare code structure for upcoming [csi-addons](https://github.com/csi-addons/spec) addition

### DIFF
--- a/controller/controller_server/csi_controller_server.py
+++ b/controller/controller_server/csi_controller_server.py
@@ -23,7 +23,7 @@ from controller.csi_general import csi_pb2_grpc
 logger = get_stdout_logger()
 
 
-class CSIControllerServicer(csi_pb2_grpc.ControllerServicer):
+class ControllerServicer(csi_pb2_grpc.ControllerServicer):
     """
     gRPC server for Digestor Service
     """
--- a/controller/controller_server/csi_controller_server.py
+++ b/controller/controller_server/csi_controller_server.py
@@ -1,7 +1,4 @@
 import os.path
-import time
-from concurrent import futures
-from argparse import ArgumentParser
 
 import grpc
 import yaml
@@ -13,28 +10,25 @@ import controller.controller_server.utils as utils
 from controller.array_action import messages
 from controller.array_action.storage_agent import get_agent, detect_array_type
 from controller.common import settings
-from controller.common.csi_logger import get_stdout_logger, set_log_level
+from controller.common.csi_logger import get_stdout_logger
 from controller.common.node_info import NodeIdInfo
 from controller.common.utils import set_current_thread_name
+from controller.controller_server import messages as controller_messages
 from controller.controller_server.errors import ObjectIdError, ValidationException
 from controller.controller_server.exception_handler import handle_common_exceptions, handle_exception, \
     build_error_response
-from controller.controller_server import messages as controller_messages
 from controller.csi_general import csi_pb2
 from controller.csi_general import csi_pb2_grpc
 
 logger = get_stdout_logger()
 
 
-class ControllerServicer(csi_pb2_grpc.ControllerServicer):
+class CSIControllerServicer(csi_pb2_grpc.ControllerServicer):
     """
     gRPC server for Digestor Service
     """
 
-    def __init__(self, array_endpoint):
-
-        self.endpoint = array_endpoint
-
+    def __init__(self):
         my_path = os.path.abspath(os.path.dirname(__file__))
         path = os.path.join(my_path, "../../common/config.yaml")
 
@@ -477,14 +471,14 @@ class ControllerServicer(csi_pb2_grpc.ControllerServicer):
         logger.info("finished ControllerGetCapabilities")
         return res
 
-    def __get_identity_config(self, attribute_name):
+    def _get_identity_config(self, attribute_name):
         return self.cfg['identity'][attribute_name]
 
     @handle_common_exceptions(csi_pb2.GetPluginInfoResponse)
     def GetPluginInfo(self, _, context):
         logger.info("GetPluginInfo")
-        name = self.__get_identity_config("name")
-        version = self.__get_identity_config("version")
+        name = self._get_identity_config("name")
+        version = self._get_identity_config("version")
 
         if not name or not version:
             message = "plugin name or version cannot be empty"
@@ -531,7 +525,7 @@ class ControllerServicer(csi_pb2_grpc.ControllerServicer):
         logger.info("GetPluginCapabilities")
         service_type = csi_pb2.PluginCapability.Service.Type
         volume_expansion_type = csi_pb2.PluginCapability.VolumeExpansion.Type
-        capabilities = self.__get_identity_config("capabilities")
+        capabilities = self._get_identity_config("capabilities")
         capability_list = []
         service_capabilities = capabilities.get('Service')
         volume_expansion_capability = capabilities.get('VolumeExpansion')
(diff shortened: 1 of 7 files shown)


controller/scripts/entrypoint.sh
```diff
@@ -1,2 +1,2 @@
 #!/bin/bash
-exec python3.8 /driver/controller/controller_server/csi_controller_server.py $@
+exec python3.8 /driver/controller/controller_server/main.py $@
```
zingero, line 2: Why not `__main__`?

controller/controller_server/csi_controller_server.py
```diff
@@ -13,28 +10,25 @@
...
     build_error_response
-from controller.controller_server import messages as controller_messages
 from controller.csi_general import csi_pb2
 from controller.csi_general import csi_pb2_grpc
 
 logger = get_stdout_logger()
 
 
-class ControllerServicer(csi_pb2_grpc.ControllerServicer):
+class CSIControllerServicer(csi_pb2_grpc.ControllerServicer):
```
zingero, line 26: I suggest aligning the file name and the class name.

controller/controller_server/main.py
```diff
@@ -0,0 +1,64 @@
...
+        csi_pb2_grpc.add_IdentityServicer_to_server(self, controller_server)
+
+        # bind the server to the port defined above
+        # controller_server.add_insecure_port('[::]:{}'.format(self.server_port))
+        # controller_server.add_insecure_port('unix://{}'.format(self.server_port))
+        controller_server.add_insecure_port(self.endpoint)
+
+        logger.info("Controller version: {}".format(self._get_identity_config("version")))
+
+        # start the server
```
zingero, line 37: Do we want to remove redundant comments in this PR?

controller/controller_server/main.py
```diff
@@ -0,0 +1,64 @@
...
+
+from controller.common import settings
+from controller.common.csi_logger import get_stdout_logger, set_log_level
+from controller.controller_server.csi_controller_server import CSIControllerServicer
+from controller.csi_general import csi_pb2_grpc
+
+logger = get_stdout_logger()
+
+
+class ControllerServicer(CSIControllerServicer):
```
zingero, line 15: I don't think the main file should contain this class.
In addition, I don't think this class should inherit from the big servicer class.

controller/controller_server/csi_controller_server.py
oriyarde: I'm not sure it's related to this PR

controller/controller_server/main.py
oriyarde: nope, the ticket is replication, this PR is separate just to reduce the replication PR diff

controller/controller_server/main.py
oriyarde: then what do you suggest?
I was thinking of changing this to `ControllerServicer(CSIControllerServicer, ReplicationControllerServicer)` in the upcoming PR

controller/scripts/entrypoint.sh
oriyarde: you mean add [`-m`](https://docs.python.org/3.8/library/__main__.html)?
couldn't it change some import behavior? it sounds to me like a separate ticket

controller/controller_server/csi_controller_server.py
zingero: Then why did you rename the class?

controller/scripts/entrypoint.sh
zingero: Yes. The CSI controller is a module with a "main" file. 
What do you mean by import behavior? No file should import the main file.

controller/controller_server/main.py
zingero: I think we should have a design review session with @ArbelNathan before doing so.

controller/scripts/entrypoint.sh
oriyarde: I mean in general, `-m` has a general effect on imports in the project (e.g. [relative imports](https://stackoverflow.com/a/22250157)).
and I don't think this PR should affect the global scope of the project

controller/controller_server/csi_controller_server.py
oriyarde: to distinguish between csi and the upcoming non-csi controller logic

controller/controller_server/main.py
oriyarde: do you have any suggestion to prepare for the session?

controller/controller_server/csi_controller_server.py
zingero: Usually, class names go together with the filename.

controller/controller_server/main.py
zingero: The session is done. We agreed to have a composition design instead of inheritance.

controller/scripts/entrypoint.sh
zingero: Next time :)

controller/controller_server/csi_controller_server.py
oriyarde: I agree, but since it was already unaligned (if anything, this diff brings it closer to be aligned), I'm not sure it's related to this PR

controller/controller_server/main.py
oriyarde: done

controller/controller_server/csi_controller_server.py
zingero: If so, I suggest reverting this change since it's not related to moving out the main function from this file.

controller/controller_server/main.py
zingero: Actually, it's not (fully) done. I still think we should have the class in a different file.

controller/controller_server/csi_controller_server.py
oriyarde: so you want me to open a separate PR just for this change?

controller/controller_server/main.py
oriyarde: and what would be the name of said file?

controller/controller_server/csi_controller_server.py
zingero: I want to align the file name with the class name, but you don't want to do that now because it's not related to the PR.
Thus I suggest not doing things that aren't related to this PR.
If you want to do it in a separate PR, I don't mind.

controller/controller_server/main.py
zingero: path/to/folder/class_name.py

Since this class isn't a controller servicer, but it's controller servicer manager/owner/initializator/starter (other suggestion?) I suggest renaming it accordingly.

oriyarde: done

controller/controller_server/csi_controller_server.py
zingero: What's the difference? MY suggestion is related to this PR more than it's related to the big replication addition PR.
@ArbelNathan , WDYT?

oriyarde: your suggestion is an improvement for a general gap in the project (class-filename inconsistency), which I think should be addressed in its own PR, while this change is meant to make current controller class more specific, to distinguish between CSI and Replication controller classes

zingero: At the moment there is no replication controller class. No need to distinguish now between them.

ArbelNathan: I think it could be done in this PR but if there is a general gap in the whole project so it could be done on a different ticket.

oriyarde: I suggested to do this distinction before the replication addition to reduce the replication diff.
so back to [my question](https://github.com/IBM/ibm-block-csi-driver/pull/355#discussion_r687647059) - would it be better to include this as part of the replication PR?

zingero: Yes.

zingero (approved):

controller/controller_server/csi_controller_server.py
```diff
@@ -477,14 +471,14 @@ def ControllerGetCapabilities(self, request, context):
...
-    def __get_identity_config(self, attribute_name):
+    def get_identity_config(self, attribute_name):
         return self.cfg['identity'][attribute_name]
 
     @handle_common_exceptions(csi_pb2.GetPluginInfoResponse)
     def GetPluginInfo(self, _, context):
         logger.info("GetPluginInfo")
-        name = self.__get_identity_config("name")
-        version = self.__get_identity_config("version")
+        name = self.get_identity_config("name")
```
ArbelNathan, line 480: why we are not getting the `self.cfg['identity']` once (maybe in init) and then "extracting" the info from there?

oriyarde: sounds like a good idea for a different ticket

ArbelNathan: as I recall our way of work is to fix and approve code where ever the PR is reaching but if we are in a problem with time here please open a ticket for it and I will resolve

zingero: We're not in a problem with time here since we waste a lot of time by commenting, opening tickets, etc.
Changing the code is faster in simple cases like this one IMO.

oriyarde: CSI-3238

ArbelNathan (approved):
LGTM

controller/controller_server/csi_controller_server.py
oriyarde: we're in a problem with time here because we are close to code freeze.
and commenting, opening tickets etc... is faster than increasing the diff, which potentially increases the amount of comments, tickets to open, etc...
in addition, the context of this whole PR is already a "collateral" change for the main ticket, so we shouldn't have "collateral" changes to the "collateral" changes in the same ticket, it has got to end somewhere

PR #356, opened 2021-08-18
Improvement/csi 3237 exception catching coverage

ArbelNathan:


controller/array_action/array_mediator_ds8k.py
```diff
@@ -167,20 +167,12 @@ def _connect(self):
...
-            raise ConnectionError()
-        except exceptions.ClientException as e:
             logger.error(
-                'Failed to connect to DS8K array {}, reason is {}'.format(
-                    self.service_address,
-                    e.details
-                )
-            )
+                'Failed to connect to DS8K array {}, reason is {}'.format(self.service_address, ex.details))
             raise ConnectionError()
```
oriyarde, line 176: while you're at it, could you change this to `raise ex`/`raise`, so we won't lose the exception info?

controller/array_action/array_mediator_svc.py
```diff
@@ -502,15 +491,20 @@ def _delete_fcmap(self, fcmap_id, force):
             self.client.svctask.rmfcmap(object_id=fcmap_id, force=force)
         except (svc_errors.CommandExecutionError, CLIFailureError) as ex:
             if not is_warning_message(ex.my_message):
-                logger.warning("Failed to delete fcmap '{0}': {1}".format(fcmap_id, ex))
+                logger.exception("Failed to delete fcmap '{0}': {1}".format(fcmap_id, ex))
```
oriyarde, line 494: `logger.error` is sufficient, no need to log the traceback (same for the other logger calls)

controller/tests/array_action/ds8k/test_array_mediator_ds8k.py
```diff
@@ -706,3 +711,8 @@ def test_expand_volume_extend_not_enough_space_error(self):
         self.client_mock.extend_volume.side_effect = [ClientException("500", message="BE531465")]
         with self.assertRaises(array_errors.NotEnoughSpaceInPool):
             self.array.expand_volume(volume_id="test_id", required_bytes=10)
+
+    def test_expand_volume_extend_raise_other_error(self):
```
oriyarde, line 715: no need for `other` IMO

controller/array_action/array_mediator_svc.py
zingero: +1

controller/array_action/array_mediator_ds8k.py
zingero: We catch the `ClientException` exception and raise a different exception, `ConnectionError`.
If we decide to link them together (I'm totally into it), I suggest using `from`:
 ```suggestion
            raise ConnectionError() from ex
```

File: controller/array_action/array_mediator_ds8k.py [hunk 4]
```diff
@@ -274,15 +266,6 @@ def _create_api_volume(self, name, size_in_bytes, space_efficiency, pool_id):
                 )
             )
             raise array_errors.VolumeCreationError(name)
-        except (exceptions.ClientError, exceptions.ClientException) as ex:
-            logger.error(
```
zingero, line 278: Why don't we want to print that?

zingero: nvm, got it

controller/array_action/array_mediator_ds8k.py
```diff
@@ -292,11 +275,12 @@ def _extend_volume(self, api_volume, new_size_in_bytes):
         try:
             self.client.extend_volume(volume_id=api_volume.id,
                                       new_size_in_bytes=new_size_in_bytes)
-        except exceptions.NotFound:
-            raise array_errors.ObjectNotFoundError(api_volume.id)
-        except (exceptions.ClientError, exceptions.ClientException) as ex:
+        except exceptions.ClientException as ex:
```
zingero, line 278: I'm not 100% like the new implementation. I think it's because the language supports having a couple of `except` statements for one `try` statement and we choose to do it in a different way. What's the reason for this change?

controller/array_action/array_mediator_ds8k.py
oriyarde: not sure if `ConnectionInfo` adds any useful info

controller/array_action/array_mediator_ds8k.py
ArbelNathan: that way we can catch everything we are looking for and reducing the chance of missing anything, plus, it's more "organized" this way.

controller/tests/array_action/ds8k/test_array_mediator_ds8k.py
ArbelNathan: done

controller/array_action/array_mediator_svc.py
ArbelNathan: done

oriyarde (dismissed):
LGTM

controller/array_action/array_mediator_ds8k.py
ArbelNathan: due to the fact we have no handler for `ConnectionInfo` I changed it to `raise ex`

controller/array_action/array_mediator_ds8k.py
zingero: We catch "everything" (`ClientException` is everything in this case) anyway, but if @oriyarde and you like this way more than my suggestion, it's fine by me.

oriyarde: https://youtu.be/ussCHoQttyQ

ArbelNathan: done

oriyarde (approved):
LGTM

zingero (approved):

PR #357, opened 2021-08-19
Task/CSI-1870 change community whitelist to blacklist for each storage

tyichye19:


scripts/csi_test/csi_tests_to_skip_community_a9k
```diff
@@ -0,0 +1,6 @@
+GetCapacity
+ListVolumes
+ValidateVolumeCapabilities
+volume lifecycle
+ListSnapshots
+NodeGetVolumeStats
```
zingero, line 6: Add a newline at EOF (in all relevant files)

tyichye19: done

Dockerfile-csi-test
```diff
@@ -15,6 +15,7 @@
 FROM golang:1.11
 
 ARG CSI_PARAMS=csi_params
+ARG STORAGE_LABEL=community_svc
```
zingero, line 18: Why do we use SCV in this case? Do we have similar labels for a9k/ds8k?

oriyarde, line 18: could you remove the `community_` prefix? it should be implied from the context.
also I think in the project `STORAGE_LABEL` would be called `ARRAY_TYPE`

scripts/ci/run_csi_test_client.sh
```diff
@@ -8,5 +8,5 @@ else
     csi_params='csi_params'
```
oriyarde, line 8: I think it would be better to pass the full tests to skip file path, from this file to the dockerfile

Dockerfile-csi-test
tyichye19: it's just a default value. 
we have label for each storage and we pass the relevant label through the docker build command

scripts/ci/run_csi_test_client.sh
oriyarde: that way you could easily add logic to unify a new general "tests to skip" file with a storage-specific "tests to skip" file

tyichye19: done

zingero (approved):

scripts/ci/run_csi_test_client.sh
oriyarde: could you also split them so that we won't maintain duplications?

tyichye19: what do you mean?

oriyarde: I mean that the common tests to skip should be maintained in a separate file, and not duplicated for each array type

tyichye19: now I understand your first comments, sorry.
done.

zingero (approved):

oriyarde (approved):
LGTM

PR #359, opened 2021-08-19
CSI 3191 checking not available in XIV

zingero:


controller/array_action/array_mediator_xiv.py
```diff
@@ -19,6 +19,7 @@
 LUN_IS_ALREADY_IN_USE_ERROR = "LUN is already in use"
 UNDEFINED_MAPPING_ERROR = "The requested mapping is not defined"
 NO_ALLOCATION_SPACE_ERROR = "No space to allocate to the volume"
+NOT_AVAILABLE = "Not Available"
```
ArbelNathan, line 22: I prefer if we could be not case sensitive..

zingero: What do you mean? This is the value we get from the storage.

oriyarde: I assume he meant to compare both to normalized values (e.g. `actual_value.lower() == NOT_AVAILABLE.lower()`)

oriyarde: not sure if it is better

oriyarde (approved):
LGTM

controller/array_action/array_mediator_xiv.py
ArbelNathan: we are doing it with all error codes but if we think it's not necessary we can leave it

ArbelNathan (approved):
LGTM

PR #365, opened 2021-08-31
Feature/csi 3130 support svc 8.4.2 lsip

ArbelNathan:


controller/array_action/array_mediator_interface.py
```diff
@@ -266,19 +266,20 @@ def delete_snapshot(self, snapshot_id):
         raise NotImplementedError
 
     @abstractmethod
-    def get_iscsi_targets_by_iqn(self):
+    def get_iscsi_targets_by_iqn(self, host_name=None):
```
oriyarde, line 269: actually I think it would be simpler to avoid default values in the interface, because the caller would also pass it

controller/array_action/array_mediator_svc.py
```diff
@@ -783,8 +784,11 @@ def _get_array_iqns_by_node_id(self):
         logger.debug("Found iqns by node id: {}".format(array_iqns_by_id))
         return array_iqns_by_id
 
-    def _list_ip_ports(self):
+    def _list_ip_ports(self, port_set_id):
```
oriyarde, line 787: consider using the same name as in the storage, i.e. `portset` instead of `port_set`, similar to how we did with `fcmap` (i.e. not `fc_map`)

controller/array_action/array_mediator_svc.py
```diff
@@ -840,3 +845,14 @@ def get_array_fc_wwns(self, host_name):
             logger.error(msg="Failed to get array fc wwn. Reason "
                              "is: {0}".format(ex))
             raise ex
+
+    def _get_host_by_name(self, host_name):
+        filter_value = 'name={}'.format(host_name)
+        hosts_list = self.client.svcinfo.lshost(filtervalue=filter_value).as_list
+        if not hosts_list:
+            raise array_errors.HostNotFoundError(host_name)
+        return hosts_list[0]
```
oriyarde, line 854: shouldn't it be `as_single_element` instead of `as_list` + `[0]`?

controller/array_action/array_mediator_interface.py
oriyarde: same for [existing method](https://github.com/IBM/ibm-block-csi-driver/blob/a91fb4de07da7c95d94be308c3d1006360fc6d9d/controller/array_action/array_mediator_interface.py#L287)s
image

ArbelNathan: done, please open a ticket for non-scoped issues..

controller/array_action/array_mediator_svc.py
ArbelNathan: done

controller/array_action/array_mediator_interface.py
oriyarde: for existing gaps of this type, we already have [this](https://github.com/IBM/ibm-block-csi-driver/blob/8a2506fc14bde06a63937c31a7e135596c3642b3/controller/scripts/lint.ini#L30), so no need for a separate ticket here

controller/array_action/array_mediator_svc.py
oriyarde: now it's no longer a host *list*

erantzabari (approved):

controller/array_action/array_mediator_svc.py
ArbelNathan: done

oriyarde (approved):
LGTM

PR #366, opened 2021-09-02
CSI-3103 add replication support

oriyarde:
https://github.com/csi-addons/spec


controller/array_action/array_action_types.py
```diff
@@ -33,3 +33,13 @@ def __init__(self, host_id, host_name, iscsi_names, wwns):
         self.name = host_name
         self.iscsi_names = iscsi_names
         self.wwns = wwns
+
+
+class Replication:
```
ArbelNathan, line 38: why not do it as a dataclass?

controller/csi_general/replication_pb2_grpc.py
```diff
@@ -0,0 +1,116 @@
+# Generated by the gRPC Python protocol compiler plugin. DO NOT EDIT!
```
ArbelNathan, line 1: will it bee part of CSI-1377?

controller/scripts/csi_pb2.sh
```diff
@@ -1,6 +1,22 @@
-#!/bin/bash
+#!/bin/bash -e
```
ArbelNathan, line 1: is it working? is it part of CSI-1377?

controller/controller_server/addons_server.py
```diff
@@ -0,0 +1,160 @@
...
+
+class ReplicationControllerServicer(pb2_grpc.ControllerServicer):
+    @handle_common_exceptions(pb2.EnableVolumeReplicationResponse)
+    def EnableVolumeReplication(self, request, context):
+        set_current_thread_name(request.volume_id)
+        logger.info("EnableVolumeReplication")
+        utils.validate_addons_request(request)
+
+        volume_id_info = utils.get_volume_id_info(request.volume_id)
+        volume_id = volume_id_info.object_id
```
ArbelNathan, line 24: it starting to be confusing, can we consider changing object_id to wwn or something like that?

controller/controller_server/addons_server.py
```diff
@@ -0,0 +1,160 @@
...
+        volume_internal_id = volume_id_info.internal_id
+
+        other_volume_id_info = utils.get_volume_id_info(request.target_volume_id)
```
ArbelNathan, line 27: is `target_volume_id` is a thing that gonna be added to the [spec](https://github.com/csi-addons/spec/tree/main#enablevolumereplication)?

controller/scripts/csi_pb2.sh
oriyarde: yes, it is the 1st bullet (out of 4)

controller/controller_server/addons_server.py
oriyarde: not sure why this PR is the "start" of the confusion, it was like this before.
but in general, I agree we can consider `wwn`/`wwid`/`uuid`/etc... across the project (need to check what is the correct terminology here)

controller/controller_server/addons_server.py
oriyarde: yes, see https://github.com/csi-addons/spec/pull/22

controller/csi_general/replication_pb2_grpc.py
oriyarde: yes, all `.proto` files should be handled in the same manner

controller/array_action/array_action_types.py
oriyarde: not sure.
why were others changed to dataclass? why wasn't `Host` changed?

ArbelNathan: it's a good idea to change `Host` too. 
dataclass helpful methods ( e.g. "toString" and other)..

controller/controller_server/addons_server.py
ArbelNathan: great, tnx

controller/array_action/array_action_types.py
oriyarde: then sounds to me like it can be done across the project (i.e. at least for `Host` as well) in a separate PR.
we don't have automation for replication yet, so I'm not keen to make optional changes with potential breakage before GA

oriyarde: one disadvantage to convert to dataclass is that it would be harder to add logic to those classes later (e.g. `__eq__`)

controller/array_action/array_mediator_svc.py
```diff
@@ -842,3 +871,192 @@ def get_array_fc_wwns(self, host_name):
...
+        is_primary = self._is_replication_endpoint_primary(rcrelationship)
+        is_ready = self._is_replication_ready(rcrelationship)
+        return Replication(name=rcrelationship.name,
+                           volume_internal_id=volume_internal_id,
+                           other_volume_internal_id=other_volume_internal_id,
+                           copy_type=copy_type,
+                           is_primary=is_primary,
+                           is_ready=is_ready)
+
+    def _get_lsrcrelationship(self, filter_value):
```
ArbelNathan, line 920: `_get_cli_rcrelationship`?
AFAIK `cli` letting us know that the object that the method returns is "storage object" and not the svc command in the name..

controller/array_action/array_mediator_svc.py
```diff
@@ -842,3 +871,192 @@ def get_array_fc_wwns(self, host_name):
...
+                                                                          OTHER_VDISK_ID=other_cli_volume_id,
+                                                                          OTHER_CLUSTER_ID=other_system_id)
+        return self._get_lsrcrelationship(filter_value).as_list
+
+    def _get_rcrelationship(self, cli_volume_id, other_cli_volume_id, other_system_id):
+        rcrelationships = self._get_rcrelationships(cli_volume_id, other_cli_volume_id,
+                                                    other_system_id, as_master=True)
+        rcrelationships.extend(self._get_rcrelationships(cli_volume_id, other_cli_volume_id,
+                                                         other_system_id, as_master=False))
+        if len(rcrelationships) != 1:
```
ArbelNathan, line 949: is it possible?

controller/array_action/array_mediator_svc.py
```diff
@@ -842,3 +871,192 @@ def get_array_fc_wwns(self, host_name):
...
+            logger.warning('found {0} rcrelationships for volume id {1} '
+                           'with volume id {2} of system {3}'.format(len(rcrelationships),
+                                                                     cli_volume_id,
+                                                                     other_cli_volume_id,
+                                                                     other_system_id))
+            return None
```
ArbelNathan, line 955: why only a warning?

controller/array_action/array_mediator_svc.py
oriyarde: since there's already a `_get_rcrelationship`, I think `_get_cli_rcrelationship` would be confusing.
how about just `_lsrcrelationship`?

controller/array_action/array_mediator_svc.py
oriyarde: I think so, yes

controller/array_action/array_mediator_svc.py
oriyarde: should I change it to `error` and raise an exception?
in such scenario it will fail afterwards anyway

controller/controller_server/addons_server.py
```diff
@@ -0,0 +1,160 @@
...
+                return build_error_response(message, context, grpc.StatusCode.FAILED_PRECONDITION, response_type)
+            logger.info("found replication {} on system {}".format(replication.name, mediator.identifier))
+
+            self._ensure_volume_role_for_replication(mediator, replication, is_to_promote)
+
+        return response_type()
+
+    @handle_common_exceptions(pb2.PromoteVolumeResponse)
+    def PromoteVolume(self, request, context):
+        return self._ensure_volume_role(request, context, is_to_promote=True, response_type=pb2.PromoteVolumeResponse)
```
ArbelNathan, line 128: `ensure` sound like validation, it's not really a good name for a method that promotes or demotes..

oriyarde: `ensure` is not really validation - it means "make sure", i.e.:

1. be idempotent
2. fail if the state can't be reached

controller/controller_server/addons_server.py
```diff
@@ -0,0 +1,160 @@
...
+        other_volume_id_info = utils.get_volume_id_info(request.target_volume_id)
+        other_volume_internal_id = other_volume_id_info.internal_id
+
+        other_system_id = request.parameters.get(config.PARAMETERS_SYSTEM_ID)
+
+        connection_info = utils.get_array_connection_info_from_secrets(request.secrets)
+        with get_agent(connection_info, volume_id_info.array_type).get_mediator() as mediator:
+            replication = mediator.get_replication(volume_internal_id, other_volume_internal_id, other_system_id)
+            if not replication:
+                message = "could not find replication for volume internal id: {} " \
```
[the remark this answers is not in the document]
oriyarde: `volume wwn/wwid/uuid..` are not the "internal" id, so we can't use those.
and IMO `volume id` is ambiguous, since it can be both `wwn/wwid/uuid..` and "an internal" vdisk id

controller/controller_server/addons_server.py
ArbelNathan: sounds good

controller/controller_server/addons_server.py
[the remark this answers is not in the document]
ArbelNathan: `volume storage id`?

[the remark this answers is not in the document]
oriyarde: both `id`s are created by the storage, so that doesn't help differentiate between them either

controller/array_action/array_action_types.py
ArbelNathan: are we? it's just like [any other class ](https://www.tutorialspoint.com/data-classes-in-python-dataclasses#:~:text=The%20dataclasses%20is%20a%20new%20module%20added%20in,a%20few%20more%29%20for%20a%20user%20defined%20class.)in that regard.. (any custom one will shadow the automatically generated ones)

oriyarde: right, my bad, it won't be harder, just inappropriate

controller/array_action/array_mediator_svc.py
ArbelNathan: we have `get_volume` and `_get_cli_volume`..

controller/array_action/array_mediator_svc.py
ArbelNathan: I'm not sure, what is the actual meaning of getting here? what is the scenario?

controller/controller_server/addons_server.py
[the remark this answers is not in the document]
ArbelNathan: but the storage never calls wwn an id..

controller/array_action/array_action_types.py
ArbelNathan: why inappropriate?

controller/array_action/array_mediator_svc.py
oriyarde: `get_volume` is public, and `cli` helps differentiate between a "server" volume and a "mediator" volume.
but in this case, `replication` is the "server" object and "rcrelationship" is the "mediator" object, so `cli` doesn't help in this case

oriyarde: if we had `get_volume` and `_get_vdisk` for example, then we wouldn't need `cli` at all

controller/array_action/array_mediator_svc.py
oriyarde: e.g. two identical rcrelationships but with opposite directions (shouldn't happen of course)

controller/array_action/array_action_types.py
oriyarde: because dataclasses by their nature shouldn't contain much logic

controller/controller_server/addons_server.py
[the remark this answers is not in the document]
oriyarde: image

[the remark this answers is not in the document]
oriyarde: https://github.com/IBM/ibm-block-csi-driver/blob/d49fdfebe4bdf64c368e896a81a3b3bf5ef7bf57/controller/array_action/array_mediator_svc.py#L194

controller/array_action/array_action_types.py
ArbelNathan: great, we got none

oriyarde: that's why I [wrote "later"](https://github.com/IBM/ibm-block-csi-driver/pull/366#discussion_r705174108)

ArbelNathan: Do you reckon there will be changes in the future to the type classes?

oriyarde: it's hard to tell, we do need to split the big files somehow...
regardless of said disadvantage, I [don't think it's a must](https://github.com/IBM/ibm-block-csi-driver/pull/366#discussion_r705171484) for this PR

ArbelNathan: ok, I still think its better but it's not blocker IMO

controller/array_action/array_mediator_svc.py
ArbelNathan: and how do we now that `vdisk` is a "mediator" or "server" object

controller/array_action/array_mediator_svc.py
ArbelNathan: so its a bug it should raise an exception IMO..

controller/controller_server/addons_server.py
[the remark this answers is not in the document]
ArbelNathan: ok, I'm forfeit.

controller/array_action/array_mediator_svc.py
oriyarde: `vdisk` is an svc-only term

controller/array_action/array_mediator_svc.py
oriyarde: I agree, it would be better to raise an exception here

oriyarde: will be done in CSI-3394

ArbelNathan (approved):
LGTM

ArbelNathan (approved):
LGTM

PR #370, opened 2021-09-09
feature CSI-3389 replication example yamls

ArbelNathan:


deploy/kubernetes/examples/demo-volumereplication.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: VolumeReplication
+metadata:
+  name: demo-volumereplication
+  namespace: default
+spec:
+  volumeReplicationClass: demo-volumereplicationclass
+  replicationState: primary
+  dataSource:
+    kind: PersistentVolumeClaim
+    name: demo-pvc-file-system  # Should be in same namespace as VolumeReplication
```
oriyarde, line 11: missing EOL

deploy/kubernetes/examples/demo-volumereplicationclass.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: replication.storage.openshift.io/v1alpha1
+kind: VolumeReplicationClass
+metadata:
+  name: demo-volumereplicationclass
+spec:
+  provisioner: block.csi.ibm.com
+  parameters:
+    system_id: demo-system-id
```
oriyarde, line 8: this might create a false impression that this value is the same as the one in topology config secret key.
should we change the term in topology?

deploy/kubernetes/examples/demo-volumereplication.yaml
rivkap: "# Place in the same namespace as Volume Replication"
If "place" isn't the right word, please suggest an alternative. We should avoid using "should"

ArbelNathan: @oriyarde "place"?

deploy/kubernetes/examples/demo-volumereplicationclass.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: replication.storage.openshift.io/v1alpha1
+kind: VolumeReplicationClass
+metadata:
+  name: demo-volumereplicationclass
+spec:
+  provisioner: block.csi.ibm.com
+  parameters:
+    system_id: demo-system-id
+    copy_type: async  # Optional. Values sync\async. The default is sync.
```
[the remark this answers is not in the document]
ArbelNathan: in [storageclass](https://github.com/IBM/ibm-block-csi-driver/blob/feature/CSI-3389_replication_example_yamls/docs/content/configuration/csi_ug_config_create_storageclasses.md) we have `ext4`\`xfs`

deploy/kubernetes/examples/demo-volumereplication.yaml
oriyarde: this was copied from [here](https://github.com/csi-addons/volume-replication-operator/blob/faf6e90cec37092f153007e667a997d00d0cd76c/README.md#volumereplication)
"ensure"?

rivkap: We can write "Ensure that this is in the same namespace as VolumeReplication."

deploy/kubernetes/examples/demo-volumereplicationclass.yaml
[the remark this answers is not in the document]
rivkap: @ArbelNathan  can we change it there? Is this standard to write like this in these types of files (in which case we can keep in both files).
If not standard - can we change the storage class file?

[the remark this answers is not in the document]
oriyarde: in general, IMO, we can change the storageclass, the choice is in our hands for both files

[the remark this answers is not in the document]
rivkap: ok - so in English the "or" is  with /. let's change in both places. thanks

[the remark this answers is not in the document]
oriyarde: note that the docs should be updated accordingly

deploy/kubernetes/examples/demo-volumereplication.yaml
ArbelNathan: done

deploy/kubernetes/examples/demo-volumereplicationclass.yaml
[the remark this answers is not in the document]
ArbelNathan: done

deploy/kubernetes/examples/demo-volumereplication.yaml
oriyarde: @rivkap could I resolve this?

rivkap: LGTM. Thanks

deploy/kubernetes/examples/demo-volumereplicationclass.yaml
ArbelNathan: I agree but isnt too late for changing the topology terms now?

oriyarde: it's now or never

ArbelNathan: it will need an internal source code rename if there is any change in doc..

oriyarde: right, but we can GA in between

ArbelNathan: if we document without change source code nothing will work

oriyarde: ok, so except for where the code has `request.system_id`

oriyarde: the only question is - what should be the name of the topology config secret key?

oriyarde: management_id?

ArbelNathan: `topology system id`?

oriyarde: in the future, theoretically, it could be used for something like replication as well, so I would avoid `topology` if possible

ArbelNathan: any suggestions?

oriyarde: [that](https://github.com/IBM/ibm-block-csi-driver/pull/370#discussion_r706840287) is the best one I could think of

ArbelNathan: would `by_system_id` have any logic if we change to `management_id`?

oriyarde: I meant that we need to change the `PARAMETERS_BY_SYSTEM` value from `by_system_id` to `by_management_id` before 1.7.0 GA

ArbelNathan: ok, [done](https://github.com/IBM/ibm-block-csi-driver/pull/374)

oriyarde (approved):
LGTM

rivkap (approved):
LGTM

PR #377, opened 2021-09-21
CSI-3103 update csi addons .proto version

oriyarde:


controller/scripts/csi_pb2.sh
```diff
@@ -2,7 +2,7 @@
 set -x
 
 CSI_VERSION="v1.2.0"
-ADDONS_VERSION="v0.2.0"
+ADDONS_VERSION="v0.1.1"
```
ArbelNathan, line 5: looks like a downgrade.. why?

oriyarde: these are all guesses

ArbelNathan (approved):
LGTM

PR #386, opened 2021-10-13
Feature csi 3424 upgrade side cars pb2 catch up

ArbelNathan:


go.mod
```diff
@@ -3,11 +3,17 @@ module github.com/ibm/ibm-block-csi-driver
...
+	github.com/davecgh/go-spew v1.1.1
+	github.com/evanphx/json-patch v4.2.0+incompatible
 	github.com/golang/mock v1.3.1
+	github.com/jessevdk/go-flags v1.5.0
+	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd
+	github.com/pkg/errors v0.9.1
 	github.com/sirupsen/logrus v1.4.2
 	golang.org/x/sync v0.0.0-20190423024810-112230192c58
 	google.golang.org/grpc v1.27.1
+	gopkg.in/inf.v0 v0.9.1
```
oriyarde, line 16: what commands did you use here? I don't get all this diff when upgrading the first dependency here

ArbelNathan: only `go mod tidy`

oriyarde: strange

ArbelNathan: I changed it to the results from the other file

PR #388, opened 2021-10-17
fix expired certificate for community tests

tyichye19:


Dockerfile-csi-test
```diff
@@ -14,6 +14,9 @@
 
 FROM golang:1.11
```
oriyarde, line 15: would bumping the base image version solve whatever is the issue?

oriyarde (approved):
LGTM

PR #395, opened 2021-11-02
1.8.0 v updates

rivkap:


docs/content/release_notes/csi_rn_changelog_1.8.0.md
```diff
@@ -0,0 +1,6 @@
+# 1.8.0 (December 2021)
+
+IBM® block storage CSI driver 1.8.0 adds new support and enhancements:
+- Now supports HyperSwap for IBM Spectrum Virtualize Family storage systems
+- New NVMe/FC support
```
oriyarde, line 5: why is hyperswap for svc but nvme isn't?
IMO the svc part here should either be for both or neither

rivkap: updated in PR 415

oriyarde: I [see](https://github.com/IBM/ibm-block-csi-driver/blob/17d919e94d523754830ebae450b82734ea9bfcec/docs/content/release_notes/changelog_1.8.0.md) - thanks!

PR #401, opened 2021-11-10
CSI-3491 update svc expand command

oriyarde:


controller/array_action/array_mediator_svc.py
```diff
@@ -317,10 +317,13 @@ def _get_object_fcmaps(self, object_name):
         all_fcmaps.extend(self._get_fcmaps_as_source_if_exist(object_name))
         return all_fcmaps
 
-    def _expand_cli_volume(self, cli_volume, increase_in_bytes):
+    def _expand_cli_volume(self, cli_volume, increase_in_bytes, is_in_remote_copy_relationship):
         volume_name = cli_volume.name
         try:
-            self.client.svctask.expandvdisksize(vdisk_id=volume_name, unit='b', size=increase_in_bytes)
+            if is_in_remote_copy_relationship:
+                self.client.svctask.expandvolume(object_id=volume_name, unit='b', size=increase_in_bytes)
```
ArbelNathan, line 324: as I understand, `expandvolume` is only for HyperSwap but `is_in_remote_copy_relationship` implies that its also for replication..

oriyarde: this is the existing check we use to delete a hyperswap volume.
I agree this existing function name might be misleading.
unlike hyperswap, replication (mirroring) does not use fcmaps, so no functional gap here

ArbelNathan: so PVC that is a part of a Replication object is not in a remote copy relationship?

ArbelNathan (approved):
LGTM

PR #405, opened 2021-11-24
Bug/csi 3586 expand nvme bad path

ArbelNathan:


node/pkg/driver/node_utils.go
```diff
@@ -177,6 +179,23 @@ func (n NodeUtils) StageInfoFileIsExist(filePath string) bool {
 	}
 	return true
 }
+
+func (n NodeUtils) DevicesAreNvme(sysDevices []string) bool {
+	args := []string{"list"}
+	out, err := n.Executer.ExecuteWithTimeout(TimeOutMultipathdCmd, nvmeCmd, args)
```
oriyarde, line 185: could we use a different timeout const here?
if you prefer, we can also have more general timeout consts, to use in specific timeout consts definition

node/pkg/driver/node_utils.go
```diff
@@ -177,6 +179,23 @@ func (n NodeUtils) StageInfoFileIsExist(filePath string) bool {
...
+	if err != nil {
+		return false
```
oriyarde, line 187: what if the command fails, but they are actually nvme devices?

node/pkg/driver/node_utils.go
```diff
@@ -177,6 +179,23 @@ func (n NodeUtils) StageInfoFileIsExist(filePath string) bool {
...
+	}
+	devices := string(out)
+	for _, deviceName := range sysDevices {
+		if strings.Contains(devices, deviceName) {
+			logger.Debugf("found device {%s} in nvme list", deviceName)
```
oriyarde, line 192:
```suggestion
	nvmeDevices := string(out)
	for _, deviceName := range sysDevices {
		if strings.Contains(nvmeDevices, deviceName) {
			logger.Debugf("found device {%s} in nvme list", deviceName)
```

node/pkg/driver/node_utils.go
```diff
@@ -177,6 +179,23 @@ func (n NodeUtils) StageInfoFileIsExist(filePath string) bool {
...
+
+func (n NodeUtils) DevicesAreNvme(sysDevices []string) bool {
+	args := []string{"list"}
+	out, err := n.Executer.ExecuteWithTimeout(TimeOutMultipathdCmd, nvmeCmd, args)
+	if err != nil {
+		return false
+	}
+	devices := string(out)
+	for _, deviceName := range sysDevices {
+		if strings.Contains(devices, deviceName) {
```
oriyarde, line 191: which of the "lists" is potentially longer?
I think we should check inside the one that is expected to be shorter

node/pkg/driver/node_utils.go
ArbelNathan: sure,
how do you suggest we call it? `TimeOutGeneralCmd`?

oriyarde: if we only have mainly 1 "default" value, then sure

node/pkg/driver/node_utils.go
ArbelNathan: sysDevices is way shorter

node/pkg/driver/node_utils.go
ArbelNathan: that the thing, the user dont have to have nvme packages on its env.. any suggestions?

node/pkg/driver/node_utils.go
ArbelNathan: done

node/pkg/driver/node_utils.go
oriyarde: maybe check if the command exists before?

node/pkg/driver/node_utils.go
oriyarde: I didn't mean to replace existing specific timeout consts.
i meant we could have a more general const in addition.
e.g. (assuming you meant to have this value for both commands):
```
TimeOutGeneralCmd           = 10 * 1000
TimeOutMultipathdCmd        = TimeOutGeneralCmd
TimeOutNvmeCmd              = TimeOutGeneralCmd
```

node/pkg/driver/node_utils.go
oriyarde: then I think it is more suitable for `Contains`

oriyarde: then again, the first sys device should be contained in nvme devices, so I'm not sure why we want to potentially go over all the volume sys devices

oriyarde: should we skip the rescan if **any** of them is nvme, or **all**?

ArbelNathan: it should find it on the first check, (if it checks nvmeDevices Contains sysDevice). no reason for going over all  the sysDevices

oriyarde: so do we need a for loop?
and if not, should the function even require a **list** of sys devices?

oriyarde: maybe if we have some sort of hybrid-connectivity configuration per volume, we should go over all sys devices before deciding to entirely skip the rescan?

oriyarde: I'm not sure which of the scenarios is even possible, hope you're right, resolving

node/pkg/driver/node_utils.go
ArbelNathan: done

node/pkg/driver/node_utils.go
ArbelNathan: done

node/pkg/driver/node_utils.go
```diff
@@ -177,6 +181,26 @@ func (n NodeUtils) StageInfoFileIsExist(filePath string) bool {
 	}
 	return true
 }
+
+func (n NodeUtils) DevicesAreNvme(sysDevices []string) (bool, error) {
+	args := []string{"list"}
+	if err := n.Executer.IsExecutable(nvmeCmd); err != nil {
+		return false, nil
```
oriyarde, line 188:
```suggestion
		return false, err
```

oriyarde (approved):
LGTM

PR #406, opened 2021-11-24
fix/CSI 3571 node id length

ArbelNathan:


node/pkg/driver/node_utils_test.go
```diff
@@ -329,33 +329,33 @@ func TestGenerateNodeID(t *testing.T) {
...
-			expNodeId: fmt.Sprintf("%s;;%s;%s", hostName, strings.Join(fcWWNs, ":"), iscsiIQN),
+			expNodeId: fmt.Sprintf("%s;;%s;%s", hostName, strings.Join(fcWWNs[:2], ":"), iscsiIQN),
 		},
 		{name: "fail long hostName on nvme port",
 			hostName: longHostName,
 			nvmeNQN:  nvmeNQN,
 			fcWWNs:   []string{},
 			iscsiIQN: "",
-			expErr:   errors.New(fmt.Sprintf("could not fit any ports in node id: %s;, length limit: 192", longHostName)),
+			expErr:   errors.New(fmt.Sprintf("could not fit any ports in node id: %s;, length limit: 128", longHostName)),
```
oriyarde, line 344: while you're at it, could we use the tested const here, instead of maintaining it separately?

ArbelNathan: done

oriyarde: should we import it, or maintain it separately?
IMO the const value is not "logic", and the tests should check logic, so it's ok to import it

ArbelNathan: done

oriyarde (approved):
LGTM

PR #407, opened 2021-11-25
Docs release 1.8.0 csi 3573 external log collection

rivkap:
Review for CSI-3573. There are some examples missing and some of the sections may be duplicates at this point.
Feel free to also suggest a "better" order for the section


docs/content/troubleshooting/log_status_collect.md
```diff
@@ -1,41 +1,108 @@
...
+
+    - If the PVCs are not bound, (in the _XXX_ state) collect the events of all unbound PVCs. (See [Log collection for unbound PVCs](#log-collection-for-unbound-pvcs).)
 
-## Status collection for CSI pods, daemonset, and statefulset
+### Log collection for all pods and containers
+To collect logs for all pods and containers, use the following commands:
 
+    nodepods=`kubectl get -n <namespace> pod -l app.kubernetes.io/component=csi-node --output=jsonpath={.items..metadata.name}`
+    
+    for pod in $nodepods; do  kubectl logs -n <namespace> --all-containers $pod > logs/$pod; done
```
oriyarde, line 56: I don't think we should encourage users to put the logs from all the pod's containers into a single file

docs/content/troubleshooting/log_status_collect.md
```diff
@@ -1,41 +1,108 @@
...
+
+### Log collection for all operator logs
+To collect all Operator logs, use the following commands:
+
+    operatorpod=`kubectl get pods --all-namespaces |grep ibm-block-csi-operator|awk '{print $2}'`
+    kubectl logs $operatorpod -n <namespace> > logs/operator
+
+### Log collection for all CSI component details
```
roysahar, line 64: I am not sure this section name should be "log collection", maybe use "Details of all CSI objects/components"
Also we should remember that when using "-l csi" if there is other csi installation (different vendor) on the same cluster with the label "csi" it will output also them (if using all namespaces).

oriyarde: I agree, we should use a different label than the general "csi"

docs/content/troubleshooting/log_status_collect.md
rivkap: so you want us to just not put this one in at all?

docs/content/troubleshooting/log_status_collect.md
rivkap: what should be written instead for `-l csi` ?
I can rename this section to : Collecting details of all CSI objects and components"
@oriyarde  @roysahar

docs/content/troubleshooting/log_status_collect.md
rivkap: @oriyarde  @roysahar  please confirm

docs/content/troubleshooting/log_status_collect.md
roysahar: @rivkap Regarding the label to use - "-l product=ibm-block-csi-driver"
As for the section name "Collecting details of all CSI objects and components" fine with me.

docs/content/troubleshooting/log_status_collect.md
roysahar: @rivkap I would write:
for pod in $nodepods;do for container in `kubectl get -n default pod $pod -o jsonpath='{.spec.containers[*].name}'`;do kubectl logs -n default $pod -c $container > /tmp/logs_collection/${pod}_${container}.log;done;done

rivkap: @roysahar  updated according to our meeting and your input
DONE - please approve

docs/content/troubleshooting/log_status_collect.md
rivkap: @roysahar  updated according to our meeting and your input
DONE - please approve

roysahar (approved):
Done.

docs/content/troubleshooting/log_status_collect.md
```diff
@@ -1,41 +1,121 @@
...
 
-## Status collection for CSI pods, daemonset, and statefulset
 
-`kubectl get all -n <namespace>  -l csi`
+### Log collection for all CSI controller containers
 
-### Log collection for IBM block storage CSI driver controller
+To collect logs for all pods and containers, use the following commands:
+    
+    for container in kubectl get -n default pod ibm-block-csi-controller-0 -o jsonpath='{.spec.containers[*].name}';do kubectl logs -n default ibm-block-csi-controller-0 -c $container > /tmp/logs_collection/ibm-block-csi-controller-0_${container}.log;done
```
oriyarde, line 64: 1. why is it sometimes `<namespace>` and sometimes `default`?
2. sometimes `> logs/` and sometimes `/tmp/logs_collection/`?
3. is it important for this to be a one-liner? if not, we could reduce its length by predefining e.g. `podname = ibm-block-csi-controller-0`, perhaps even making it the same for both node and controller, thus reducing duplication as well

docs/content/troubleshooting/log_status_collect.md
```diff
@@ -1,41 +1,121 @@
...
+
+    - If the PVCs are not in the _Bound_ state collect the events of all unbound PVCs. (See [Log collection for unbound PVCs](#log-collection-for-unbound-pvcs).)
+
+### Log collection for all CSI nodepods and their containers
+
+To collect logs for all nodepods, use the following commands:
+
+    nodepods=`kubectl get -n <namespace> pod -l app.kubernetes.io/component=csi-node --output=jsonpath={.items..metadata.name}`
+    
+    for pod in $nodepods; kubectl logs -n <namespace> $pod -c $container > /tmp/logs_collection/${pod}_${container}.log;done;done
```
oriyarde, line 57: where did this code come from?
I think it's missing something like `for container in kubectl get -n default pod $pod -o jsonpath='{.spec.containers[*].name}';`
@roysahar could you write here the ticket number to test the command changes? I think it should be done before merging each of these PRs.
could we combine this code with the controller log collection into one section/script? most of the code is the same

docs/content/troubleshooting/log_status_collect.md
```diff
@@ -1,41 +1,121 @@
...
+
+### General status and log collection for IBM Support
+Be sure to run the following steps and copy the output to an external file, when engaging IBM Support and sending log collections.
+
+1. Check the node status.
+    
+    `kubectl get nodes`
+2. Check the CSI driver component status.
+
+    `kubectl get -n <namespace> pod -o wide | grep ibm-block-csi`
```
oriyarde, line 44: why is it sometimes `<namespace>` and sometimes `--all-namespaces`?
IMO `--all-namespaces` makes us less blind and easier for the user to run

docs/content/troubleshooting/log_status_collect.md
```diff
@@ -1,41 +1,121 @@
...
+    kubectl logs $operatorpod -n <namespace> > logs/operator
+
+
+### Collecting details of all CSI objects and components
+
+`kubectl describe all -l product=ibm-block-csi-driver -n <namespace>`
+
+For example:
+
+    kubectl describe all -l csi -n <namespace> > logs/describe_ibm-block-csi-driver
```
oriyarde, line 80: I thought [we decided in this PR](https://github.com/IBM/ibm-block-csi-driver/pull/407#discussion_r764824848) not to use `-l csi`

docs/content/troubleshooting/log_status_collect.md
```diff
@@ -1,41 +1,121 @@
...
+
+
+### Status collection for CSI pods, daemonset, and statefulset
+`kubectl get all -n <namespace>  -l product=ibm-block-csi-driver`
+
+For example:
+
+    kubectl get all -n <namespace> -l csi > get_all_ibm-block-csi-driver
```
oriyarde, line 88: this does not look to me like an example for this command.
also why should we maintain both the command and its example?

docs/content/troubleshooting/log_status_collect.md
```diff
@@ -1,41 +1,121 @@
...
+### Log collection for pods not in the _Running_ state
+`kubectl describe pod <not-running-pod-name>`
 
-    ```
-    kubectl describe -n <namespace> pod/<pod-name>
-    ```
+For example:
 
-    View the logs.
+    kubectl describe pod <pod-name> > logs/pod_not_running
```
oriyarde, line 121: I don't see here anything specific to our csi, nor any csi.
I don't think we should maintain this

docs/content/troubleshooting/log_status_collect.md
```diff
@@ -1,41 +1,121 @@
...
+
+1. Check the node status.
+    
+    `kubectl get nodes`
+2. Check the CSI driver component status.
+
+    `kubectl get -n <namespace> pod -o wide | grep ibm-block-csi`
+3. Check if the PersistentVolumeClaims (PVCs) are _Bound_.
+
+    `kubectl get pvc`
```
oriyarde, line 47: @roysahar do we want to include non-ibm-block-csi PVCs here?

docs/content/troubleshooting/log_status_collect.md
oriyarde: I think we need [test confirmation](https://github.com/IBM/ibm-block-csi-driver/pull/407#discussion_r766512218) (specifically here, and in general), before approving

docs/content/troubleshooting/log_status_collect.md
oriyarde: I will let @roysahar approve, since he opened this conversation

docs/content/troubleshooting/log_status_collect.md
rivkap: @oriyarde 
I updated this file together with @roysahar . who else needs to test this section? In any case, testing can be done from the finalized docs_release-1.8.0 branch

docs/content/troubleshooting/log_status_collect.md
rivkap: if it's not log collection would it be a status collection?
In any case - it was requested by Support and is used by them in cases.  While it may not be csi specific there is value in keeping it here.
@roysahar  please also comment here as we reviewed this together as well.

docs/content/troubleshooting/log_status_collect.md
oriyarde: I don't think @roysahar had a "test" ticket. I meant that he should have one in addition to his review.
I agree testing can be done from the finalized branch, but IMO it would be better to split the effort into separate PRs, separate test tickets, etc...

docs/content/troubleshooting/log_status_collect.md
rivkap: The only place it is --all-namespaces is for 

> ### Log collection for CSI operator logs
> To collect CSI operator logs, use the following commands:
> 
>     operatorpod=`kubectl get pods --all-namespaces |grep ibm-block-csi-operator|awk '{print $2}'`
>     kubectl logs $operatorpod -n <namespace> > logs/operator

@roysahar  can this be updated to <namespace>?

docs/content/troubleshooting/log_status_collect.md
oriyarde: unlike the log, the status is one of the (many) fields in the `describe` command output, so it's better, but still [not accurate](https://kubernetes.io/docs/tasks/debug-application-cluster/debug-application-introspection/#using-kubectl-describe-pod-to-fetch-details-about-pods).
I'm not sure why Support wants this to be specifically in the **our CSI** docs.
if they are not looking for a complete procedure to follow in all cases, then IMO a link to [this](https://kubernetes.io/docs/reference/kubectl/cheatsheet/#viewing-finding-resources) should be more than enough

docs/content/troubleshooting/log_status_collect.md
rivkap: FYI opened [PR 418](https://github.com/IBM/ibm-block-csi-driver/pull/418) for updates

docs/content/troubleshooting/log_status_collect.md
rivkap: 1. it was a mistake on my part. I was supposed to change it. it is now changed in the new PR
2. updated
3. it is not important for it to be a one liner but node and controller are separate and so they can be separate commands. It's what I discussed with @roysahar . Roy - if you want to update it please let me know.
all changes can be found in  [PR 418](https://github.com/IBM/ibm-block-csi-driver/pull/418) for updates

docs/content/troubleshooting/log_status_collect.md
rivkap: FYI opened [PR 418](https://github.com/IBM/ibm-block-csi-driver/pull/418) for updates

docs/content/troubleshooting/log_status_collect.md
rivkap: @roysahar ? can this be changed?

docs/content/troubleshooting/log_status_collect.md
rivkap: i'm not sure how these got missed. i thought that i had updated them all. in any case - updates are now found in [PR 418](https://github.com/IBM/ibm-block-csi-driver/pull/418) for updates

docs/content/troubleshooting/log_status_collect.md
oriyarde: thanks!

docs/content/troubleshooting/log_status_collect.md
oriyarde: thanks!

docs/content/troubleshooting/log_status_collect.md
oriyarde: thanks!

docs/content/troubleshooting/log_status_collect.md
oriyarde: thanks!

docs/content/troubleshooting/log_status_collect.md
rivkap: it doesn't look like the command because the -l csi should have been -l product=ibm-block-csi-driver. it's updated in the example now as well.
1. it's not uncommon to maintain both a command and an example of implementation.
2. this way we have the command along with an example of where they may be storing the logs. we give them the best practice / option to store it in a logs-type file.

documentation-wise this is not incorrect.
If you want, we can make a note within the section saying that all commands are using the `logs` folder as an example and _only_ keep a single command together with the `>logs/`  for each of them.

oriyarde: > only keep a single command together with the >logs/ for each of them
sounds good to me

docs/content/troubleshooting/log_status_collect.md
rivkap: updated in other place with grep command

rivkap: done

docs/content/troubleshooting/log_status_collect.md
rivkap: completed per our meeting

PR #417, opened 2021-12-08
Bug/csi 3652 fc invalid lun

ArbelNathan:


controller/array_action/array_mediator_svc.py
```diff
@@ -798,10 +800,11 @@ def _get_free_lun(self, host_name):
         # do not support more than 255 or 511 mappings today irrespective of
         # our constraint).
         lun_range_gen = range(self.MIN_LUN_NUMBER, self.MAX_LUN_NUMBER + 1)
-        lun_range = {str(lun) for lun in lun_range_gen}
-        free_luns = list(lun_range - set(luns_in_use))
+        lun_range = [str(lun) for lun in lun_range_gen]
+        free_luns = [lun for lun in lun_range if lun not in luns_in_use]
```
matancarmeli7, line 804:
```suggestion
        lun_range = [str(lun) for lun in range(self.MIN_LUN_NUMBER, self.MAX_LUN_NUMBER + 1) if lun not in luns_in_use
]
```
(this suggestion is also for [this](https://github.com/IBM/ibm-block-csi-driver/blob/e39363f0791c7961708d006fb143320344cca882/controller/array_action/array_mediator_svc.py#L802) line, I couldn't mark it with 803 and 804)
maybe we can save like these two similar for loops
(does `MIN_LUN_NUMBER` variable is necessary when his value is 0? or it's more for the future? )

ArbelNathan: you mean:
`lun_range = [str(lun) for lun in range(self.MIN_LUN_NUMBER, self.MAX_LUN_NUMBER + 1) if str(lun) not in luns_in_use]`

1.  its less readable - "lun in range()" have no literal meaning.
2. hard to maintain - even in your suggestion you missed one parsing.
3. less efficient - 2 parsing instead of 1.

matancarmeli7 (approved):
LGTM

erantzabari (approved):

PR #425, opened 2021-12-20
Bug/csi 3657 controller unpublish spec gap

ArbelNathan:


controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -1334,38 +1334,45 @@ def test_unpublish_volume_get_host_by_host_identifiers_exception(self, storage_a
...
+
+    @patch("controller.controller_server.csi_controller_server.get_agent")
+    def test_unpublish_volume_unmap_volume_host_not_found_error(self, storage_agent):
+        self._test_unpublish_volume_unmap_volume_with_error(storage_agent, array_errors.HostNotFoundError("host"),
+                                                            grpc.StatusCode.OK)
+
+    @patch("controller.controller_server.csi_controller_server.get_agent")
+    def test_unpublish_volume_unmap_volume_permission_denied_unmapping_error(self, storage_agent):
+        self._test_unpublish_volume_unmap_volume_with_error(storage_agent, array_errors.UnmappingError("", "", ""),
+                                                            grpc.StatusCode.INTERNAL)
```
oriyarde, line 1375: do we want the unit tests to enforce status codes that are not in the spec, for specific mediator exceptions?
if not at all, please remove this test.
if so, but not for specific mediator exceptions, please replace `UnmappingError` with e.g. `Exception`

ArbelNathan: its not a new test, just separated.. and we are testing the code, not only the spec..

oriyarde: ok, since it's not new - resolving

controller/controller_server/utils.py
```diff
@@ -491,7 +490,7 @@ def get_node_id_info(node_id):
     elif len(split_node) == 2:
         hostname, nvme_nqn = split_node
     else:
-        raise array_errors.HostNotFoundError(node_id)
+        raise ValidationException(messages.wrong_format_message.format("node id"))
```
oriyarde, line 493: I think this should be performed as part of a function like `validate_publish_volume_request`, instead of here

ArbelNathan: and what will be raised here?

oriyarde: `ValueError`/`return None`/whatever

oriyarde: `ValueError` seems more convenient for the caller than `None`

controller/controller_server/utils.py
```diff
@@ -428,6 +427,17 @@ def validate_delete_volume_request(request):
     logger.debug("delete volume validation finished")
 
 
+def _validate_node_id(node_id):
+    logger.debug("validating node id")
+
+    split_node = node_id.split(config.PARAMETERS_NODE_ID_DELIMITER)
+
+    if len(split_node) > 4 or len(split_node) < 2:
```
oriyarde, line 435:
```suggestion
    delimiter_count = node_id.count(config.PARAMETERS_NODE_ID_DELIMITER)

    if not 2 <= delimiter_count <= 4:
```

controller/controller_server/utils.py
```diff
@@ -428,6 +427,17 @@ def validate_delete_volume_request(request):
...
+        raise ValidationException(messages.wrong_format_message.format("node id"))
+
+    logger.debug("secrets validation finished")
```
oriyarde, line 438: ?

controller/tests/controller_server/utils_test.py
```diff
@@ -28,10 +27,33 @@ def setUp(self):
...
+        self.util_method = ""
+
+    def _test_validation_exception(self, method_arg, msg="", raised_error=ValidationException):
+        with self.assertRaises(raised_error) as ex:
+            getattr(utils, self.util_method)(method_arg)
+        if msg:
+            self.assertIn(msg, str(ex.exception))
+
+    def _test_validate_node_id_validation_exception(self, node_id):
+        self.util_method = "_validate_node_id"
```
oriyarde, line 39: I don't think we should test private functions.
and I think it would be simpler to pass the function itself, instead of its name

controller/tests/controller_server/utils_test.py
```diff
@@ -392,9 +402,9 @@ def _check_node_id_parameters(self, node_id_info, nvme_nqn, fc_wwns, iscsi_iqn):
         self.assertEqual(node_id_info.initiators.iscsi_iqn, iscsi_iqn)
 
     def test_get_node_id_info(self):
-        with self.assertRaises(array_errors.HostNotFoundError) as ex:
+        with self.assertRaises(ValueError) as ex:
             utils.get_node_id_info("bad-node-format")
-            self.assertTrue("node" in str(ex))
+        self.assertIn("node", str(ex.exception))
```
oriyarde, line 407: does it fail without `.exception`?
if not, then I would remove it

ArbelNathan: yes

controller/tests/controller_server/utils_test.py
ArbelNathan: its just like validate_secret, I just dont understand why all the specific validation functions are private..

controller/controller_server/utils.py
ArbelNathan: `1 <= delimiter_count <= 3` - delimiter is one less than parts..
done

controller/controller_server/utils.py
ArbelNathan: oops
done

controller/tests/controller_server/utils_test.py
ArbelNathan: and done with the "function itself"

oriyarde: they are private because they are not called directly from outside the file

oriyarde: then why is `self.util_method` [still initialized as a string](https://github.com/IBM/ibm-block-csi-driver/blob/beb4cc25b7fd99c888e8ba2656ebf28d71facd6f/controller/tests/controller_server/utils_test.py#L30)?
I don't think we need it in `self`

ArbelNathan: I dont want to send it for every call..

oriyarde: you set it up before every call, so it doesn't really add code.
and it's hard to see which method is being tested for each call (especially when it is set before other "preparations")

controller/tests/controller_server/utils_test.py
```diff
@@ -28,10 +27,33 @@ def setUp(self):
                                       "publish_context_array_iqn": "array_iqn",
                                       "publish_context_fc_initiators": "fc_wwns"}
                        }
+        self.util_method = None
+
+    def _test_validation_exception(self, method_arg, msg="", raised_error=ValidationException):
+        with self.assertRaises(raised_error) as ex:
+            self.util_method(method_arg)
```
oriyarde, line 34: I think they are regular functions, not methods

controller/tests/controller_server/utils_test.py
oriyarde: it would also eliminate the need for `_test_validate_node_id_validation_exception`

controller/tests/controller_server/utils_test.py
ArbelNathan: done

controller/tests/controller_server/utils_test.py
ArbelNathan: done

controller/tests/controller_server/utils_test.py
ArbelNathan: done

oriyarde: the `util_function` is still being [set up before the call](https://github.com/IBM/ibm-block-csi-driver/blob/26f5c8277053a94c8d9ba26e7fb46a562302cc9b/controller/tests/controller_server/utils_test.py#L150), so it's still hard to see which method is being tested for each call (the argument name should be more specific than the generic parameter name).

if you don't want to function names in one line (e.g. `self._test_validation_exception(utils.validate_csi_volume_capabilities, [])`), then consider wrapping it (e.g. `self._test_validate_csi_volume_capabilities_validation_exception([])`). same for all `self._test_validation_exception` calls

controller/tests/controller_server/utils_test.py
oriyarde: I don't see any change in the line about which this conversation is

controller/tests/controller_server/utils_test.py
ArbelNathan: I [wrapped it](https://github.com/IBM/ibm-block-csi-driver/pull/425#discussion_r773858554), you asked to remove it..

ArbelNathan: done

controller/tests/controller_server/utils_test.py
ArbelNathan: done

controller/tests/controller_server/utils_test.py
```diff
@@ -29,48 +28,68 @@ def setUp(self):
                                       "publish_context_fc_initiators": "fc_wwns"}
                        }
 
-    def _test_validate_secrets_validation_errors(self, secrets):
-        with self.assertRaises(ValidationException):
-            utils.validate_secrets(secrets)
+    def _test_validation_exception(self, util_function, method_arg, msg="", raised_error=ValidationException):
+        with self.assertRaises(raised_error) as context:
+            util_function(method_arg)
```
oriyarde, line 33: it's still not a "method"

controller/tests/controller_server/utils_test.py
```diff
@@ -29,48 +28,68 @@ def setUp(self):
                                       "publish_context_fc_initiators": "fc_wwns"}
                        }
 
-    def _test_validate_secrets_validation_errors(self, secrets):
-        with self.assertRaises(ValidationException):
-            utils.validate_secrets(secrets)
+    def _test_validation_exception(self, util_function, method_arg, msg="", raised_error=ValidationException):
```
oriyarde, line 31: `msg` is not the "full message". perhaps `msg_substr`? `str_in_msg`?

controller/tests/controller_server/utils_test.py
oriyarde: I didn't ask to remove, I meant that two function names in one line could eliminate the wrapping need, then I saw a "done" without two function names in one line, so back to wrapping

controller/tests/controller_server/utils_test.py
```diff
@@ -392,9 +409,7 @@ def _check_node_id_parameters(self, node_id_info, nvme_nqn, fc_wwns, iscsi_iqn):
         self.assertEqual(node_id_info.initiators.iscsi_iqn, iscsi_iqn)
 
     def test_get_node_id_info(self):
-        with self.assertRaises(array_errors.HostNotFoundError) as ex:
-            utils.get_node_id_info("bad-node-format")
-            self.assertTrue("node" in str(ex))
+        self._test_validation_exception(utils.get_node_id_info, "bad-node-format", msg="node", raised_error=ValueError)
```
oriyarde, line 412:
```suggestion
        self._test_validation_exception(utils.get_node_id_info, "bad-node-id-format", msg="node", raised_error=ValueError)
```

controller/tests/controller_server/utils_test.py
```diff
@@ -375,15 +391,16 @@ def test_get_volume_id_info_with_internal_id(self):
...
-            utils.get_volume_id_info("xiv:0;volume;id")
-        self.assertIn("Wrong volume id format", str(ex.exception))
+        self._test_get_volume_id_info_validation_exception("xiv:0;volume;id", msg="Wrong volume id format",
+                                                           raised_error=ObjectIdError)
 
     def test_get_volume_id_info_no_id_fail(self):
-        with self.assertRaises(ObjectIdError) as ex:
-            utils.get_volume_id_info("badvolumeformat")
-        self.assertIn("Wrong volume id format", str(ex.exception))
+        self._test_get_volume_id_info_validation_exception("badvolumeformat", msg="Wrong volume id format",
```
oriyarde, line 402:
```suggestion
        self._test_get_volume_id_info_validation_exception("badformatvolumeid", msg="Wrong volume id format",
```

oriyarde (approved):
LGTM

PR #429, opened 2021-12-26
CSI-3467 fix svc warnings handling

oriyarde:


controller/array_action/array_mediator_svc.py
```diff
@@ -435,12 +436,10 @@ def _create_cli_volume(self, name, size_in_bytes, space_efficiency, pool):
             cli_kwargs = build_kwargs_from_parameters(space_efficiency, pool,
                                                       name, size)
             self.client.svctask.mkvolume(**cli_kwargs)
-            cli_volume = self._get_cli_volume(name)
-            logger.info("finished creating cli volume : {}".format(cli_volume.name))
-            return cli_volume
         except (svc_errors.CommandExecutionError, CLIFailureError) as ex:
+            logger.warning("exception encountered during creation of volume {0}: {1}".format(name,
+                                                                                             ex.my_message))
```
ArbelNathan, line 441: why do we need those logs? we know in which method we are and the err will be logged anyway

oriyarde: the warning would not be logged anyway

ArbelNathan: so can we log it only in a warning case? this way we would have double logs

oriyarde: so we won't have log messages with "error" level, even when it is more than a warning?

oriyarde: I removed the double logging now. what do you think about this way?

ArbelNathan: yes, it looks better IMO.

ArbelNathan (approved):
LGTM

PR #307, opened 2021-04-08
CSI-1516 more mediator alignment and renaming

oriyarde:
mainly:
`controller_errors` (where they are *not* errors raised by controller server) -> `array_errors`
`vol` -> `volume`
`snap` -> `snapshot`

PR #431, opened 2022-01-02
CSI-2896 more controller renaming

oriyarde:
continuation of https://github.com/IBM/ibm-block-csi-driver/pull/307


controller/controller_server/csi_controller_server.py
```diff
@@ -104,21 +104,22 @@ def CreateVolume(self, request, context):
                         return build_error_response(message, context, grpc.StatusCode.ALREADY_EXISTS,
                                                     csi_pb2.CreateVolumeResponse)
 
-                    copy_source_res = self._handle_existing_volume_source(volume, source_id, source_type,
-                                                                          array_connection_info.system_id,
-                                                                          context)
-                    if copy_source_res:
-                        return copy_source_res
+                    response = self._handle_existing_volume_source(volume, source_id, source_type,
```
ArbelNathan, line 107: what does `"response"` means?

ArbelNathan: I understand `"response"` from spec method, not else..

oriyarde: what else could it be?

ArbelNathan: could it be more specific? like `create_volume_response`?
if it's not a `spec` method (or something like `utils.generate_csi_create_volume_response` which is clear from the name what the response is.) `response` could be anything..

oriyarde: would it be clear enough if instead I rename `_handle_existing_volume_source` to something like `_get_response_for_existing_volume_source`?
I'm still not sure what `response` could it be besides `spec`

ArbelNathan: the `response` is a `grpc` object that of "type" `CreateVolumeResponse`, a `response` could be of any kind and the method name is not telling which one. that why I suggested `create_volume_response` for the name here, every other place indicates the `response` type by the function name. I don't think that `_get_response_for_existing_volume_source` is better for that reason.

oriyarde: >a response could be of any kind

e.g.?

>the method name is not telling which one

this code is in a method called `CreateVolume`


how about `_get_create_volume_response_for_existing_volume_source`?

ArbelNathan: > > a response could be of any kind
> 
> e.g.?
`OK`
why? IDK, but why would it be `CreateVolumeResponse`? where is it saying that it is? it's in create volume but it does not create volume..

`_get_create_volume_response_for_existing_volume_source` is ok but long.. if you don't care so Im ok

ArbelNathan (approved):
LGTM

PR #433, opened 2022-01-04
Reduce log fillings

ArbelNathan:


node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
```diff
@@ -195,7 +196,8 @@ func (r OsDeviceConnectivityHelperScsiGeneric) GetMpathDevice(volumeId string) (
 
 func (r OsDeviceConnectivityHelperScsiGeneric) flushDeviceBuffers(deviceName string) error {
 	devicePath := filepath.Join(DevPath, deviceName)
-	_, err := r.Executer.ExecuteWithTimeout(TimeOutBlockDevCmd, blockDevCmd, []string{"--flushbufs", devicePath})
+	logger.Debugf("executing command : {--flushbufs} on device : {%v}.", deviceName)
```
oriyarde, line 199: it's not clear from the message that the command in use is `blockdev`, nor its timeout.
is this to avoid the empty `Output from command: ` message? if so, we could just avoid it somehow inside `ExecuteWithTimeout` when the output is empty

ArbelNathan: no, its to avoid putting timeout and anything that is not interesting

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
```diff
@@ -228,7 +230,7 @@ func (r OsDeviceConnectivityHelperScsiGeneric) FlushMultipathDevice(mpathDevice
 	logger.Debugf("Try to acquire lock for running the command multipath -f {%v} (to avoid concurrent multipath commands)", mpathDevice)
 	r.MutexMultipathF.Lock()
 	logger.Debugf("Acquired lock for multipath -f command")
-	_, err = r.Executer.ExecuteWithTimeout(TimeOutMultipathCmd, multipathCmd, []string{"-f", fullDevice})
+	_, err = r.Executer.ExecuteWithTimeout(TimeOutMultipathCmd, multipathCmd, []string{"-f", fullDevice}, false)
```
oriyarde, line 233: this is how it looks:
```
... (device_connectivity_helper_scsigeneric.go:232) - Acquired lock for multipath -f command
... (executer.go:75) - Finished executing command
```
I think the log should explicitly state the full command that is about to be used, right before it is executed. ideally without maintaining a duplication of all its details (e.g. parameters) in a separate message.

i.e. I would remove the `false` here

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
ArbelNathan: why do we need the time out in the node?
plus, `ExecuteWithTimeout ` can be called in a for loop and the log gets bigger in every call so even if part of the log there contains what I added I still want it out of there and out of the loop altogether.

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
```diff
@@ -251,7 +253,7 @@ func (r OsDeviceConnectivityHelperScsiGeneric) RemovePhysicalDevice(sysDevices [
 	}
 
 	// sysDevices  = sdb, sda,...
-	logger.Debugf("Removing scsi device : {%v}", sysDevices)
+	logger.Debugf("Removing scsi device : {%v} by open the devices delete file : {%v}", sysDevices, fmt.Sprintf(deviceDeletePath, "<deviceName>"))
```
oriyarde, line 256: while you're at it
```suggestion
	logger.Debugf(`Removing scsi device : {%v} by writing "1" to the delete file of each device: {%v}`, sysDevices, fmt.Sprintf(deviceDeletePath, "<deviceName>"))
```

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
ArbelNathan: the full command is multipath -f

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
oriyarde: the timeout is useful info (e.g. when encountering long executions), same as `deviceDeletePath`, timestamps, etc...

I don't see a loop in `flushDeviceBuffers` nor in `ExecuteWithTimeout`, so the change in the line which this conversation is about, will not take the `executing command` message out of any loop

ArbelNathan: specifically, here the whole function was called in a loop so I didn't have a way to put the command out of it, and I don't think the calling function (that calls to `flushDeviceBuffers`) should know about the args of the command inside..

ArbelNathan: I don't understand your example for useful info,

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
oriyarde: this is how the full command looks like before this PR:
```
... (device_connectivity_helper_scsigeneric.go:230) - Acquired lock for multipath -f command
... (executer.go:46) - Executing command : {multipath} with args : {[-f /dev/dm-3]}. and timeout : {60000} mseconds
... (executer.go:67) - Output from command: 
... (executer.go:72) - Finished executing command
```
note the `dm-3` (/potentially other parameters in the future)

node/pkg/driver/executer/executer.go
```diff
@@ -42,9 +42,10 @@ type ExecuterInterface interface { // basic host dependent functions
 type Executer struct {
 }
 
-func (e *Executer) ExecuteWithTimeout(mSeconds int, command string, args []string) ([]byte, error) {
-	logger.Debugf("Executing command : {%v} with args : {%v}. and timeout : {%v} mseconds", command, args, mSeconds)
-
+func (e *Executer) ExecuteWithTimeout(mSeconds int, command string, args []string, logRequired bool) ([]byte, error) {
```
oriyarde, line 45:
```suggestion
func (e *Executer) ExecuteWithTimeout(mSeconds int, command string, args []string, isLogRequired bool) ([]byte, error) {
```

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
oriyarde: e.g. if any of our timeouts is not quite calibrated, and the relevant timeout value is included with the messages around the time of the issue, it would be easier to see spot it

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
```diff
@@ -429,7 +430,7 @@ func (o OsDeviceConnectivityHelperGeneric) GetWwnByScsiInq(dev string) (string,
 	args := []string{"-p", "0x83", dev}
 	// add timeout in case the call never comes back.
 	logger.Debugf("Calling [%s] with timeout", sgInqCmd)
-	outputBytes, err := o.Executer.ExecuteWithTimeout(TimeOutSgInqCmd, sgInqCmd, args)
+	outputBytes, err := o.Executer.ExecuteWithTimeout(TimeOutSgInqCmd, sgInqCmd, args, true)
```
oriyarde, line 433: could we have a variable for the boolean in each caller?

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
```diff
@@ -560,9 +561,10 @@ func (o GetDmsPathHelperGeneric) WaitForDmToExist(volumeUuid string, volumeNguid
 	formatTemplate := strings.Join([]string{"%d", "%w"}, mpathdSeparator)
 	args := []string{"show", "maps", "raw", "format", "\"", formatTemplate, "\""}
 	var err error
+	logger.Debugf("Waiting for dm to exist, executing command : {%v} with args : {%v}.", multipathdCmd, args)
 	for i := 0; i < maxRetries; i++ {
 		err = nil
-		out, err := o.executer.ExecuteWithTimeout(TimeOutMultipathdCmd, multipathdCmd, args)
+		out, err := o.executer.ExecuteWithTimeout(TimeOutMultipathdCmd, multipathdCmd, args, false)
```
oriyarde, line 567: from past experience in debugging customer logs, the output of this command is important to understand which DMs exist at each time

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
oriyarde: since we usually want `true`, consider adding wrapper functions (e.g. `ExecuteWithTimeoutSilently`), so most callers won't have to deal with this

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
```diff
@@ -72,6 +72,7 @@ const (
 	WaitForMpathWaitIntervalSec = 1
 	FcHostSysfsPath             = "/sys/class/fc_remote_ports/rport-*/port_name"
 	IscsiHostRexExPath          = "/sys/class/iscsi_host/host*/device/session*/iscsi_session/session*/targetname"
+	deviceDeletePath            = "/sys/block/%s/device/delete"
```
oriyarde, line 75:
```suggestion
	deviceDeletePathFormat      = "/sys/block/%s/device/delete"
```

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
ArbelNathan: I added timeout

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
ArbelNathan: I removed it from the Silent group and conditioned the logging of the output in case it's empty.l

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
ArbelNathan: didnt understand the first comment
done the second

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
```diff
@@ -205,12 +206,14 @@ func (r OsDeviceConnectivityHelperScsiGeneric) flushDeviceBuffers(deviceName str
...
 
 func (r OsDeviceConnectivityHelperScsiGeneric) flushDevicesBuffers(deviceNames []string) error {
+	logger.Debugf("executing commands : {%v --flushbufs} on devices : {%v}.", blockDevCmd, deviceNames)
 	for _, deviceName := range deviceNames {
 		err := r.flushDeviceBuffers(deviceName)
 		if err != nil {
 			return err
 		}
 	}
+	logger.Debugf("Finished executing commands")
```
oriyarde, line 216:
```suggestion
	logger.Debugf("Finished executing commands: {%v --flushbufs}", blockDevCmd)
```

node/pkg/driver/executer/executer.go
```diff
@@ -64,10 +64,21 @@ func (e *Executer) ExecuteWithTimeout(mSeconds int, command string, args []strin
...
+	logger.Debugf("Executing command : {%v} with args : {%v}. and timeout : {%v} mseconds", command, args, mSeconds)
+
+	out, err := e.ExecuteWithTimeoutSilently(mSeconds, command, args)
+
+	outAsStr := string(out)
+	if strings.TrimSpace(outAsStr) != "" {
+		logger.Debugf("Output from command: %s", outAsStr)
+	}
 
 	logger.Debugf("Finished executing command")
```
oriyarde, line 83:
```suggestion
	outAsStr := string(out)
	if strings.TrimSpace(outAsStr) != "" {
		logger.Debugf("Output from command: %s", outAsStr)
		noOutputMessage := ""
	}
	else {
		noOutputMessage := " (no output)"
	}

	logger.Debugf("Finished executing command" + noOutputMessage)
```

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
```diff
@@ -560,6 +562,7 @@ func (o GetDmsPathHelperGeneric) WaitForDmToExist(volumeUuid string, volumeNguid
 	formatTemplate := strings.Join([]string{"%d", "%w"}, mpathdSeparator)
 	args := []string{"show", "maps", "raw", "format", "\"", formatTemplate, "\""}
 	var err error
+	logger.Debugf("Waiting for dm to exist, executing command : {%v} with args : {%v}.", multipathdCmd, args)
 	for i := 0; i < maxRetries; i++ {
 		err = nil
 		out, err := o.executer.ExecuteWithTimeout(TimeOutMultipathdCmd, multipathdCmd, args)
```
oriyarde, line 568: wouldn't the `executing command` part be logged twice, even for success on first attempt?

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
oriyarde: I don't see the added timeout - link to line?

node/pkg/driver/executer/executer.go
ArbelNathan: the all point is to reduce the log, if we are not expecting any return so why log it?

oriyarde: what do you mean "not expecting"?
this just explicitly clarifies in the log that there was no output.
I don't think this significantly hampers the log reduction

ArbelNathan: when we are talking about logs that are thousands of lines long I think that every line counts.

oriyarde: this suggestion doesn't add more lines

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
ArbelNathan: right, done

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
ArbelNathan: done

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
ArbelNathan: done

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
ArbelNathan: didnt commit, now its in

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
```diff
@@ -560,6 +562,7 @@ func (o GetDmsPathHelperGeneric) WaitForDmToExist(volumeUuid string, volumeNguid
 	formatTemplate := strings.Join([]string{"%d", "%w"}, mpathdSeparator)
 	args := []string{"show", "maps", "raw", "format", "\"", formatTemplate, "\""}
 	var err error
+	logger.Debugf("Waiting for dm to exist.")
```
oriyarde, line 565: AFAIK log messages usually don't end with a period

node/pkg/driver/executer/executer.go
ArbelNathan: its adding to my change in case of empty respond from command

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
```diff
@@ -205,12 +206,14 @@ func (r OsDeviceConnectivityHelperScsiGeneric) flushDeviceBuffers(deviceName str
 }
 
 func (r OsDeviceConnectivityHelperScsiGeneric) flushDevicesBuffers(deviceNames []string) error {
+	logger.Debugf("executing commands : {%v --flushbufs} on devices : {%v} and timeout : {%v} mseconds.", blockDevCmd, deviceNames, TimeOutBlockDevCmd)
```
oriyarde, line 209: I think `--flushbufs` should now be a const too (`flushBufsFlag`?)

node/pkg/driver/executer/executer.go
oriyarde: yes, but as I wrote, the addition is insignificant (12 chars in an existing message, and only when there is no output), and makes the lack of output clear

ArbelNathan: done

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
ArbelNathan: done (removed period compeletly)

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
```diff
@@ -80,6 +81,7 @@ const (
 	VolumeStorageIdsDelimiter   = ";"
 	WwnOuiEnd                   = 7
 	WwnVendorIdentifierEnd      = 16
+	flushBufsFlag               = "--flushbufs"
```
oriyarde, line 84: I would put it right after `blockDevCmd`

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
```diff
@@ -195,22 +197,24 @@ func (r OsDeviceConnectivityHelperScsiGeneric) GetMpathDevice(volumeId string) (
 
 func (r OsDeviceConnectivityHelperScsiGeneric) flushDeviceBuffers(deviceName string) error {
 	devicePath := filepath.Join(DevPath, deviceName)
-	_, err := r.Executer.ExecuteWithTimeout(TimeOutBlockDevCmd, blockDevCmd, []string{"--flushbufs", devicePath})
+	_, err := r.Executer.ExecuteWithTimeoutSilently(TimeOutBlockDevCmd, blockDevCmd, []string{flushBufsFlag, devicePath})
 	if err != nil {
-		logger.Errorf("blockdev --flushbufs {%v} did not succeed to flush the device buffers. err={%v}", devicePath,
+		logger.Errorf("blockdev %v {%v} did not succeed to flush the device buffers. err={%v}", flushBufsFlag, devicePath,
```
oriyarde, line 202: `blockdev` can come from `blockDevCmd` here as well

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
ArbelNathan: done

ArbelNathan: done

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
ArbelNathan: done

oriyarde (approved):
LGTM

controller/array_action/array_mediator_ds8k.py
```diff
@@ -444,9 +444,8 @@ def unmap_volume(self, volume_id, host_name):
 
     def _get_api_volume_from_volumes(self, volume_candidates, volume_name):
         for volume in volume_candidates:
-            logger.info("Checking volume: {}".format(volume.name))
```
oriyarde, line 447: should we log the length of `volume_candidates` somewhere?
it might be useful for performance analysis (e.g. relation between `get_volumes_by_pool` execution time and number of volumes in pool)

PR #436, opened 2022-01-16
CSI-3720 downgrade controller builder base image

oriyarde:


Dockerfile-csi-controller.test
```diff
@@ -16,7 +16,7 @@
 #    This Dockerfile.test is for running the csi controller local tests inside a container.
 #    Its similar to the Dockerfile, but with additional requirements-tests.txt and ENTRYPOINT to run the local tests.
 
-FROM registry.access.redhat.com/ubi8/python-38:1-77 as builder
+FROM registry.access.redhat.com/ubi8/python-38:1-75.1638364053 as builder
```
ArbelNathan, line 19: no need for the 3 more places that the version mentioned?

oriyarde: nope

ArbelNathan: ok

ArbelNathan (approved):
LGTM

Dockerfile-csi-controller.test
oriyarde: you were right, need to to do it in the non-test dockerfile as well

PR #441, opened 2022-01-20
CSI 1377 external csi pb2

ArbelNathan:


Dockerfile-csi-controller
```diff
@@ -13,15 +13,21 @@
...
 USER default
 RUN pip3 install --ignore-installed --default-timeout=100 --upgrade pip==21.2.4
 # avoid default boringssl lib, since it does not support z systems
 ENV GRPC_PYTHON_BUILD_SYSTEM_OPENSSL=True
+COPY controller/requirements.txt /tmp/
 RUN pip3 install -r /tmp/requirements.txt
 
+USER root
+COPY controller/scripts/csi_general .
+RUN chmod 755 ./csi_pb2.sh
```
oriyarde, line 27: is 755 the lowest value that works?
could it be submitted that way to begin with? I think it's currently with 644

Dockerfile-csi-controller
```diff
@@ -13,15 +13,21 @@
...
+RUN ./csi_pb2.sh
+RUN pip3 install .
+USER default
```
oriyarde, line 30: I don't think this is necessary

ArbelNathan: so why requirements installed again?

oriyarde: not sure what you mean. there is no need to install the requirements again.
I just don't see why we need the default user here

ArbelNathan: the conversation is only on the  `USER default`?

oriyarde: yes...

ArbelNathan: done

Dockerfile-csi-controller
ArbelNathan: seems that it worked
done

Dockerfile-csi-controller.test
```diff
@@ -17,18 +17,24 @@
 #    Its similar to the Dockerfile, but with additional requirements-tests.txt and ENTRYPOINT to run the local tests.
 
 FROM registry.access.redhat.com/ubi8/python-38:1-75.1638364053 as builder
-COPY controller/requirements.txt /tmp/
 USER root
 RUN if [[ "$(uname -m)" != "x86"* ]]; then yum install -y rust-toolset; fi
 USER default
 RUN pip3 install --ignore-installed --default-timeout=100 --upgrade pip==21.2.4
 # avoid default boringssl lib, since it does not support z systems
 ENV GRPC_PYTHON_BUILD_SYSTEM_OPENSSL=True
```
oriyarde, line 25: I think these two lines should come right before the `pip3 install`

oriyarde (approved):
LGTM

Dockerfile-csi-controller.test
ArbelNathan: done

PR #445, opened 2022-01-27
CSI-3216 add unauthenticated status code

oriyarde:
before:
`rpc error: code = Internal desc = Credential error has occurred while connecting to endpoint ...`
after:
`rpc error: code = Unauthenticated desc = Credential error has occurred while connecting to endpoint ...`


ArbelNathan (approved):
LGTM

PR #449, opened 2022-02-06
Fix/csi 3649 ds8k volume map cache

ArbelNathan:
the cache will be removed in CSI-3865


controller/array_action/array_mediator_ds8k.py
```diff
@@ -280,6 +278,9 @@ def _create_api_volume(self, name, size_in_bytes, array_space_efficiency, pool_i
     def create_volume(self, name, size_in_bytes, space_efficiency, pool):
         array_space_efficiency = get_array_space_efficiency(space_efficiency)
         api_volume = self._create_api_volume(name, size_in_bytes, array_space_efficiency, pool)
+        logger.debug("updating volume {} in cache".format(api_volume.name))
+        with volume_cache_lock:
+            volume_cache.update({api_volume.name: api_volume.id})
```
oriyarde, line 283: isn't this the same as:
```suggestion
            volume_cache[api_volume.name] = api_volume.id
```
?

controller/tests/array_action/ds8k/test_array_mediator_ds8k.py
```diff
@@ -180,19 +210,26 @@ def test_create_volume_fail_with_pool_not_found(self):
             self.array.create_volume("fake_name", 1, 'thin', "fake_pool")
 
     def test_create_volume_fail_with_incorrect_id(self):
-        self.client_mock.get_volumes_by_pool.side_effect = InternalServerError("500", message="BE7A0005")
+        self.client_mock.create_volume.side_effect = InternalServerError("500", message="BE7A0005")
```
oriyarde, line 213: I think the `get` would fail for an invalid id before we reach `create`

controller/array_action/array_mediator_ds8k.py

oriyarde: I think I [still see it](https://github.com/IBM/ibm-block-csi-driver/blob/e1129dd15eb66b6d2f7303006161283234c77e84/controller/tests/array_action/ds8k/test_array_mediator_ds8k.py#L137-L138)

oriyarde: I don't think we need to assert in the beginning of a test.
but if you want control over the cache in the test, can mock it

controller/tests/array_action/ds8k/test_array_mediator_ds8k.py
ArbelNathan: that is a unit test.. in create_volume there is no get

controller/array_action/array_mediator_ds8k.py
ArbelNathan: it is needed for later, If I change it we will get dups.. 
the test is to compare the before and after method. I don't see the problem

oriyarde: I don't see how the assert (before calling the tested method) tests the method.
what happens later? what will get dups?

ArbelNathan: its showing that the cache has been actually updated and not been at the same state as before the method call.
dup for update and assert.

controller/array_action/array_mediator_ds8k.py
```diff
@@ -44,6 +46,9 @@
 ARRAY_SPACE_EFFICIENCY_THIN = ds8k_types.DS8K_TP_ESE
 ARRAY_SPACE_EFFICIENCY_NONE = ds8k_types.DS8K_TP_NONE
 
+volume_cache = dict()
```
oriyarde, line 49: I think this should be per management address, in case there is more than 1.
i.e. per mediator instance.
it should work since we also cache the mediator instances per management address

controller/array_action/array_mediator_ds8k.py

controller/array_action/array_mediator_ds8k.py
ArbelNathan: no, didn't worked

oriyarde: you mean the mediator cache is missing hits? which part didn't work?

ArbelNathan: the cache is always empty like it creates a new one every time

oriyarde: sounds like a bug in the mediator cache.
is the ds8k volume id universally unique? or can it be the same on separate ds8k systems?

controller/array_action/array_mediator_ds8k.py
```diff
@@ -280,6 +278,9 @@ def _create_api_volume(self, name, size_in_bytes, array_space_efficiency, pool_i
     def create_volume(self, name, size_in_bytes, space_efficiency, pool):
         array_space_efficiency = get_array_space_efficiency(space_efficiency)
         api_volume = self._create_api_volume(name, size_in_bytes, array_space_efficiency, pool)
+        logger.debug("updating volume {} in cache".format(api_volume.name))
+        with volume_cache_lock:
+            volume_cache[api_volume.name] = api_volume.id
```
oriyarde, line 283: consider wrapping all `with` blocks with cache class methods

controller/array_action/array_mediator_ds8k.py

controller/array_action/ds8k_volume_cache.py
```diff
@@ -0,0 +1,60 @@
...
+
+from controller.common.csi_logger import get_stdout_logger
+
+logger = get_stdout_logger()
+
+
+class volumeCacheByManagement:
+    def __init__(self):
+        logger.debug("creating a new cache")
+        self.volume_cache_by_address = defaultdict(dict)
```
oriyarde, line 12: I think it should either be "address" or "management", not both.
I think "address" is more accurate, because we cannot tell that e.g. an ip and a hostname are the same management

controller/array_action/ds8k_volume_cache.py
```diff
@@ -0,0 +1,60 @@
...
+    def add(self, key, value):
+        volume_cache_by_management.add(self.service_address, key, value)
+
+    def remove(self, key):
+        volume_cache_by_management.remove(self.service_address, key)
+
+    def get(self, key=None):
+        if key:
+            return volume_cache_by_management.get(self.service_address, key)
+        return volume_cache_by_management.volume_cache_by_address.get(self.service_address)
```
oriyarde, line 57: I think this `if` (+default key value) should be inside the class above, to be lock protected, to be on the safe side

controller/array_action/array_mediator_ds8k.py
```diff
@@ -355,34 +345,45 @@ def _delete_object(self, object_id, object_is_snapshot=False):
...
+        logger.info("deleting volume with id : {0}".format(volume_id))
         self._delete_object(volume_id)
-        logger.info("Finished deleting volume {}".format(volume_id))
+        logger.info("finished deleting volume {}".format(volume_id))
+
+    def _get_api_volume_with_cache(self, name, pool_id):
+        cached_volume_id = self.volume_cache.get(name)
+        api_volume = None
+        if cached_volume_id:
+            logger.debug("found object: {} in cache".format(cached_volume_id))
```
oriyarde, line 360:
```suggestion
            logger.debug("found object id: {} in cache".format(cached_volume_id))
```

controller/array_action/ds8k_volume_cache.py
oriyarde: also I think it will be cleaner to have a separate `get_all` method, instead of empty `get()` calls

ArbelNathan: done

controller/array_action/ds8k_volume_cache.py
ArbelNathan: done

controller/array_action/ds8k_volume_cache.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def get(self, address, key):
+        logger.debug("getting {} from cache".format(key))
+        with self.cache_lock:
+            return self.volume_cache_by_address[address].get(key)
+
+    def get_all(self, address):
+        logger.debug("getting cache for address {}".format(address))
+        with self.cache_lock:
+            return self.volume_cache_by_address.get(address)
```
oriyarde, line 34: if this function is only used in tests, then it should be implemented there instead.
and AFAIK the tests don't asynchronously share state, so this function probably won't need the lock either

controller/array_action/ds8k_volume_cache.py
```diff
@@ -0,0 +1,66 @@
...
+from controller.common.csi_logger import get_stdout_logger
+
+logger = get_stdout_logger()
+
+
+class volumeCacheByAddress:
+    def __init__(self):
+        logger.debug("creating a new cache")
+        self.volume_cache_by_address = defaultdict(dict)
+        self.cache_lock = RLock()
```
oriyarde, line 13: try to keep members private by default wherever possible, in every class

controller/tests/array_action/ds8k/test_array_mediator_ds8k.py
```diff
@@ -70,6 +71,7 @@ def setUp(self):
         )
 
         self.array = DS8KArrayMediator("user", "password", self.endpoint)
+        self.volume_cache = self.array.volume_cache
```
oriyarde, line 74: are you sure this uses the `patch.dict`?

controller/array_action/ds8k_volume_cache.py
ArbelNathan: implement a class method that takes privet params out of the class?

oriyarde: I think it would be best to just mock the cache, and assert the mock calls, without actually needing to compare entire dictionaries

ArbelNathan: done

controller/array_action/ds8k_volume_cache.py
ArbelNathan: done

controller/tests/array_action/ds8k/test_array_mediator_ds8k.py
ArbelNathan: not relevant

controller/tests/array_action/ds8k/test_array_mediator_ds8k.py
```diff
@@ -70,6 +70,10 @@ def setUp(self):
         )
 
         self.array = DS8KArrayMediator("user", "password", self.endpoint)
+        volume_cache_by_address_patcher = patch.dict(
+            "controller.array_action.ds8k_volume_cache.volume_cache_by_address._volume_cache_by_address", clear=True)
```
oriyarde, line 74: I think we can have more abstraction in the unit tests by patching the `VolumeCache` to return a mock, similar to how the ds8k client mediator member is mocked

ArbelNathan: `VolumeCache` does not represent in any way the cache, only refer to it, mocking it wont mock the cache...

ArbelNathan: ok, done

PR #450, opened 2022-02-06
CSI-1285 add more node iscsi login unit tests

oriyarde:
continuation of https://github.com/IBM/ibm-block-csi-driver/pull/131

### DIFF
--- a/node/pkg/driver/device_connectivity/device_connectivity_iscsi_test.go
+++ b/node/pkg/driver/device_connectivity/device_connectivity_iscsi_test.go
@@ -1,0 +1,203 @@
+/**
+ * Copyright 2019 IBM Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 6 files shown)


node/pkg/driver/device_connectivity/device_connectivity_iscsi_test.go
```diff
@@ -0,0 +1,203 @@
...
+			for _, c := range tc.executeWithTimeoutCalls {
+				fakeExecuter.EXPECT().ExecuteWithTimeout(c.input.mSeconds, c.input.command, c.input.commandArguments).Return(
+					c.output.stdOut, c.output.err)
+				if c.output.err != nil && c.output.exitCode != 0 {
+					fakeExecuter.EXPECT().GetExitCode(gomock.Any()).Return(c.output.exitCode, true)
+				}
+			}
+
+			o := NewOsDeviceConnectivityIscsiForTest(fakeExecuter, fakeHelper)
+			o.EnsureLogin(tc.portalsByTarget)
```
ArbelNathan, line 200: fail login not blocking us?

oriyarde: nope

ArbelNathan: why?

oriyarde: we never added a logic that can tell whether the final connection state is "good enough" (I don't think it's trivial).
currently if the final connection state is not "good enough", it will cause failure later in the flow, same as before the login feature

oriyarde: you could call it a "best effort, no guarantee" feature

ArbelNathan: ok

ArbelNathan (approved):
LGTM

PR #451, opened 2022-02-08
CSI-1196 fix invalid name

oriyarde:
continuation of https://github.com/IBM/ibm-block-csi-driver/pull/434 and https://github.com/IBM/ibm-block-csi-driver/pull/444


controller/array_action/ds8k_rest_client.py
```diff
@@ -143,13 +143,13 @@ def get_used_lun_numbers_by_host(self, host_name):
         mappings = self._client.get_mappings_by_host(host_name)
         return [mapping.id for mapping in mappings]
 
-    def create_volume(self, name, capacity_in_bytes, pool_id, tp):
+    def create_volume(self, name, capacity_in_bytes, pool_id, thin_provisioning):
```
ArbelNathan, line 146: even if tp is `thin_provisioning` I don't think it a good name for the parameter.
tp can be every space efficiency we are supporting - including `ese` that we call thin, so it's not the "cross storage" `thin`..

controller/array_action/messages.py
```diff
@@ -1,70 +1,71 @@
-NoConnectionAvailableException_message = "Currently no connection is available to endpoint: {0}"
+NO_CONNECTION_AVAILABLE_EXCEPTION_MESSAGE = "Currently no connection is available to endpoint: {0}"
```
ArbelNathan, line 1: Do we don't want to remove the upper case?

controller/controller_server/csi_controller_server.py
```diff
@@ -495,7 +495,7 @@ def get_identity_config(self, attribute_name):
         return self.cfg['identity'][attribute_name]
 
     @csi_method(error_response_type=csi_pb2.GetPluginInfoResponse)
-    def GetPluginInfo(self, _, context):
+    def GetPluginInfo(self, _, context):  # pylint: disable=invalid-name
```
ArbelNathan, line 498: why not in the rest of the `spec` funcs?

controller/array_action/messages.py
oriyarde: it's a const and this is the common convention for consts

controller/controller_server/csi_controller_server.py
oriyarde: the linter doesn't complain.
perhaps when these ones are disabled, the class majority becomes `PascalCase`, then the linter gives up or something

controller/array_action/messages.py
ArbelNathan: I meant for the msg

controller/controller_server/csi_controller_server.py
ArbelNathan: weird..

controller/array_action/ds8k_rest_client.py
ArbelNathan: we would still need to use tp eventually where?

controller/array_action/messages.py
oriyarde: oh, sure, but I think there is enough diff in this PR.
we can align all messages content in a single separate PR

ArbelNathan: can you open a ticket?

controller/array_action/ds8k_rest_client.py
ArbelNathan: we are using `tp` there

oriyarde: exactly.
so in the function call "flow", the value is initially called "space efficiency", but would eventually have to be called `tp`.
we can either call it "space efficiency" all the way until the last call (shown in the link above), or have this intermediate name to explain why "space efficiency" becomes "tp"

controller/array_action/messages.py
oriyarde: CSI-3899

controller/array_action/ds8k_rest_client.py
ArbelNathan: but why using `thin_provisioning` and not stay with `tp`?

oriyarde: because `tp` is too short, it's not descriptive enough (pylint default minimum length is 3)

ArbelNathan: ok, fair enough

ArbelNathan (approved):
LGTM

PR #454, opened 2022-02-13
CSI-2142 fix unstable mediator pool unit test

oriyarde:
the non-reviewed https://github.com/IBM/ibm-block-csi-driver/pull/170 has many issues, this fixes one of them:
a `sleep` does not guarantee that the desired state is always reached (in this case, that all remaining threads have created, and are still using, their mediator in the pool), no matter what hardcoded waiting time is chosen


ArbelNathan (approved):
LGTM

PR #438, opened 2022-01-19
CSI-3720 downgrade controller builder base image

oriyarde:
continuation of https://github.com/IBM/ibm-block-csi-driver/pull/436

PR #455, opened 2022-02-15
CSI-3736 bump base images

oriyarde:


Dockerfile-csi-controller
```diff
@@ -29,7 +29,7 @@ RUN ./csi_pb2.sh
 RUN pip3 install .
 
 
-FROM registry.access.redhat.com/ubi8/python-38:1-77
+FROM registry.access.redhat.com/ubi8/python-38:1-80
```
ArbelNathan, line 32: what about the `FROM` [above](https://github.com/IBM/ibm-block-csi-driver/blob/16ad98f5c311925648a2bcb3464b8db7cc0c8b33/Dockerfile-csi-controller#L15)?

ArbelNathan: same in .test

oriyarde: not relevant since https://github.com/IBM/ibm-block-csi-driver/pull/438

ArbelNathan: it looks there that both images in the file have been downgraded, why is the second one is ok to bump?

oriyarde: because it doesn't fail on compilation, it just contains the complied files

ArbelNathan (approved):
LGTM

PR #457, opened 2022-02-20
Feature/csi 3897 implement node get volume stats

matancarmeli7:
it passed community tests and sanity tests

### DIFF
--- a/node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
+++ b/node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
@@ -75,7 +75,7 @@ const (
 	IscsiHostRexExPath          = "/sys/class/iscsi_host/host*/device/session*/iscsi_session/session*/targetname"
 	sysDeviceSymLinkFormat      = "/sys/block/%s/device"
 	sysDeviceDeletePathFormat   = sysDeviceSymLinkFormat + "/delete"
-	BlockDevCmd                 = "blockdev"
+	blockDevCmd                 = "blockdev"
 	flushBufsFlag               = "--flushbufs"
 	mpathdSeparator             = ","
 	multipathdCmd               = "multipathd"
@@ -201,9 +201,9 @@ func (r OsDeviceConnectivityHelperScsiGeneric) GetMpathDevice(volumeId string) (
 
 func (r OsDeviceConnectivityHelperScsiGeneric) flushDeviceBuffers(deviceName string) error {
 	devicePath := filepath.Join(DevPath, deviceName)
-	_, err := r.Executer.ExecuteWithTimeoutSilently(TimeOutBlockDevCmd, BlockDevCmd, []string{flushBufsFlag, devicePath})
+	_, err := r.Executer.ExecuteWithTimeoutSilently(TimeOutBlockDevCmd, blockDevCmd, []string{flushBufsFlag, devicePath})
 	if err != nil {
-		logger.Errorf("%v %v {%v} did not succeed to flush the device buffers. err={%v}", BlockDevCmd, flushBufsFlag, devicePath,
+		logger.Errorf("%v %v {%v} did not succeed to flush the device buffers. err={%v}", blockDevCmd, flushBufsFlag, devicePath,
 			err.Error())
 		return err
 	}
@@ -211,14 +211,14 @@ func (r OsDeviceConnectivityHelperScsiGeneric) flushDeviceBuffers(deviceName str
 }
 
 func (r OsDeviceConnectivityHelperScsiGeneric) flushDevicesBuffers(deviceNames []string) error {
-	logger.Debugf("executing commands : {%v %v} on devices : {%v} and timeout : {%v} mseconds", BlockDevCmd, flushBufsFlag, deviceNames, TimeOutBlockDevCmd)
+	logger.Debugf("executing commands : {%v %v} on devices : {%v} and timeout : {%v} mseconds", blockDevCmd, flushBufsFlag, deviceNames, TimeOutBlockDevCmd)
 	for _, deviceName := range deviceNames {
 		err := r.flushDeviceBuffers(deviceName)
 		if err != nil {
 			return err
 		}
 	}
-	logger.Debugf("Finished executing commands: {%v %v}", BlockDevCmd, flushBufsFlag)
+	logger.Debugf("Finished executing commands: {%v %v}", blockDevCmd, flushBufsFlag)
 	return nil
 }
 
--- a/node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
+++ b/node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
@@ -75,7 +75,7 @@ const (
 	IscsiHostRexExPath          = "/sys/class/iscsi_host/host*/device/session*/iscsi_session/session*/targetname"
 	sysDeviceSymLinkFormat      = "/sys/block/%s/device"
 	sysDeviceDeletePathFormat   = sysDeviceSymLinkFormat + "/delete"
-	blockDevCmd                 = "blockdev"
+	BlockDevCmd                 = "blockdev"
 	flushBufsFlag               = "--flushbufs"
 	mpathdSeparator             = ","
 	multipathdCmd               = "multipathd"
@@ -201,9 +201,9 @@ func (r OsDeviceConnectivityHelperScsiGeneric) GetMpathDevice(volumeId string) (
 
 func (r OsDeviceConnectivityHelperScsiGeneric) flushDeviceBuffers(deviceName string) error {
 	devicePath := filepath.Join(DevPath, deviceName)
-	_, err := r.Executer.ExecuteWithTimeoutSilently(TimeOutBlockDevCmd, blockDevCmd, []string{flushBufsFlag, devicePath})
+	_, err := r.Executer.ExecuteWithTimeoutSilently(TimeOutBlockDevCmd, BlockDevCmd, []string{flushBufsFlag, devicePath})
 	if err != nil {
-		logger.Errorf("%v %v {%v} did not succeed to flush the device buffers. err={%v}", blockDevCmd, flushBufsFlag, devicePath,
+		logger.Errorf("%v %v {%v} did not succeed to flush the device buffers. err={%v}", BlockDevCmd, flushBufsFlag, devicePath,
 			err.Error())
 		return err
 	}
@@ -211,14 +211,14 @@ func (r OsDeviceConnectivityHelperScsiGeneric) flushDeviceBuffers(deviceName str
 }
 
 func (r OsDeviceConnectivityHelperScsiGeneric) flushDevicesBuffers(deviceNames []string) error {
-	logger.Debugf("executing commands : {%v %v} on devices : {%v} and timeout : {%v} mseconds", blockDevCmd, flushBufsFlag, deviceNames, TimeOutBlockDevCmd)
+	logger.Debugf("executing commands : {%v %v} on devices : {%v} and timeout : {%v} mseconds", BlockDevCmd, flushBufsFlag, deviceNames, TimeOutBlockDevCmd)
 	for _, deviceName := range deviceNames {
 		err := r.flushDeviceBuffers(deviceName)
 		if err != nil {
 			return err
 		}
 	}
-	logger.Debugf("Finished executing commands: {%v %v}", blockDevCmd, flushBufsFlag)
+	logger.Debugf("Finished executing commands: {%v %v}", BlockDevCmd, flushBufsFlag)
 	return nil
 }
 
--- a/node/pkg/driver/node.go
+++ b/node/pkg/driver/node.go
@@ -631,7 +631,6 @@ func (d *NodeService) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVo
 	defer goid_info.DeleteAdditionalIDInfo()
 	volumePath := req.VolumePath
 	volumePathWithHostPrefix := d.NodeUtils.GetPodPath(req.VolumePath)
-	isFSVolume := true
 
 	err := d.nodeGetVolumeStatsRequestValidation(volumeId, volumePath)
 	if err != nil {
@@ -642,21 +641,8 @@ func (d *NodeService) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVo
 	if !isPathExists {
 		return nil, status.Errorf(codes.NotFound, "volume path %q does not exist", volumePath)
 	}
-	isBlock, err := d.NodeUtils.IsBlock(volumePathWithHostPrefix)
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to determine if %q is block device: %s", volumePath, err)
-	}
-	if isBlock {
-		isFSVolume = false
-	}
-	isMounted, err := d.IsMounted(volumePathWithHostPrefix, isFSVolume)
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to check if volume path %q is mounted", volumePath)
-	}
-	if !isMounted {
-		return nil, status.Errorf(codes.NotFound, "volume path %q is not mounted", volumePath)
-	}
-	volumeStats, err := d.NodeUtils.GetVolumeStats(volumePathWithHostPrefix, isFSVolume)
+
+	volumeStats, err := d.GetVolumeStats(volumePathWithHostPrefix, volumeId)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "Failed to get statistics for volume path %q: %s", volumePath, err)
 	}
@@ -687,6 +673,23 @@ func (d *NodeService) nodeGetVolumeStatsRequestValidation(volumeId string, volum
 	}
 
 	return nil
+}
+
+func (d *NodeService) GetVolumeStats(path string, volumeId string) (VolumeStatistics, error) {
+	isBlock, err := d.NodeUtils.IsBlock(path)
+	if err != nil {
+		return VolumeStatistics{}, status.Errorf(codes.Internal, "failed to determine if %q is block device: %s", path, err)
+	}
+	
+	if isBlock {
+		mpathDevice, err := d.OsDeviceConnectivityHelper.GetMpathDevice(volumeId)
+		if err != nil {
+			return VolumeStatistics{}, status.Errorf(codes.Internal, "Error while discovering the device : %s", err)
+		}
+		return d.NodeUtils.GetBlockVolumeStats(mpathDevice)
+	} else {
+		return d.NodeUtils.GetFileSystemVolumeStats(path)
+	}
 }
 
 func (d *NodeService) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolumeRequest) (*csi.NodeExpandVolumeResponse, error) {
--- a/node/pkg/driver/node.go
+++ b/node/pkg/driver/node.go
@@ -508,14 +508,7 @@ func (d *NodeService) publishRawBlockVolume(mpathDevice string, targetPath strin
 // Returns: is <target mounted, error if occured>
 func (d *NodeService) isTargetMounted(targetPathWithHostPrefix string, isFSVolume bool) (bool, error) {
 	logger.Debugf("Check if target {%s} is mounted", targetPathWithHostPrefix)
-	isMounted, err := d.IsMounted(targetPathWithHostPrefix, isFSVolume)
-	if isMounted && err == nil {
-		logger.Warningf("Idempotent case : targetPath already mounted (%s), so no need to mount again. Finish NodePublishVolume", targetPathWithHostPrefix)
-	}
-	return isMounted, err
-}
-
-func (d *NodeService) IsMounted(targetPathWithHostPrefix string, isFSVolume bool) (bool, error) {
+	logger.Debugf("Check if target {%s} is mounted", targetPathWithHostPrefix)
 	isNotMounted, err := d.NodeUtils.IsNotMountPoint(targetPathWithHostPrefix)
 	if err != nil {
 		logger.Warningf("Failed to check if (%s), is mounted", targetPathWithHostPrefix)
@@ -530,6 +523,7 @@ func (d *NodeService) IsMounted(targetPathWithHostPrefix string, isFSVolume bool
 		} else if !isFSVolume && targetIsDir {
 			return true, status.Errorf(codes.AlreadyExists, "Required raw block volume but target {%s} is mounted and it is a directory.", targetPathWithHostPrefix)
 		}
+		logger.Warningf("Idempotent case : targetPath already mounted (%s), so no need to mount again. Finish NodePublishVolume", targetPathWithHostPrefix)
 		return true, nil
 	}
 }
@@ -684,7 +678,12 @@ func (d *NodeService) GetVolumeStats(path string, volumeId string) (VolumeStatis
 	if isBlock {
 		mpathDevice, err := d.OsDeviceConnectivityHelper.GetMpathDevice(volumeId)
 		if err != nil {
-			return VolumeStatistics{}, status.Errorf(codes.Internal, "Error while discovering the device : %s", err)
+			switch err.(type) {
+			case *device_connectivity.MultipathDeviceNotFoundForVolumeError:
+				return VolumeStatistics{}, status.Errorf(codes.NotFound, "Multipath Device of volume id %q does not exist", volumeId)
+			default:
+				return VolumeStatistics{}, status.Errorf(codes.Internal, "Error while discovering the device : %s", err)
+			}
 		}
 		return d.NodeUtils.GetBlockVolumeStats(mpathDevice)
 	} else {
--- a/node/pkg/driver/node.go
+++ b/node/pkg/driver/node.go
@@ -670,10 +670,9 @@ func (d *NodeService) nodeGetVolumeStatsRequestValidation(volumeId string, volum
 
 func (d *NodeService) getVolumeStats(path string, volumeId string) (VolumeStatistics, error) {
 	var volumeStats VolumeStatistics
-	errorMsg := "Failed to get statistics:"
 	isBlock, err := d.NodeUtils.IsBlock(path)
 	if err != nil {
-		return VolumeStatistics{}, status.Errorf(codes.Internal, "%q failed to determine if %q is block device: %s", errorMsg, path, err)
+		return VolumeStatistics{}, status.Errorf(codes.Internal, "Failed to determine if %q is block device: %s", path, err)
 	}
 
 	if isBlock {
@@ -681,9 +680,9 @@ func (d *NodeService) getVolumeStats(path string, volumeId string) (VolumeStatis
 		if err != nil {
 			switch err.(type) {
 			case *device_connectivity.MultipathDeviceNotFoundForVolumeError:
-				return VolumeStatistics{}, status.Errorf(codes.NotFound, "%q multipath Device of volume id %q does not exist", errorMsg, volumeId)
+				return VolumeStatistics{}, status.Errorf(codes.NotFound, "Multipath device of volume id %q does not exist", volumeId)
 			default:
-				return VolumeStatistics{}, status.Errorf(codes.Internal, "%q error while discovering the device : %s", errorMsg, err)
+				return VolumeStatistics{}, status.Errorf(codes.Internal, "Error while discovering the device : %s", err)
 			}
 		}
 		volumeStats, err = d.NodeUtils.GetBlockVolumeStats(mpathDevice)
@@ -692,7 +691,7 @@ func (d *NodeService) getVolumeStats(path string, volumeId string) (VolumeStatis
 	}
 
 	if err != nil {
-		return VolumeStatistics{}, status.Errorf(codes.Internal, "%q failed to determine if %q is block device: %s", errorMsg, path, err)
+		return VolumeStatistics{}, status.Errorf(codes.Internal, "Failed to get statistics: %s", err)
 	}
 	return volumeStats, nil
 }
--- a/node/pkg/driver/node.go
+++ b/node/pkg/driver/node.go
@@ -19,8 +19,6 @@ package driver
 import (
 	"context"
 	"fmt"
-	"io"
-	"os"
 	"path"
 	"strings"
 
@@ -87,6 +85,11 @@ type NodeService struct {
 	VolumeIdLocksMap            SyncLockInterface
 	OsDeviceConnectivityMapping map[string]device_connectivity.OsDeviceConnectivityInterface
 	OsDeviceConnectivityHelper  device_connectivity.OsDeviceConnectivityHelperScsiGenericInterface
+}
+
+type volumeStatistics struct {
+	availableBytes, totalBytes, usedBytes    int64
+	availableInodes, totalInodes, usedInodes int64
 }
 
 // newNodeService creates a new node service
@@ -648,18 +651,6 @@ func (d *NodeService) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVo
 	}
 	if isBlock {
 		isFSVolume = false
-		//blockSize, err := d.getBlockSize(volumePathWithHostPrefix)
-		//if err != nil {
-		//	return nil, status.Errorf(codes.Internal, "failed to retrieve capacity statistics for block volume %q: %s", volumePath, err)
-		//}
-		//return &csi.NodeGetVolumeStatsResponse{
-		//	Usage: []*csi.VolumeUsage{
-		//		{
-		//			Unit:  csi.VolumeUsage_BYTES,
-		//			Total: blockSize,
-		//		},
-		//	},
-		//}, nil
 	}
 	isMounted, err := d.IsMounted(volumePathWithHostPrefix, isFSVolume)
 	if err != nil {
@@ -668,7 +659,7 @@ func (d *NodeService) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVo
 	if !isMounted {
 		return nil, status.Errorf(codes.NotFound, "volume path %q is not mounted", volumePath)
 	}
-	available, capacity, usage, inodes, inodesFree, inodesUsed, err := d.getFilesystemStats(volumePathWithHostPrefix)
+	volumeStats, err := d.getVolumeStats(volumePathWithHostPrefix)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "Failed to get capacity statistics for volume path %q: %s", volumePath, err)
 	}
@@ -676,15 +667,15 @@ func (d *NodeService) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVo
 		Usage: []*csi.VolumeUsage{
 			{
 				Unit:      csi.VolumeUsage_BYTES,
-				Available: available,
-				Total:     capacity,
-				Used:      usage,
+				Available: volumeStats.availableBytes,
+				Total:     volumeStats.totalBytes,
+				Used:      volumeStats.usedBytes,
 			},
 			{
 				Unit:      csi.VolumeUsage_INODES,
-				Available: inodesFree,
-				Total:     inodes,
-				Used:      inodesUsed,
+				Available: volumeStats.availableInodes,
+				Total:     volumeStats.totalInodes,
+				Used:      volumeStats.usedInodes,
 			},
 		},
 	}, nil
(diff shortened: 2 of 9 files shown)


node/pkg/driver/node.go
```diff
@@ -613,9 +627,95 @@ func (d *NodeService) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+	}
+	if !isMounted {
+		return nil, status.Errorf(codes.NotFound, "volume path %q is not mounted", volumePath)
+	}
+	volumeStats, err := d.getVolumeStats(volumePathWithHostPrefix)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Failed to get capacity statistics for volume path %q: %s", volumePath, err)
+	}
+	return &csi.NodeGetVolumeStatsResponse{
+		Usage: []*csi.VolumeUsage{
```
ArbelNathan, line 667: don't we want volume_condition too?
I can see it's Must if we support "VOLUME_CONDITION node capability" but any reason why not do it anyway?

matancarmeli7: `A Node plugin MUST implement this RPC call if it has GET_VOLUME_STATS node capability or VOLUME_CONDITION node capability`
this is what they write in the spec, so it's enough to add GET_VOLUME_STATS node capability, no?

node/pkg/driver/node.go
```diff
@@ -613,9 +627,95 @@ func (d *NodeService) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+	goid_info.SetAdditionalIDInfo(volumeId)
 	defer goid_info.DeleteAdditionalIDInfo()
-	return nil, status.Error(codes.Unimplemented, "NodeGetVolumeStats is not implemented yet")
+
+	if volumeId == "" {
+		return nil, status.Error(codes.InvalidArgument, "NodeGetVolumeStats Volume ID must be provided")
+	}
+	if volumePath == "" {
+		return nil, status.Error(codes.InvalidArgument, "NodeGetVolumeStats Volume Path must be provided")
+	}
```
oriyarde, line 642: please move this to a `nodeGetVolumeStatsRequestValidation` function

oriyarde (commented):
please rename in this PR title `2897` to `3897`

node/pkg/driver/node.go
oriyarde: we don't support [volume health monitoring](https://kubernetes.io/blog/2021/04/16/volume-health-monitoring-alpha-update/#node-side-volume-health-monitoring), I couldn't find other CSI drivers that support this capacility, I don't think it's trivial and it's still alpha anyway

node/pkg/driver/node.go
matancarmeli7: done

matancarmeli7:
> please rename in this PR title `2897` to `3897`

done

node/pkg/driver/node.go
```diff
@@ -613,9 +627,104 @@ func (d *NodeService) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 }
 
 func (d *NodeService) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVolumeStatsRequest) (*csi.NodeGetVolumeStatsResponse, error) {
-	goid_info.SetAdditionalIDInfo(req.VolumeId)
+	volumeId := req.VolumeId
+	volumePath := req.VolumePath
+	volumePathWithHostPrefix := d.NodeUtils.GetPodPath(req.VolumePath)
+	isFSVolume := true
+	goid_info.SetAdditionalIDInfo(volumeId)
```
oriyarde, line 634: please set the additional id info before any other logic (before e.g. `GetPodPath`)

matancarmeli7: also the defer line? or it's doesn't matter?

node/pkg/driver/node_test.go
```diff
@@ -895,23 +895,32 @@ func TestNodeUnpublishVolume(t *testing.T) {
...
-	if err == nil {
-		t.Fatalf("Expected error code %d, got nil", expErrCode)
-	}
-	srvErr, ok := status.FromError(err)
-	if !ok {
-		t.Fatalf("Could not get error status code from error: %v", srvErr)
-	}
-	if srvErr.Code() != expErrCode {
-		t.Fatalf("Expected error code %d, got %d message %s", expErrCode, srvErr.Code(), srvErr.Message())
+	if err != nil {
```
oriyarde, line 916: shouldn't the test fail if `err` is `nil`?
in addition, please add a "happy flow" test case, to see the (fake) stats values actually exist in the response volume usage

oriyarde: another test case I just thought of: assert unix stat failure leads to a CSI internal status code error

node/pkg/driver/node.go
oriyarde: the defer should also come right after the set

node/pkg/driver/node.go
```diff
@@ -613,9 +627,104 @@ func (d *NodeService) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+		return nil, status.Errorf(codes.NotFound, "volume path %q does not exists", volumePath)
+	}
+	isBlock, err := d.isBlock(volumePathWithHostPrefix)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to determine if %q is block device: %s", volumePath, err)
+	}
+	if isBlock {
+		isFSVolume = false
+	}
+	isMounted, err := d.IsMounted(volumePathWithHostPrefix, isFSVolume)
```
oriyarde, line 653: what happens if we don't check the volume mode?

node/pkg/driver/node.go
ArbelNathan: ok, tnx.

ArbelNathan (approved):
LGTM

node/pkg/driver/node.go
matancarmeli7: done

node/pkg/driver/node.go
matancarmeli7: it will probably show stats on unmounted file/directory

node/pkg/driver/node_test.go
matancarmeli7: added an handle to nil error

node/pkg/driver/node.go
oriyarde: sounds similar to https://github.com/IBM/ibm-block-csi-driver/pull/414
do you think we could unify these logics into a common function?

oriyarde: I don't see this check as a requirement from the SP side in the [CSI spec](https://github.com/container-storage-interface/spec/blob/v1.5.0/spec.md#nodegetvolumestats).
but in general we are expected to ensure that the specified path leads to the specified volume id

node/pkg/driver/node.go
```diff
@@ -613,9 +627,104 @@ func (d *NodeService) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+}
+
+//func (d *NodeService) isBlock(devicePath string) (bool, error) {
+//	var stat unix.Stat_t
+//	err := unix.Stat(devicePath, &stat)
+//	if err != nil {
+//		return false, err
+//	}
+//	return (stat.Mode & unix.S_IFMT) == unix.S_IFBLK, nil
+//}
```
oriyarde, line 700: if you want to experiment, please move this PR to draft until you finish (we don't include commented-out code in PRs)

node/pkg/driver/node_test.go
matancarmeli7: done both of the unit tests

node/pkg/driver/node.go
matancarmeli7: sorry it was for Arbel to see something, removed it

node/pkg/driver/node_test.go
```diff
@@ -895,23 +895,34 @@ func TestNodeUnpublishVolume(t *testing.T) {
...
+	volumePath := "/test/path"
+	stagingTargetPath := "/staging/test/path"
+	expErrCode := codes.NotFound
 
-	req := &csi.NodeGetVolumeStatsRequest{}
-
-	d := newTestNodeService(nil, nil, nil)
-
-	expErrCode := codes.Unimplemented
+	mockNodeUtils.EXPECT().IsPathExists("/host" + volumePath).Return(false)
```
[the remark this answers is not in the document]
matancarmeli7: done

node/pkg/driver/node_test.go
```diff
@@ -895,23 +896,141 @@ func TestNodeUnpublishVolume(t *testing.T) {
...
+		{
+			name: "fail volumePath does not exists",
+			testFunc: func(t *testing.T) {
+				mockCtl := gomock.NewController(t)
+				defer mockCtl.Finish()
+				mockNodeUtils := mocks.NewMockNodeUtilsInterface(mockCtl)
+				d := newTestNodeService(mockNodeUtils, nil, nil)
+				expErrCode := codes.NotFound
+				mockNodeUtils.EXPECT().IsPathExists(volumePathWithHostPrefix).Return(false)
+				mockNodeUtils.EXPECT().GetPodPath(volumePath).Return(volumePathWithHostPrefix)
```
oriyarde, line 917: it would be easier to follow if the EXPECTED are ordered according to actual call order

node/pkg/driver/node_test.go
```diff
@@ -895,23 +896,141 @@ func TestNodeUnpublishVolume(t *testing.T) {
...
+				if err != nil {
+					srvErr, ok := status.FromError(err)
+					if !ok {
+						t.Fatalf("Could not get error status code from error: %v", srvErr)
+					}
+					if !strings.Contains(err.Error(), "Failed to get capacity statistics") {
+						t.Fatalf("Expected to get error from GetVolumeStats function, got error message: %s", err.Error())
+					}
+				} else {
+					t.Fatalf("Expected to get an error, got nil")
```
oriyarde, line 967: this looks like code dup.
would moving the actual `NodeGetVolumeStats` calls into a common loop at the end of the function, make the code shorter/cleaner (as was done e.g. [here](https://github.com/IBM/ibm-block-csi-driver/pull/450))?

node/pkg/driver/node.go
```diff
@@ -613,9 +627,104 @@ func (d *NodeService) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+	totalBytes := int64(statfs.Blocks) * int64(statfs.Bsize)
+	usedBytes := (int64(statfs.Blocks) - int64(statfs.Bfree)) * int64(statfs.Bsize)
+
+	totalInodes := int64(statfs.Files)
+	availableInodes := int64(statfs.Ffree)
+	usedInodes := totalInodes - availableInodes
+
+	volumeStats := volumeStatistics{
+		availableBytes: availableBytes,
+		totalBytes:     totalBytes,
```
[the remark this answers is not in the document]
matancarmeli7: I check again and saw that I didn't check it well the first time and I changed the way we get `totalBytes` for block devices and changed that when it's block device we will return 0 on every other stat that we return on file system volumes.

I also saw what digital ocean did, and I didn't like the way they got the `totalBytes` from block devices so I used another way that I checked and it works.

node/pkg/driver/node_test.go
matancarmeli7: done

matancarmeli7: done

node/pkg/driver/node_test.go
matancarmeli7: I moved some calls to functions (not all the `NodeGetVolumeStats` calls), I think it's cleaner, is it enough? or there is something else I can do here?

node/pkg/driver/node.go
```diff
@@ -613,9 +626,67 @@ func (d *NodeService) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+	isFSVolume := true
+
+	err := d.nodeGetVolumeStatsRequestValidation(volumeId, volumePath)
+	if err != nil {
+		return nil, err
+	}
+
+	isPathExists := d.NodeUtils.IsPathExists(volumePathWithHostPrefix)
+	if !isPathExists {
+		return nil, status.Errorf(codes.NotFound, "volume path %q does not exists", volumePath)
```
oriyarde, line 643:
```suggestion
		return nil, status.Errorf(codes.NotFound, "volume path %q does not exist", volumePath)
```

node/pkg/driver/node.go
```diff
@@ -613,9 +626,67 @@ func (d *NodeService) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+	isMounted, err := d.IsMounted(volumePathWithHostPrefix, isFSVolume)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to check if volume path %q is mounted", volumePath)
+	}
+	if !isMounted {
+		return nil, status.Errorf(codes.NotFound, "volume path %q is not mounted", volumePath)
+	}
+	volumeStats, err := d.NodeUtils.GetVolumeStats(volumePathWithHostPrefix, isFSVolume)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Failed to get capacity statistics for volume path %q: %s", volumePath, err)
```
oriyarde, line 661:
```suggestion
		return nil, status.Errorf(codes.Internal, "Failed to get statistics for volume path %q: %s", volumePath, err)
```
I think "capacity" is only the "total"

node/pkg/driver/node_test.go
```diff
@@ -1338,6 +1459,16 @@ func assertError(t *testing.T, err error, expectedErrorCode codes.Code) {
 	}
 }
 
+func assertExpectedStats(t *testing.T, expResp *csi.NodeGetVolumeStatsResponse, req *csi.NodeGetVolumeStatsRequest, node driver.NodeService) {
```
oriyarde, line 1462: I would put this next to the `NodeGetVolumeStats` tests, since it's not relevant to other tests

node/pkg/driver/node_test.go
```diff
@@ -895,23 +896,136 @@ func TestNodeUnpublishVolume(t *testing.T) {
...
+				mockNodeUtils.EXPECT().IsPathExists(volumePathWithHostPrefix).Return(true)
+				mockNodeUtils.EXPECT().IsBlock(volumePathWithHostPrefix).Return(false, nil)
+				mockNodeUtils.EXPECT().IsNotMountPoint(volumePathWithHostPrefix).Return(false, nil)
+				mockNodeUtils.EXPECT().IsDirectory(volumePathWithHostPrefix).Return(true)
+				mockNodeUtils.EXPECT().GetVolumeStats(volumePathWithHostPrefix, true).Return(driver.VolumeStatistics{}, status.Errorf(codes.Internal, "fail to get stats"))
+				req := &csi.NodeGetVolumeStatsRequest{
+					VolumeId:          volId,
+					VolumePath:        volumePath,
+					StagingTargetPath: stagingTargetPath,
+				}
```
oriyarde, line 942: it seems that this can be moved to the beginning of `TestNodeGetVolumeStats`

node/pkg/driver/node_utils.go
```diff
@@ -515,3 +518,65 @@ func (n NodeUtils) GetTopologyLabels(ctx context.Context, nodeName string) (map[
...
+    if err != nil {
+        return VolumeStatistics{}, err
+    }
+
+	volumeStats := VolumeStatistics{
+		TotalBytes:     totalSize,
+	}
+
+	return volumeStats, nil
+}
```
oriyarde, line 582: image

node/pkg/driver/node_utils.go
```diff
@@ -515,3 +518,65 @@ func (n NodeUtils) GetTopologyLabels(ctx context.Context, nodeName string) (map[
...
+		AvailableInodes: availableInodes,
+		TotalInodes:     totalInodes,
+		UsedInodes:      usedInodes,
+	}
+
+	return volumeStats, nil
+}
+
+func (d NodeUtils) getBlockVolumeStats(path string) (VolumeStatistics, error) {
+	blockDevice, err := os.Open(path)
```
oriyarde, line 568: I'm not sure we are suppose to open a user's device. if e.g. there's some limit on the number of open file descriptors, we are limiting the user

node/pkg/driver/node_test.go
oriyarde: much better

node/pkg/driver/node.go
```diff
@@ -613,9 +626,67 @@ func (d *NodeService) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+		isFSVolume = false
+	}
+	isMounted, err := d.IsMounted(volumePathWithHostPrefix, isFSVolume)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to check if volume path %q is mounted", volumePath)
+	}
+	if !isMounted {
+		return nil, status.Errorf(codes.NotFound, "volume path %q is not mounted", volumePath)
+	}
+	volumeStats, err := d.NodeUtils.GetVolumeStats(volumePathWithHostPrefix, isFSVolume)
```
oriyarde, line 659: according to the [spec](https://github.com/container-storage-interface/spec/blob/v1.5.0/spec.md#nodegetvolumestats), regarding the `VolumePath`:
>  // It can be any valid path where volume was previously
  // staged or published.

and we currently [don't](https://github.com/IBM/ibm-block-csi-driver/blob/v1.8.0/node/pkg/driver/node.go#L159) mount block volumes on stage, so I think it would be better to find the relevant path from the volume id (we need it for [NOT_FOUND](https://github.com/IBM/ibm-block-csi-driver/pull/457#discussion_r810976948) [grpc code](https://github.com/container-storage-interface/spec/blob/v1.5.0/spec.md#nodegetvolumestats-errors) anyway)

node/pkg/driver/node.go
matancarmeli7: done

node/pkg/driver/node.go
matancarmeli7: done

node/pkg/driver/node_test.go
matancarmeli7: done

node/pkg/driver/node_test.go
matancarmeli7: done

node/pkg/driver/node_test.go
```diff
@@ -895,23 +896,136 @@ func TestNodeUnpublishVolume(t *testing.T) {
...
+
+				_, err := d.NodeGetVolumeStats(context.TODO(), req)
+				assertError(t, err, expErrCode)
+				if err != nil && !strings.Contains(err.Error(), "Failed to get capacity statistics") {
+					t.Fatalf("Expected to get error from GetVolumeStats function, got error message: %s", err.Error())
```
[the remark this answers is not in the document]
matancarmeli7: done

node/pkg/driver/node_utils.go
matancarmeli7: done

node/pkg/driver/node_utils.go
matancarmeli7: I found two ways, the first one is to use `blockdev` command as digital ocean used, my fear from this use case is that the command can be changed or the output can be changed.

The second option is what I did here.

What do you think we should do? or there is another way that you know how to do it?

node/pkg/driver/node_utils.go
```diff
@@ -515,3 +518,65 @@ func (n NodeUtils) GetTopologyLabels(ctx context.Context, nodeName string) (map[
...
+    if err != nil {
```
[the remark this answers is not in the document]
matancarmeli7: done my bad

node/pkg/driver/node.go
matancarmeli7: how can I get the volume path from the volume id? it doesn't have a path in the volume id?

oriyarde: not sure, but there is `GetMpathDevice`, and from there, there should be a way

node/pkg/driver/node_test.go
```diff
@@ -895,23 +896,132 @@ func TestNodeUnpublishVolume(t *testing.T) {
 }
 
 func TestNodeGetVolumeStats(t *testing.T) {
```
oriyarde, line 898: do you think we should add tests in `node_utils_test.go` as well?

oriyarde: can't we just mock the executer calls (as was done in https://github.com/IBM/ibm-block-csi-driver/pull/450)?

matancarmeli7: can I open a ticket for it and do it next sprint?

node/pkg/driver/node_utils.go
matancarmeli7: done

node/pkg/driver/node_test.go
oriyarde: sure, not urgent

node/pkg/driver/node.go
```diff
@@ -613,9 +626,70 @@ func (d *NodeService) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+func (d *NodeService) GetVolumeStats(path string, volumeId string) (VolumeStatistics, error) {
+	isBlock, err := d.NodeUtils.IsBlock(path)
+	if err != nil {
+		return VolumeStatistics{}, status.Errorf(codes.Internal, "failed to determine if %q is block device: %s", path, err)
+	}
+	
+	if isBlock {
+		mpathDevice, err := d.OsDeviceConnectivityHelper.GetMpathDevice(volumeId)
+		if err != nil {
+			return VolumeStatistics{}, status.Errorf(codes.Internal, "Error while discovering the device : %s", err)
```
oriyarde, line 687: consider returning "not found" code for `MultipathDeviceNotFoundForVolumeError`

node/pkg/driver/node.go
matancarmeli7: opened a ticket for it as we discussed CSI-3946

node/pkg/driver/node.go
matancarmeli7: opened a ticket for it as we discussed CSI-3946

node/pkg/driver/node_utils.go
```diff
@@ -515,3 +519,60 @@ func (n NodeUtils) GetTopologyLabels(ctx context.Context, nodeName string) (map[
...
+		TotalInodes:     totalInodes,
+		UsedInodes:      usedInodes,
+	}
+
+	return volumeStats, nil
+}
+
+func (d NodeUtils) GetBlockVolumeStats(mpathDevice string) (VolumeStatistics, error) {
+	args := []string{"--getsize64", mpathDevice}
+	out, err := d.Executer.ExecuteWithTimeoutSilently(device_connectivity.TimeOutBlockDevCmd, device_connectivity.BlockDevCmd, args)
```
oriyarde, line 562: why silently?

node/pkg/driver/node_test.go
```diff
@@ -895,23 +896,129 @@ func TestNodeUnpublishVolume(t *testing.T) {
...
+		t.Run(tc.name, tc.testFunc)
 	}
-	if srvErr.Code() != expErrCode {
-		t.Fatalf("Expected error code %d, got %d message %s", expErrCode, srvErr.Code(), srvErr.Message())
+}
+
+func assertExpectedStats(t *testing.T, expResp *csi.NodeGetVolumeStatsResponse, req *csi.NodeGetVolumeStatsRequest, node driver.NodeService) {
+	resp, err := node.NodeGetVolumeStats(context.TODO(), req)
+	if err != nil {
+		t.Fatalf("Expect no error but got: %v", err)
```
oriyarde, line 1018:
```suggestion
		t.Fatalf("Expected no error but got: %v", err)
```

node/pkg/driver/node_utils.go
```diff
@@ -515,3 +519,60 @@ func (n NodeUtils) GetTopologyLabels(ctx context.Context, nodeName string) (map[
...
+	if err != nil {
+		return VolumeStatistics{}, err
```
oriyarde, line 564: can we return `nil` instead of `VolumeStatistics{}`?

node/pkg/driver/node_test.go
matancarmeli7: done CSI-3947

node/pkg/driver/node_utils.go
```diff
@@ -515,3 +519,60 @@ func (n NodeUtils) GetTopologyLabels(ctx context.Context, nodeName string) (map[
...
+	}
+	
+	strOut := strings.TrimSpace(string(out))
+	gotSizeBytes, err := strconv.ParseInt(strOut, 10, 64)
```
oriyarde, line 568:
```suggestion
	sizeInBytes, err := strconv.ParseInt(strOut, 10, 64)
```

node/pkg/driver/node_utils.go
matancarmeli7: I don't want that every that on every `grpc` call from the `kubelet` we will see all the information, it won't mess up the logs?

node/pkg/driver/node_utils.go
matancarmeli7: the function returns object from VolumeStatistics type, how can I do it with nil without changing the return objects type

node/pkg/driver/node_utils.go
oriyarde: it doesn't seem like a lot of information. what is the frequency we see per volume?

node/pkg/driver/node_utils.go
oriyarde: oh, it's not a pointer, I see

node/pkg/driver/node.go
```diff
@@ -502,6 +508,14 @@ func (d *NodeService) publishRawBlockVolume(mpathDevice string, targetPath strin
...
 func (d *NodeService) isTargetMounted(targetPathWithHostPrefix string, isFSVolume bool) (bool, error) {
 	logger.Debugf("Check if target {%s} is mounted", targetPathWithHostPrefix)
+	isMounted, err := d.IsMounted(targetPathWithHostPrefix, isFSVolume)
+	if isMounted && err == nil {
+		logger.Warningf("Idempotent case : targetPath already mounted (%s), so no need to mount again. Finish NodePublishVolume", targetPathWithHostPrefix)
+	}
+	return isMounted, err
+}
+
+func (d *NodeService) IsMounted(targetPathWithHostPrefix string, isFSVolume bool) (bool, error) {
```
oriyarde, line 518: do we still need this?

matancarmeli7: I think it's better to separate it that every function will do its own purpose, if you think it's not necessary I can roll it back to his older version

node/pkg/driver/node_utils.go
matancarmeli7: done

node/pkg/driver/node_test.go
matancarmeli7: done

node/pkg/driver/node.go
oriyarde: let's keep the diff minimal and roll it back for now

matancarmeli7: done

node/pkg/driver/node.go
matancarmeli7: done

node/pkg/driver/node_utils.go
matancarmeli7: about 30 seconds or less

node/pkg/driver/node.go
```diff
@@ -501,6 +507,7 @@ func (d *NodeService) publishRawBlockVolume(mpathDevice string, targetPath strin
 // isFSVolume: if we check volume with file system - true, otherwise for raw block false
 // Returns: is <target mounted, error if occured>
 func (d *NodeService) isTargetMounted(targetPathWithHostPrefix string, isFSVolume bool) (bool, error) {
+	logger.Debugf("Check if target {%s} is mounted", targetPathWithHostPrefix)
 	logger.Debugf("Check if target {%s} is mounted", targetPathWithHostPrefix)
```
oriyarde, line 511: ?

node/pkg/driver/node_utils.go
oriyarde: got it

node/pkg/driver/node.go
matancarmeli7: oh sorry done

node/pkg/driver/node.go
```diff
@@ -613,9 +619,75 @@ func (d *NodeService) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 }
 
 func (d *NodeService) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVolumeStatsRequest) (*csi.NodeGetVolumeStatsResponse, error) {
-	goid_info.SetAdditionalIDInfo(req.VolumeId)
+	volumeId := req.VolumeId
+	goid_info.SetAdditionalIDInfo(volumeId)
 	defer goid_info.DeleteAdditionalIDInfo()
-	return nil, status.Error(codes.Unimplemented, "NodeGetVolumeStats is not implemented yet")
+	volumePath := req.VolumePath
+	volumePathWithHostPrefix := d.NodeUtils.GetPodPath(req.VolumePath)
```
oriyarde, line 626:
```suggestion
	volumePathWithHostPrefix := d.NodeUtils.GetPodPath(volumePath)
```

node/pkg/driver/node.go
```diff
@@ -613,9 +619,75 @@ func (d *NodeService) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+		return status.Error(codes.InvalidArgument, "NodeGetVolumeStats Volume ID must be provided")
+	}
+	if volumePath == "" {
+		return status.Error(codes.InvalidArgument, "NodeGetVolumeStats Volume Path must be provided")
+	}
+
+	return nil
+}
+
+func (d *NodeService) GetVolumeStats(path string, volumeId string) (VolumeStatistics, error) {
```
oriyarde, line 671:
```suggestion
func (d *NodeService) getVolumeStats(path string, volumeId string) (VolumeStatistics, error) {
```

node/pkg/driver/node_test.go
```diff
@@ -895,23 +896,129 @@ func TestNodeUnpublishVolume(t *testing.T) {
...
+		},
+		{
+			name: "fail to get stats",
+			testFunc: func(t *testing.T) {
+				expErrCode := codes.Internal
+				expSubString := "Failed to get statistics"
+				mockNodeUtils.EXPECT().GetPodPath(volumePath).Return(volumePathWithHostPrefix)
+				mockNodeUtils.EXPECT().IsPathExists(volumePathWithHostPrefix).Return(true)
+				mockNodeUtils.EXPECT().IsBlock(volumePathWithHostPrefix).Return(false, nil)
+				mockNodeUtils.EXPECT().GetFileSystemVolumeStats(volumePathWithHostPrefix).Return(driver.VolumeStatistics{}, status.Errorf(codes.Internal, "fail to get stats"))
```
oriyarde, line 938: `status` is part of grpc and node utils should not return grpc errors

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
```diff
@@ -75,7 +75,7 @@ const (
 	IscsiHostRexExPath          = "/sys/class/iscsi_host/host*/device/session*/iscsi_session/session*/targetname"
 	sysDeviceSymLinkFormat      = "/sys/block/%s/device"
 	sysDeviceDeletePathFormat   = sysDeviceSymLinkFormat + "/delete"
-	blockDevCmd                 = "blockdev"
+	BlockDevCmd                 = "blockdev"
 	flushBufsFlag               = "--flushbufs"
 	mpathdSeparator             = ","
 	multipathdCmd               = "multipathd"
```
oriyarde, line 81: it seems `multipathdCmd` is duplicated, I suggest we do the same for `blockdevCmd`, at least until we find a common place for these

node/pkg/driver/node.go
matancarmeli7: done

node/pkg/driver/node.go
matancarmeli7: done

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
matancarmeli7: done

oriyarde: I still see a diff in this file

node/pkg/driver/node.go
oriyarde: I don't see any change

node/pkg/driver/node.go
oriyarde: nope

node/pkg/driver/node_test.go
matancarmeli7: done

node/pkg/driver/node_test.go
```diff
@@ -895,23 +896,129 @@ func TestNodeUnpublishVolume(t *testing.T) {
 }
 
 func TestNodeGetVolumeStats(t *testing.T) {
+	volId := "someStorageType:vol-test"
```
oriyarde, line 899:
```suggestion
	volumeId := "someStorageType:vol-test"
```

node/pkg/driver/node_test.go
oriyarde: I don't see any change

node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
matancarmeli7: done now

node/pkg/driver/node_test.go
matancarmeli7: done now

node/pkg/driver/node.go
matancarmeli7: done now

node/pkg/driver/node.go
matancarmeli7: done now

node/pkg/driver/node_test.go
matancarmeli7: done

node/pkg/driver/node.go
```diff
@@ -613,9 +619,75 @@ func (d *NodeService) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+	if err != nil {
+		return VolumeStatistics{}, status.Errorf(codes.Internal, "failed to determine if %q is block device: %s", path, err)
+	}
+
+	if isBlock {
+		mpathDevice, err := d.OsDeviceConnectivityHelper.GetMpathDevice(volumeId)
+		if err != nil {
+			switch err.(type) {
+			case *device_connectivity.MultipathDeviceNotFoundForVolumeError:
+				return VolumeStatistics{}, status.Errorf(codes.NotFound, "Multipath Device of volume id %q does not exist", volumeId)
```
oriyarde, line 682: wouldn't kubelet still get an internal error [here](https://github.com/IBM/ibm-block-csi-driver/blob/00ee18991eb5c624f26c6c54f8b694fc98d88be5/node/pkg/driver/node.go#L640)?

matancarmeli7: yes you right, I will move this function to the `NodeGetVolumeStats` function

oriyarde: another option is to have `getVolumeStats` return only grpc errors, so that `NodeGetVolumeStats` can simply return the error

matancarmeli7: yes yes I thought about it later and wanted to do like you suggested

matancarmeli7: done

node/pkg/driver/node.go
```diff
@@ -613,9 +619,82 @@ func (d *NodeService) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+				return VolumeStatistics{}, status.Errorf(codes.Internal, "%q error while discovering the device : %s", errorMsg, err)
+			}
+		}
+		volumeStats, err = d.NodeUtils.GetBlockVolumeStats(mpathDevice)
+	} else {
+		volumeStats, err = d.NodeUtils.GetFileSystemVolumeStats(path)
+	}
+
+	if err != nil {
+		return VolumeStatistics{}, status.Errorf(codes.Internal, "%q failed to determine if %q is block device: %s", errorMsg, path, err)
```
oriyarde, line 695: if node utils fail, it doesn't mean it failed to determine if the path is block device

node/pkg/driver/node.go
```diff
@@ -613,9 +619,82 @@ func (d *NodeService) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+	}
+
+	if isBlock {
+		mpathDevice, err := d.OsDeviceConnectivityHelper.GetMpathDevice(volumeId)
+		if err != nil {
+			switch err.(type) {
+			case *device_connectivity.MultipathDeviceNotFoundForVolumeError:
+				return VolumeStatistics{}, status.Errorf(codes.NotFound, "%q multipath Device of volume id %q does not exist", errorMsg, volumeId)
```
oriyarde, line 684:
```suggestion
				return VolumeStatistics{}, status.Errorf(codes.NotFound, "%q multipath device of volume id %q does not exist", errorMsg, volumeId)
```

node/pkg/driver/node.go
matancarmeli7: I wanted that it will tell us the function that the error was in, if you think it's not necessary I can remove it

oriyarde: the error could come from `GetBlockVolumeStats`/`GetFileSystemVolumeStats`, so I think it's misleading

matancarmeli7: ok I will remove it

node/pkg/driver/node.go
matancarmeli7: done

node/pkg/driver/node.go
matancarmeli7: done

oriyarde (approved):
LGTM GJ

ArbelNathan (approved):
LGTM

node/pkg/driver/node.go
```diff
@@ -613,9 +619,81 @@ func (d *NodeService) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 }
 
 func (d *NodeService) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVolumeStatsRequest) (*csi.NodeGetVolumeStatsResponse, error) {
-	goid_info.SetAdditionalIDInfo(req.VolumeId)
+	volumeId := req.VolumeId
+	goid_info.SetAdditionalIDInfo(volumeId)
 	defer goid_info.DeleteAdditionalIDInfo()
-	return nil, status.Error(codes.Unimplemented, "NodeGetVolumeStats is not implemented yet")
```
oriyarde, line 618: what about enabling the [community `NodeGetVolumeStats` tests](https://github.com/IBM/ibm-block-csi-driver/blob/275618b2b53acf9c72e57f5ad30a89153801d1a6/scripts/csi_test/common_csi_tests_to_skip#L5)?

matancarmeli7: https://github.com/IBM/ibm-block-csi-driver/pull/480

PR #459, opened 2022-02-28
Feature/csi 3847 call home scripts

ArbelNathan:


oriyarde (approved):
approving for pen test, will actually be reviewed in a following PR

PR #474, opened 2022-03-13
fix CSI-2155 spec gap capacity_bytes

ArbelNathan:
[spec](https://github.com/container-storage-interface/spec/blob/v1.5.0/spec.md#createvolume): Volume MUST be at least this big.


oriyarde (approved):
LGTM (for 1.10.0)

controller/controller_server/csi_controller_server.py
```diff
@@ -97,7 +97,7 @@ def CreateVolume(self, request, context):
                 else:
                     logger.debug("volume found : {}".format(volume))
 
-                    if not source_id and volume.capacity_bytes != request.capacity_range.required_bytes:
+                    if not source_id and volume.capacity_bytes < request.capacity_range.required_bytes:
                         message = "Volume was already created with different size."
```
oriyarde, line 101: btw, regardless of this PR, I think it would be better to include the numbers in the error message

controller/controller_server/csi_controller_server.py
```diff
@@ -97,8 +97,11 @@ def CreateVolume(self, request, context):
                 else:
                     logger.debug("volume found : {}".format(volume))
 
-                    if not source_id and volume.capacity_bytes != request.capacity_range.required_bytes:
-                        message = "Volume was already created with different size."
+                    volume_capacity_bytes = volume.capacity_bytes
+                    if not source_id and volume_capacity_bytes < required_bytes:
```
oriyarde, line 101: not necessarily for this PR, but what if there is a `source_id` in the request?
should we handle it in a different ticket?

ArbelNathan: its handled at the mediator lvl.. resizing and such..

oriyarde: I don't see where/why the mediator should handle `ALREADY_EXISTS` when the request has a `source_id` and the existing volume size is smaller than requested

ArbelNathan: amm, why do we even check that `not source_id`?

oriyarde: by using "git blame", I found the [relevant conversation](https://github.com/IBM/ibm-block-csi-driver/pull/158#discussion_r433097034):
>Only in case of create from snapshot size can change

so maybe it does have something to do with resize?

ArbelNathan: size can be lower? why? we create volume by its own parameters and then handle the copy..

oriyarde: maybe the copy changes the size (on a9k)?

ArbelNathan: the check is before the copy..

oriyarde: let's deal with it in CSI-1509

oriyarde (approved):
LGTM

PR #483, opened 2022-03-27
CSI-1509 copy def to abstract

ArbelNathan:
[origin](https://github.com/IBM/ibm-block-csi-driver/pull/158#discussion_r432929283)

### DIFF
--- a/controller/array_action/array_mediator_abstract.py
+++ b/controller/array_action/array_mediator_abstract.py
@@ -72,3 +72,30 @@ class ArrayMediatorAbstract(ArrayMediator, ABC):
         if connectivity_type == NVME_OVER_FC_CONNECTIVITY_TYPE:
             vol_id = convert_scsi_id_to_nguid(vol_id)
         self.unmap_volume(vol_id, host_name)
+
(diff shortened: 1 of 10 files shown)


oriyarde (approved):
LGTM

controller/array_action/array_mediator_abstract.py
```diff
@@ -72,3 +72,30 @@ def unmap_volume_by_initiators(self, vol_id, initiators):
         if connectivity_type == NVME_OVER_FC_CONNECTIVITY_TYPE:
             vol_id = convert_scsi_id_to_nguid(vol_id)
         self.unmap_volume(vol_id, host_name)
+
+    def copy_to_existing_volume_from_source(self, volume, source_id, source_type, minimum_volume_size):
```
matancarmeli7, line 76: I know that you only copied this function, but can you refactor this function that it will do only one thing? divide it into meaningful functions that it calls them

ArbelNathan: the cost is higher than the value.
@oriyarde what do you think?

PR #494, opened 2022-04-24
CSI-4114 improve svc controller publish volume

oriyarde:


controller/array_action/array_mediator_svc.py
```diff
@@ -732,8 +732,8 @@ def _get_host_ports(self, host, attribute_name):
         ports = host.get(attribute_name, [])
         return ports if isinstance(ports, list) else [ports]
 
-    def get_host_by_host_identifiers(self, initiators):
-        logger.debug("Getting host name for initiators : {0}".format(initiators))
+    def _get_host_by_host_identifiers_slow(self, initiators):
+        logger.debug("Getting host name slowly for initiators : {0}".format(initiators))
```
ArbelNathan, line 736: it sounds like we are laughing at ourselves..

oriyarde: I just want an easy way to distinguish between the short and the long way

ArbelNathan: I think the method name is enough.. (it will be shown in logs)

oriyarde: what will be shown in the logs?

ArbelNathan: the `slow` in the method name

oriyarde: where? I don't see it in the logs (only the string sent to `logger.debug`)

oriyarde: oh, I see it now

oriyarde: done

ArbelNathan (approved):
LGTM

controller/array_action/array_mediator_svc.py
oriyarde: could you resolve this conversation?

controller/array_action/array_mediator_svc.py
```diff
@@ -764,6 +764,65 @@ def get_host_by_host_identifiers(self, initiators):
...
+
+    def get_host_by_host_identifiers(self, initiators):
+        logger.debug("Getting host name for initiators : {0}".format(initiators))
+        host_names = set()
+        connectivity_types = set()
+        for connectivity_type, initiator in initiators:
+            if connectivity_type == config.NVME_OVER_FC_CONNECTIVITY_TYPE:
+                nvme_host_names = self._get_host_names_by_nqn(initiator)
+                if nvme_host_names:
+                    host_names.update(nvme_host_names)
```
matancarmeli7, line 807: why are you adding it to a set if it's already returned as a set?

controller/array_action/array_mediator_svc.py
```diff
@@ -764,6 +764,65 @@ def get_host_by_host_identifiers(self, initiators):
...
+        iscsi_login = self._lshostiplogin(iqn)
+        if iscsi_login:
+            return iscsi_login.host_name
+        return None
+
+    def get_host_by_host_identifiers(self, initiators):
+        logger.debug("Getting host name for initiators : {0}".format(initiators))
+        host_names = set()
+        connectivity_types = set()
+        for connectivity_type, initiator in initiators:
```
matancarmeli7, line 803: can you move this for into a dedicated function?

controller/array_action/array_mediator_svc.py
```diff
@@ -764,6 +764,65 @@ def get_host_by_host_identifiers(self, initiators):
...
+    def _get_host_name_by_iqn(self, iqn):
+        iscsi_login = self._lshostiplogin(iqn)
+        if iscsi_login:
+            return iscsi_login.host_name
+        return None
+
+    def get_host_by_host_identifiers(self, initiators):
+        logger.debug("Getting host name for initiators : {0}".format(initiators))
+        host_names = set()
+        connectivity_types = set()
```
matancarmeli7, line 802: why do we need to use it? and add to it a value in the if statements? 
I think we just can use the `connectivity_type` value, because if it's `iscsi`, `nvme` or `fc`, so it will be good and if it isn't one of them, we won't use this value anyway after the if statments

controller/array_action/array_mediator_svc.py
oriyarde: sure, how should I call it?

controller/array_action/array_mediator_svc.py
oriyarde: > we won't use this value anyway after the if statments

we return `connectivity_types`, because according to the current mediator interface, this method returns all the connectivity types it found for the host definition.
so I'm not sure what you mean

controller/array_action/array_mediator_svc.py
oriyarde: `host_names` is a set of all the host names found with relevant initiators defined (for every worker connectivity type).
what's wrong with adding a set to a set? what are you suggesting to do?

controller/array_action/array_mediator_svc.py
matancarmeli7: maybe `_update_host_names_and_connectivity_types`? there is a beeter name that you can think of?

controller/array_action/array_mediator_svc.py
matancarmeli7: never mind my bad I missed

controller/array_action/array_mediator_svc.py
matancarmeli7: never mind my bad I missed

controller/tests/array_action/svc/array_mediator_svc_test.py
```diff
@@ -734,7 +734,49 @@ def test_properties(self):
...
+                                  config.FC_CONNECTIVITY_TYPE,
+                                  config.ISCSI_CONNECTIVITY_TYPE}
+
+        self.svc.client.svcinfo.lsnvmefabric.return_value = Mock(as_list=nvme_host_mocks)
+        if config.NVME_OVER_FC_CONNECTIVITY_TYPE in connectivity_types:
+            nvme_host_names = host_names if nvme_host_names is None else nvme_host_names
+            if nvme_host_names:
+                nvme_host_mocks = [Mock(object_name=host_name) for host_name in nvme_host_names]
+                lsnvmefabric_return_values = [Mock(as_list=[host_mock]*4) for host_mock in nvme_host_mocks]
+                self.svc.client.svcinfo.lsnvmefabric.side_effect = lsnvmefabric_return_values
```
matancarmeli7, line 754: can you create this code in a function? `_prepare_mocks_for_get_nvme_host`
and for [fc](https://github.com/IBM/ibm-block-csi-driver/blob/ab486a6bdb3520f6fed53d00cd63895d7948aa8d/controller/tests/array_action/svc/array_mediator_svc_test.py#L756) and [iscsi](https://github.com/IBM/ibm-block-csi-driver/blob/ab486a6bdb3520f6fed53d00cd63895d7948aa8d/controller/tests/array_action/svc/array_mediator_svc_test.py#L767) accordingly

controller/array_action/array_mediator_svc.py
oriyarde: done

controller/tests/array_action/svc/array_mediator_svc_test.py
oriyarde: done

matancarmeli7 (approved):
LGTM

ArbelNathan (dismissed):
LGTM

ArbelNathan (approved):
lgtm

matancarmeli7 (approved):

PR #500, opened 2022-05-24
add svc volume group storageclass parameter

oriyarde:


controller/array_action/array_mediator_ds8k.py
```diff
@@ -266,7 +266,7 @@ def _create_api_volume(self, name, size_in_bytes, array_space_efficiency, pool_i
             )
             raise array_errors.VolumeCreationError(name)
 
-    def create_volume(self, name, size_in_bytes, space_efficiency, pool, io_group):
+    def create_volume(self, name, size_in_bytes, space_efficiency, pool, io_group, volume_group):
```
matancarmeli7, line 269: don't it better to add a default to `volume_group` parameter in order to avoid `None` passes like [here](https://github.com/IBM/ibm-block-csi-driver/blob/7610c0643d2d44d7f3dfdeeaf04b7ba913929639/controller/tests/array_action/ds8k/test_array_mediator_ds8k.py#L198)?

matancarmeli7: never mind

matancarmeli7 (approved):
LGTM

controller/tests/controller_server/csi_controller_server_test.py
```diff
@@ -686,7 +686,7 @@ def test_create_volume_with_required_bytes_zero(self, storage_agent):
 
         self.servicer.CreateVolume(self.request, self.context)
         self.assertEqual(self.context.code, grpc.StatusCode.OK)
-        self.mediator.create_volume.assert_called_once_with(self.request.name, 2, None, POOL, None)
+        self.mediator.create_volume.assert_called_once_with(self.request.name, 2, None, POOL, None, None)
```
ArbelNathan, line 689: can you please add UT with volume_group?

controller/array_action/array_mediator_svc.py
```diff
@@ -101,6 +101,8 @@ def build_kwargs_from_parameters(space_efficiency, pool_name, io_group, volume_n
     cli_kwargs.update(space_efficiency_kwargs)
     if io_group:
         cli_kwargs['iogrp'] = io_group
+    if volume_group:
+        cli_kwargs['volumegroup'] = volume_group
```
ArbelNathan, line 105: no new expected errors? we want to say if the volumegroup does not exist..

controller/tests/controller_server/csi_controller_server_test.py
oriyarde: done

controller/array_action/array_mediator_svc.py
oriyarde: 1. how is it different than https://github.com/IBM/ibm-block-csi-driver/pull/448?
2. at least in svc - pool, io_group and volume_group could all cause the same error ([CMMVC5754E](https://github.com/IBM/ibm-block-csi-driver/blob/v1.9.0/controller/array_action/array_mediator_svc.py#L24)), so should we add more commands inside the expect, to keep this resolution? or should the mediator interface just not require such granularity?

ArbelNathan: I'm not familiar with the possible exception, if you think we need no new error codes so I will resolve it.

oriyarde: I could change the existing "pool doesn't exist" error to a more general "invalid argument" error.
what do you think?

ArbelNathan: if it's indeed the same `error code` so yes, please do it.

oriyarde: done

ArbelNathan (approved):
LGTM

matancarmeli7 (approved):

PR #501, opened 2022-05-26
assertion order

ArbelNathan:


controller/tests/array_action/test_storage_agent.py
```diff
@@ -79,32 +79,32 @@ def tearDown(self):
...
-        self.assertEqual(len(get_agents()), 0)
+        self.assertEqual(0, len(get_agents()))
         agent = get_agent(ArrayConnectionInfo(array_addresses=endpoints, user=name, password=password))
-        self.assertEqual(len(get_agents()), 1)
+        self.assertEqual(1, len(get_agents()))
         new_agent = get_agent(ArrayConnectionInfo(array_addresses=endpoints, user=name, password=password))
-        self.assertEqual(len(get_agents()), 1)
-        self.assertEqual(id(agent), id(new_agent))
+        self.assertEqual(1, len(get_agents()))
+        self.assertEqual(id(new_agent), id(agent))
```
oriyarde, line 98: isn't the "existing" (original) the "expected" one?

ArbelNathan: both are tested here, non of them is mocked so I guessed it doesn't matter.. I figured because `agent` is the first one to be created so I figured `new_agent` is the one we are using for comparison..

oriyarde: from the test name "test_get_agent_return_existing", it sounds to me like the test asserts, that the actual returned value is the existing one, i.e. it expects the actual to be the existing. in other words, the existing is the expected.
and I don't see how the second agent (before knowing whether they are equal) could be considered the existing one

ArbelNathan: done

oriyarde (approved):
LGTM

matancarmeli7 (approved):
LGTM

PR #522, opened 2022-06-21
CSI-4369 fix sg_inq timeout

oriyarde:
it [seems](https://github.com/hreinecke/sg3_utils/blob/v1.46/src/sg_inq.c#L140) the internal timeout of the command is 60 seconds, IMO we can wait at least 10 seconds, in case there are network delays


matancarmeli7 (approved):
LGTM

PR #541, opened 2022-07-14
Handle ExpectedSnapshotButFoundVolumeError in createVolume function

ELENAGER:
… it inside volume struct


controllers/array_action/array_action_types.py
```diff
@@ -13,6 +13,7 @@ class Volume:
     source_id: str
     array_type: str
     pool: str
+    source_type: str
```
oriyarde, line 16: I would put it next to `source_id`

controllers/servers/csi/csi_controller_server.py
```diff
@@ -93,7 +93,7 @@ def CreateVolume(self, request, context):
                         utils.validate_parameters_match_source_volume(space_efficiency, required_bytes, source_volume)
 
                 try:
-                    volume = array_mediator.get_volume(volume_final_name, pool, is_virt_snap_func)
+                    volume = array_mediator.get_volume(volume_final_name, pool, source_type, is_virt_snap_func)
```
oriyarde, line 96: I think the point here is to compare between the k8s-requested source type and the actual source type on storage.
currently in this PR, the server only passes the requested source type, and gets the same value back as-is, from the mediator.
so this PR currently lacks:

1. the mediators logic to discover the actual source on storage
2. the comparison between requested and actual

Ideally, the mediators shouldn't know the requested source type, i.e. the comparison should be done outside the mediators, in e.g. a server private/utils function

HIH

controllers/array_action/array_action_types.py
ELENAGER: Agreed

ELENAGER: Moved next to source_id

oriyarde (approved):
LGTM

ArbelNathan (approved):
LGTM

PR #543, opened 2022-07-18
Validate source type when creating volume from source and raise excer…

ELENAGER:
While creating volume, source type must be only SNAPSHOT_TYPE_NAME or VOLUME_TYPE_NAME
https://github.com/container-storage-interface/spec/blob/v1.5.0/spec.md#createvolume - see VolumeContentSource.


controllers/servers/messages.py
```diff
@@ -18,6 +18,7 @@
 UNSUPPORTED_FS_TYPE_MESSAGE = "unsupported fs_type : {}"
 UNSUPPORTED_MOUNT_FLAGS_MESSAGE = "mount_flags is unsupported"
 UNSUPPORTED_VOLUME_ACCESS_TYPE_MESSAGE = "unsupported volume access type"
+UNSUPPORTED_VOLUME_SOURCE_TYPE_MESSAGE = "unsupported volume source type"
```
ArbelNathan, line 21:
```suggestion
UNSUPPORTED_VOLUME_SOURCE_TYPE_MESSAGE = "unsupported volume source type : {}"
```
so we will have in the log the "forbidden" type..

ELENAGER: Added the wrong type to an error message inside exception

oriyarde (approved):
LGTM

ArbelNathan (approved):
LGTM

PR #586, opened 2022-11-02
CSI 4731 servicer delete vg implementation

ArbelNathan:


controllers/servers/csi/csi_controller_server.py
```diff
@@ -595,7 +595,30 @@ def CreateVolumeGroup(self, request, context):
...
+        secrets = request.secrets
+        utils.validate_delete_volume_group_request(request)
+
+        try:
+            volume_group_id_info = utils.get_volume_group_id_info(request.volume_group_id)
+        except ObjectIdError as ex:
+            logger.warning("volume group id is invalid. error : {}".format(ex))
+            return csi_pb2.DeleteVolumeGroupResponse()
+
+        array_type = volume_group_id_info.array_type
```
matancarmeli7, line 607: I think this can be redundant if we are using `array_type` in only one place

controllers/servers/csi/csi_controller_server.py
```diff
@@ -595,7 +595,30 @@ def CreateVolumeGroup(self, request, context):
...
+        volume_group_id = volume_group_id_info.ids.uid
+        array_connection_info = utils.get_array_connection_info_from_secrets(secrets)
+
+        with get_agent(array_connection_info, array_type).get_mediator() as array_mediator:
+            logger.debug(array_mediator)
```
matancarmeli7, line 612: do we want to expose it without any additional content?

ArbelNathan: sure, we are doing it for most methods

matancarmeli7: so why do we do it in some places, and in some places not? 
and it not better to just put it in the `get_mediator` function?

ArbelNathan: Im not sure why, and maybe.. would you like me to open a ticket for it?

matancarmeli7: it will reduce all the places with `logger.debug(array_mediator)`, yes open it thanks

matancarmeli7 (approved):
LGTM

ELENAGER (approved):
LGTM

PR #592, opened 2022-11-06
CSI 4731 servicer modify vg implementation

ArbelNathan:


controllers/servers/csi/csi_controller_server.py
```diff
@@ -620,6 +620,57 @@ def DeleteVolumeGroup(self, request, _):
...
+
+            volume_group = self._get_volume_group(array_mediator, volume_group_id)
+
+            volumes_in_volume_group = volume_group.volumes
+            volume_ids_in_volume_group = [volume.id for volume in volumes_in_volume_group]
+            volume_ids_in_request = []
+            for volume_id in request.volume_ids:
+                volume_id_info = utils.get_volume_id_info(volume_id)
+                volume_ids_in_request.append(volume_id_info.ids.uid)
+
```
ELENAGER, line 667: Is it possible to add functions like 
volume_ids_in_request =build_volume_ids_in_request(request) 
and 
volume_ids_in_volume_group = build_volume_ids_in_group(volume_group)?

ArbelNathan: for what reason? how is it different?

ELENAGER: I think it will be cleaner to move it into separate function. Why not?

matancarmeli7: I agree with @ELENAGER

ArbelNathan: I really don't understand, how is it gonna look?

ELENAGER: Something like this:
volume_ids_in_request = self._build_volume_ids_in_request(request)

ArbelNathan: for where?
what in _build_volume_ids_in_request?
what line is it even for?
what is it building?

matancarmeli7: I think we can create two functions:
```
def _get_volume_ids_in_request(self, request):
volume_ids_in_request = []
for volume_id in request.volume_ids:
    volume_id_info = utils.get_volume_id_info(volume_id)
    volume_ids_in_request.append(volume_id_info.ids.uid)
return volume_ids_in_request
```

```
def _get_volume_ids_in_volume_group(self, volume_group):
volumes_in_volume_group = volume_group.
return [volume.id for volume in volumes_in_volume_group]

```

matancarmeli7: and in the `ModifyVolumeGroup` function, we will have two lines that each of them calls each function

ArbelNathan: tnx, tnx

matancarmeli7 (approved):
LGTM

PR #597, opened 2022-11-24
Task/csi 4885 ear on volume in group

ELENAGER:


matancarmeli7 (approved):
LGTM

controllers/servers/csi/addons_server.py
```diff
@@ -156,3 +157,14 @@ def _ensure_replication_idempotency(replication_request, replication, volume):
...
+    @staticmethod
+    def _get_and_validate_volume(mediator, volume_id):
+        volume = mediator.get_object_by_id(volume_id, servers_settings.VOLUME_TYPE_NAME)
+        if not volume:
+            raise array_errors.ObjectNotFoundError(volume_id)
+        if volume.volume_group_id:
+            error_message = "could not enable replication, " \
+                      "volume {} already belongs to volume group {}".format(volume.name, volume.volume_group_id)
+            return volume, error_message
+        return volume, None
```
ArbelNathan, line 170:
```suggestion
        return volume, ""
```

controllers/servers/csi/addons_server.py
```diff
@@ -27,13 +27,14 @@ def EnableVolumeReplication(self, request, context):
 
         connection_info = utils.get_array_connection_info_from_secrets(request.secrets)
         with get_agent(connection_info, volume_id_info.array_type).get_mediator() as mediator:
-            volume = mediator.get_object_by_id(volume_id, servers_settings.VOLUME_TYPE_NAME)
-            if not volume:
-                raise array_errors.ObjectNotFoundError(volume_id)
+            volume, error_message = self._get_and_validate_volume(mediator, volume_id)
```
ArbelNathan, line 30: `and` indicates that we are not keeping SOLID.. please split it

controllers/servers/csi/addons_server.py
ELENAGER: I'd rather prefer to return None as no-error-message. Especially because the function _ensure_replication_idempotency returns None in the similar case. And I want to be consistent

controllers/servers/csi/addons_server.py
ELENAGER: Done

matancarmeli7 (approved):
LGTM

PR #626, opened 2022-12-14
When disk is in EAR group it has 2 fc mappings. This case must not prevent snapshot creation

ELENAGER:
…ot prevent snapshot creating


controllers/array_action/array_mediator_svc.py
```diff
@@ -55,6 +55,7 @@
 HOSTS_LIST_ERR_MSG_MAX_LENGTH = 300
```
matancarmeli7, line 55: do we need a unit test for it?

ELENAGER: I added two unit tests - success for EAR volume with 2 fc mappings and failure for EAR volume with another number.

matancarmeli7 (dismissed):
LGTM

matancarmeli7 (approved):

PR #628, opened 2022-12-15
CSI 4921 support port files comments

ArbelNathan:


node/pkg/driver/node_utils.go
```diff
@@ -17,6 +17,7 @@
 package driver
```
matancarmeli7, line 17: can you add a unit test for it?

ELENAGER (approved):
LGTM

node/pkg/driver/node_utils.go
ArbelNathan: done

ELENAGER (approved):

matancarmeli7 (approved):
LGTM

PR #630, opened 2022-12-20
Feature/csi 4744 vg replication addon implementation

ELENAGER:


controllers/servers/utils.py
```diff
@@ -742,3 +767,21 @@ def validate_modify_volume_group_request(request):
...
+    if object_info:
+        logger.info(object_info)
+        if object_info.HasField(servers_settings.VOLUME_TYPE_NAME):
+            object_id = object_info.volume.volume_id
+            object_type = servers_settings.VOLUME_TYPE_NAME
+        elif object_info.HasField(servers_settings.VOLUME_GROUP_TYPE_NAME):
+            object_id = object_info.volume.volume_group_id
+            object_type = servers_settings.VOLUME_GROUP_TYPE_NAME
+        else:
+            return object_type, object_id
```
matancarmeli7, line 785: you set object_info as None, so it will ever continue to the `if` statement?

ELENAGER: Yes, I know. Until yesterday we had no object_info inside the request, now I updated spec and will add the real object_info

controllers/servers/utils.py
```diff
@@ -770,16 +770,17 @@ def validate_modify_volume_group_request(request):
 
 
 def get_replication_object_type_and_id_info(request):
+    object_id = request.volume_id
+    object_type = servers_settings.VOLUME_TYPE_NAME
```
ArbelNathan, line 774: why is this needed?

controllers/servers/utils.py
```diff
@@ -770,16 +770,17 @@ def validate_modify_volume_group_request(request):
...
-        if replication_source.HasField(servers_settings.VOLUME_TYPE_NAME):
-            object_id = replication_source.volume.replication_volume_id
-            object_type = servers_settings.VOLUME_TYPE_NAME
-        elif replication_source.HasField(servers_settings.VOLUME_GROUP_TYPE_NAME):
+        if replication_source.HasField(servers_settings.VOLUME_GROUP_TYPE_NAME):
             object_id = replication_source.volumegroup.replication_volume_group_id
             object_type = servers_settings.VOLUME_GROUP_TYPE_NAME
-        else:
-            return servers_settings.VOLUME_TYPE_NAME, request.volume_id
+        elif replication_source.HasField(servers_settings.VOLUME_TYPE_NAME):
```
ArbelNathan, line 782: can you add `else` and raise some `not supported` error?

controllers/scripts/csi_general/csi_pb2.sh
```diff
@@ -8,8 +8,8 @@ PB2_DIR="csi_general"
 mkdir -p ./proto/${PB2_DIR}
 cd ./proto/${PB2_DIR}
 
-curl -O https://raw.githubusercontent.com/container-storage-interface/spec/${CSI_VERSION}/csi.proto
-curl -O https://raw.githubusercontent.com/csi-addons/spec/${ADDONS_VERSION}/replication.proto
+curl -O https://raw.githubusercontent.com/IBM/csi-volume-group/add/CSI-5164_add_all_spec_with_vg/csi.proto
+curl -O https://raw.githubusercontent.com/ELENAGER/spec/from_tag_v0.1.1/replication.proto
```
ArbelNathan, line 12: you cant submit PR with non-official versions

controllers/servers/utils.py
ELENAGER: It's a default case for backward compatibility. In our replication side car a new field for volume group will arrive and handled in switch-case below

controllers/servers/utils.py
ELENAGER: Done

controllers/scripts/csi_general/csi_pb2.sh
ArbelNathan: ok lets leave it like that and make sure it changes before GA

ArbelNathan (approved):
LGTM

matancarmeli7 (approved):
LGTM

PR #633, opened 2022-12-26
make sure all that full io_group is the default

matancarmeli7:
Signed-off-by: matancarmeli7 <matan.carmeli7@gmail.com>


ELENAGER (approved):
LGTM

PR #639, opened 2023-01-08
CSI-5192 uid to name in volume_group_id

ArbelNathan:
now volumeGroupHandle will look like SVC:3;CSI_volumegroup-d2cf8cc7-93b9-4303-b850-079c66525bf0
instead of SVC:3;3


controllers/array_action/array_mediator_svc.py
```diff
@@ -2020,7 +2021,7 @@ def delete_volume_group(self, volume_group_id):
     def add_volume_to_volume_group(self, volume_group_id, volume_id):
         volume_name = self._get_volume_name_by_wwn(volume_id)
         cli_volume = self._get_cli_volume(volume_name)
-        if cli_volume.volume_group_id and cli_volume.volume_group_id != volume_group_id:
+        if cli_volume.volume_group_name and cli_volume.volume_group_name != volume_group_id:
```
matancarmeli7, line 2024: why do we do it?

controllers/servers/csi/volume_group_server.py
```diff
@@ -63,15 +63,15 @@ def DeleteVolumeGroup(self, request, _):
...
         array_connection_info = utils.get_array_connection_info_from_secrets(secrets)
 
         with get_agent(array_connection_info, array_type).get_mediator() as array_mediator:
             logger.debug(array_mediator)
 
             try:
-                logger.debug("Deleting volume group {}".format(volume_group_id))
-                array_mediator.delete_volume_group(volume_group_id)
+                logger.debug("Deleting volume group {}".format(volume_group_name))
+                array_mediator.delete_volume_group(volume_group_name)
```
matancarmeli7, line 74: deleting by id is not safer?

ArbelNathan: id is a running number - like "3", it is not unique enough, this way we can be sure we are deleting the right volume. if we had cross storage unique id for volume-group that could be the best.
if we delete by id we can accidentally delete the next volume group that catches the id (on retry, for example)

controllers/array_action/array_mediator_svc.py
ArbelNathan: checking if the volume is in vg..

matancarmeli7 (approved):
LGTM

ELENAGER (approved):
LGTM

PR #648, opened 2023-01-16
fix proto import problem

matancarmeli7:
Signed-off-by: matancarmeli7 <matan.carmeli7@gmail.com>


ELENAGER (approved):
Very clean and elegant solution.

PR #654, opened 2023-01-26
Add new decorator for replication lock

ELENAGER:


controllers/servers/csi/decorators.py
```diff
@@ -4,6 +4,7 @@
 from controllers.common.csi_logger import get_stdout_logger
```
matancarmeli7, line 4: do you know if we have unit tests for the decorators?

controllers/servers/csi/addons_server.py
```diff
@@ -8,15 +8,16 @@
 from controllers.array_action.storage_agent import get_agent
 from controllers.common.csi_logger import get_stdout_logger
 from controllers.servers import utils
-from controllers.servers.csi.decorators import csi_method
+from controllers.servers.csi.decorators import csi_method, csi_replication_method
```
ArbelNathan, line 11: are we using csi_method?

controllers/servers/csi/decorators.py
```diff
@@ -14,15 +15,38 @@ def csi_method(error_response_type, lock_request_attribute=''):
...
-            return handle_exception(ex, context, grpc.StatusCode.ABORTED, error_response_type)
-        logger.info("finished {}".format(controller_method_name))
-        return response
+        return _set_sync_lock(lock_id, lock_request_attribute, error_response_type,
+                              controller_method, servicer, request, context)
 
     return call_csi_method
+
+
+def csi_replication_method(error_response_type, lock_replication_request_attribute=''):
```
ArbelNathan, line 24: why do we need `lock_replication_request_attribute` if its always the same?

controllers/servers/csi/decorators.py
ELENAGER: Yes
    def test_disable_replication_already_processing(self):
        self._test_request_already_processing("replication_source", self.request.volume_id)

controllers/servers/csi/addons_server.py
ELENAGER: Not here, you are right. Removed

ArbelNathan (approved):
LGTM

PR #674, opened 2023-05-15
Task/csi 4868 remove default values from dataclass

matancarmeli7:


controllers/servers/host_definer/definition_manager/request.py
```diff
@@ -20,32 +20,30 @@ def generate_request(self, host_definition_info):
...
-        return request
+        prefix = utils.get_prefix()
+        connectivity_type_from_user = utils.get_connectivity_type_from_user(connectivity_type_label_on_node)
+        array_connection_info = self.secret_manager.get_array_connection_info(secret_name, secret_namespace, labels)
+
+        if array_connection_info:
+            return DefineHostRequest(
+                prefix=prefix, connectivity_type_from_user=connectivity_type_from_user,
+                array_connection_info=array_connection_info, node_id_from_host_definition='', node_id_from_csi_node='',
+                io_group='')
```
ArbelNathan, line 41: if its ok to values to be empty so default value is better..

PR #677, opened 2023-06-05
Feature/csi 5609 basic flow call home

matancarmeli7:
this is phase 1 of call-home without the option for the user to disable call-home


controllers/array_action/array_mediator_svc.py
```diff
@@ -193,6 +199,16 @@ def build_change_host_protocol_kwargs(host_name, protocol):
     }
 
 
+def build_register_plugin_kwargs(unique_key, metadata):
+    cli_kwargs = {
+        'uniquekey': unique_key,
```
ArbelNathan, line 204: I think 'uniquekey' should be const

controllers/servers/csi/decorators.py
```diff
@@ -50,3 +50,14 @@ def _set_sync_lock(lock_id, lock_request_attribute, error_response_type,
         return handle_exception(ex, context, grpc.StatusCode.ABORTED, error_response_type)
     logger.info("finished {}".format(controller_method_name))
     return response
+
+
+def register_csi_plugin(registration_map):
```
ArbelNathan, line 55: why not just import SVC_REGISTRATION_MAP?

controllers/servers/csi/csi_controller_server.py
```diff
@@ -99,6 +99,11 @@ def CreateVolume(self, request, context):
                         "volume was not found. creating a new volume with parameters: {0}".format(request.parameters))
 
                     array_mediator.validate_supported_space_efficiency(space_efficiency)
+                    if topologies:
+                        try:
+                            array_mediator.register_plugin('topology', '')
+                        except NotImplementedError:
+                            pass
```
ArbelNathan, line 106: if its not an actual error can we not raise `NotImplementedError` from those mediators?

controllers/array_action/array_mediator_svc.py
```diff
@@ -2016,17 +2047,61 @@ def get_volume_group(self, volume_group_id):
...
+
+    def _is_plugin_needs_to_be_registered(self, unique_key):
+        current_time = datetime.now()
+        endpoint_cache = SVC_REGISTRATION_CACHE.get(self.endpoint, '')
+        if not endpoint_cache:
+            return True
+        last_registration_time = endpoint_cache.get(unique_key, '')
+        if last_registration_time:
+            time_difference = current_time - last_registration_time
+            return time_difference >= timedelta(hours=2)
```
ArbelNathan, line 2081: const for the hours (2)

controllers/array_action/array_mediator_svc.py
```diff
@@ -2016,17 +2047,61 @@ def get_volume_group(self, volume_group_id):
...
             raise array_errors.VolumeAlreadyInVolumeGroup(volume_id, cli_volume.volume_group_name)
         self._change_volume_group(cli_volume.id, volume_group_id)
 
+    @register_csi_plugin(registration_map=SVC_REGISTRATION_MAP)
     def remove_volume_from_volume_group(self, volume_id):
         cli_volume = self._get_cli_volume_by_wwn(volume_id, not_exist_err=True)
         self._change_volume_group(cli_volume.id, None)
+
+    def register_plugin(self, unique_key,  metadata):
+        if self._is_plugin_needs_to_be_registered(unique_key):
```
ArbelNathan, line 2069:
```suggestion
        if self._is_plugin_needs_to_be_registered(unique_key) and self._is_registerplugin_supported():
```

controllers/array_action/array_mediator_svc.py
```diff
@@ -2016,17 +2047,61 @@ def get_volume_group(self, volume_group_id):
...
+            thread = Thread(target=self._register_plugin, args=(unique_key, metadata,))
+            thread.start()
```
ArbelNathan, line 2071: I believe its better to not use threads.

controllers/servers/csi/decorators.py
matancarmeli7: wanted that this decorator could serve non SVC functions in the future

ArbelNathan: I don't believe that we would have something similar in other storages, and if we do, we could use the same map - it's keyed on endpoints..

matancarmeli7: right, changed it

ArbelNathan (approved):
LGTM GJ

controllers/servers/csi/decorators.py
```diff
@@ -50,3 +53,14 @@ def _set_sync_lock(lock_id, lock_request_attribute, error_response_type,
         return handle_exception(ex, context, grpc.StatusCode.ABORTED, error_response_type)
     logger.info("finished {}".format(controller_method_name))
     return response
+
+
+def register_csi_plugin():
```
leonid-s-usov, line 58: have you considered adding the UNIQUE_KEY as an argument to this decorator?

leonid-s-usov:
Apparently, the grpc module is built with `concurrent.futures`, so the preferred async execution mode should be by using the [concurrent.futures.Executor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor). 

```
controller_server = grpc.server(futures.ThreadPoolExecutor(max_workers=max_workers))
```

It's possible that you will need to pass the executor separately to the controller servicers as you need an executor instance to submit a callable

controllers/servers/csi/decorators.py
matancarmeli7: no, because it gets the unique key from extracting it from the registration_map using the function name as a key

leonid-s-usov: I know, so now you have to keep the registration array separate from the function definition. Passing the unique key as the decorator argument could keep the code in one place and obvious to a reader who doesn't know about the `REGISTRATION_MAP`.

It would be

```   
    @register_csi_plugin('volume_group')
    def remove_volume_from_volume_group(self, volume_id):
```

and no need to keep the separate file of `registration_maps.py`, as you'll capture the values inside the decorator. The below wasn't tested, but something along the lines:

```
def register_csi_plugin(unique_key, metadata=''):
    def wrapper():
          @decorator
          def call_csi_plugin_registration(mediator_method, mediator_class, *args):
              mediator_class.register_plugin(unique_key, metadata)
              return mediator_method(mediator_class, *args)
      
          return call_csi_plugin_registration
    return wrapper
```

matancarmeli7: I see some advantages and disadvantages in this approach as the approach that I took, I don't think that it will change much, so I can leave with both ways

leonid-s-usov: what are the disadvantages?

matancarmeli7: I think it can require more maintenance when you want to change a unique key because you will change it in multiple places, and I believe using registration maps will let us see which functions belong to which unique keys

leonid-s-usov: If someone decides to rename a handler method they will have to know that they need to go and also rename the key in the registration map, which is not obvious, and that's my main concern. When you have the unique id configured in the decorator to a handler, that's the only place you need to change it if you want to, and it's obvious which handler also reports a particular plugin usage.

If you're worried about multiple different handlers using the same unique id, you can surely define constants and use those instead of a literal, like

```
VOLUME_GROUP_FEATURE='volume_group'
...

 @register_csi_plugin(VOLUME_GROUP_FEATURE)
    def remove_volume_from_volume_group(self, volume_id):
...

 @register_csi_plugin(VOLUME_GROUP_FEATURE)
    def another_handler(self, volume_id):
```

PR #683, opened 2023-07-06
Improvement/csi 4547 clone idempotency

ArbelNathan:


controllers/array_action/array_mediator_svc.py
```diff
@@ -951,12 +962,23 @@ def _rmsnapshot(self, internal_snapshot_id):
                 raise array_errors.ObjectNotFoundError(internal_snapshot_id)
             raise ex
 
+    def _handle_delete_snapshot(self, snapshot_name, internal_snapshot_id):
+        if self._is_addsnapshot_supported():
+            if snapshot_name:
+                cli_snapshot = self._get_cli_snapshot_by_name(snapshot_name)
+                if cli_snapshot and cli_snapshot.snapshot_name == snapshot_name:
```
leonid-s-usov, line 970: is there a chance `_get_cli_snapshot_by_name` will return a snapshot with a different name?

controllers/servers/csi/exception_handler.py
```diff
@@ -16,7 +16,8 @@
     array_errors.HostNotFoundError: grpc.StatusCode.NOT_FOUND,
     array_errors.PermissionDeniedError: grpc.StatusCode.PERMISSION_DENIED,
     array_errors.ObjectIsStillInUseError: grpc.StatusCode.FAILED_PRECONDITION,
-    array_errors.CredentialsError: grpc.StatusCode.UNAUTHENTICATED
+    array_errors.CredentialsError: grpc.StatusCode.UNAUTHENTICATED,
+    array_errors.ObjectAlreadyExistError: grpc.StatusCode.ALREADY_EXISTS,
```
leonid-s-usov, line 20: Naming suggestion: ObjectAlreadyExists sounds like the same object is there and we should just ignore it in case it was an idempotent create request. The way it is used rather looks like ObjectParameterConflictError, meaning that we are trying to create an incompatible object with the same name.

leonid-s-usov (approved):
Looks good, no critical comments

controllers/servers/csi/exception_handler.py
ArbelNathan: it's following the Spec naming, this is not visible by the user, the user will get the grpc status and the [description](https://github.com/IBM/ibm-block-csi-driver/blob/159a88d7a5ee0225cc632f9280783656032cc4f4/controllers/array_action/messages.py#L89)

leonid-s-usov: ack

matancarmeli7 (approved):
LGTM

controllers/array_action/array_mediator_svc.py
ArbelNathan: right, I cleaned the code

matancarmeli7 (approved):
LGTM

leonid-s-usov (approved):

PR #756, opened 2025-01-09
When volume is not mapped to a host, allow the unmap to fail

shlomitn:


csiblock (approved):

shlomitn:
First, I tested it manually, as I described in the ticket, and it's possible to see in the controller's logs the message I have added.
Second, after your comment I added UT similar to other errors' UTs. I don't know how much they're worth, but I wasn't able to run the UT at all yet (not talking about mine - just failing to run UT). So didn't push that yet.
(To document:
Running scripts/run_unitests.sh (after I changed "docker" to "podman" in scripts/run_unitests.sh, and after adding chmod+x ./csi_pb2.sh in Dockerfile-controllers.test), resulted in:
+ podman run --entrypoint ./controllers/scripts/unitests.sh --rm -t csi-controller-tests
Error: preparing container ea937cc5f10e479d87fbcf82a17496c1321852614eb639112020aca65e4f76b8 for attach: crun: open executable: Permission denied: OCI permission denied
)

csiblock:
after speaking over Slack, UTs should be run as sudo

@shlomitn where can i find the UTs or have they not been pushed yet? your attention that i thought this PR was already tested, so i merged it...

shlomitn:
I'll just say that it _was_ tested - just not with UT, but in reality on a real cluster+storage.

PR #757, opened 2025-01-12
CSI-5774: UT for new error code

shlomitn:


Dockerfile-controllers.test
```diff
@@ -30,6 +30,7 @@ RUN pip3 install -r ./csi_requirements.txt -r ./host_definer_requirements.txt
 
 USER root
 COPY controllers/scripts/csi_general .
+RUN chmod +x csi_pb2.sh
```
csiblock, line 33: how did it work until now without this?

shlomitn: I don't know. I mean, I don't know why it's not working now without it, but it's not the first time I've had to make such a fix. Seems like before our time, the copies copied the files with their original mods, so if a file had been executable, then after the copy it still was executable, but now it's not. I don't know if it's the OS or what it is, but that's the situation.

kasserater (approved):

PR #827, opened 2025-11-30
Uriziv1/csi 5997

uriziv1:
This PR is partial.
It contains only the Node side, and kind of start of the HostDefiner side. Controller side still not started.


uriziv1:
closing this PR as it should be merged to 1.13.0

PR #846, opened 2026-02-09
Sumit/csi 5841

sumitguptaibm:
In case of nvme, we don't need to run "multipath -f", we just need to escape it.
Once volume is unstaged, nvme devices are themselves removed from the multipath devices list.
For checking whether a device is nvme or not, I am checking existence of /sys/block/devicename/device/subsysnqn
(`nvme disconnect` would have removed all multipath nvme devices, so that was not exercised).


node/pkg/driver/device_connectivity/device_connectivity_helper_scsigeneric.go
shlomitn, line 1: I see the logic was: Run multipathd, if it fails (returns with error) - exit immediately with the error. Otherwise: check if the output contains contains the volumeIdVariation. If it does - exit. If not - sleep for intervalSeconds and retry - for maxRetries.
The logic after your change is: run multipathd. If it fails - sleep and retry. Otherwise - same as above, return if found the volumeIdVariation or else retry.
That's a change of the logic. Why?... Who says we want to retry in case of multipathd error?

node/pkg/driver/node.go
```diff
@@ -388,13 +389,22 @@ func (d *NodeService) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstag
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	err = d.OsDeviceConnectivityHelper.FlushMultipathDevice(baseDevice)
+	isNvme, err := d.NodeUtils.IsNvmeBaseDevice(baseDevice)
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, "Multipath -f command failed with error: %v", err)
+		logger.Warningf("Failed to check if device %s is NVMe, assuming non-NVMe: %v", baseDevice, err)
```
shlomitn: this is... a problem. But I'm not sure there's a better way.
What would happen here? Either we're not NVMe, and then it's ok, or we may be NVMe and still try multipath -f, and it will fail, and so we won't clean the device.
What does actually clean the device eventually for NVMe? The function ClearStageInfoFile? Or what?...

PR #852, opened 2026-03-09
Add Logic for NVMe oFC for Discovery + Login

parthraaj:


controllers/array_action/array_mediator_ds8k.py
```diff
@@ -726,6 +726,9 @@ def get_host_io_group(self, host_name):
 
     def change_host_protocol(self, host_name, protocol):
         raise NotImplementedError
+    
+    def get_nvme_target_ports(self, host_name):
```
komal-56, line 730: These stubs are only here to satisfy the @abstractmethod contract in 
array_mediator_interface.py — DS8K and XIV don't actually support NVMe-oFC.

The actual working implementation of get_nvme_target_ports() is only in 
array_mediator_svc.py (FlashSystem/SVC arrays) which is correct — SVC 
is the only array that supports NVMe-oFC here.

Suggestion: Instead of marking get_nvme_target_ports() as @abstractmethod 
in the interface, put a default implementation in the abstract base class 
(array_mediator_abstract.py):

    def get_nvme_target_ports(self):
        raise NotImplementedError("NVMe over FC is not supported for this array type")

Benefits:
- DS8K and XIV don't need these stubs — they just inherit the default
- SVC keeps its real implementation as-is, no changes needed
- Any future array added will also get this default automatically
- Same behaviour, less code to write and maintain

PR #891, opened 2026-08-18
Komal/csi 5846

komal-56:


controllers/servers/csi/csi_controller_server.py
```diff
@@ -355,12 +355,13 @@ def CreateSnapshot(self, request, context):
...
                                                                 system_id=array_connection_info.system_id)
-            use_snap_object = bool(snapshot_parameters.virt_snap_func and not array_connection_info.partition_name)
+            virt_snap_func = utils.get_virt_snap_func_from_storage_class(source_id)
+            use_snap_object = bool(virt_snap_func and not array_connection_info.partition_name)
             pool = snapshot_parameters.pool
             space_efficiency = snapshot_parameters.space_efficiency
             # Apparently this was never supposed to be a restriction - meanwhile enabling for partitions only for
             # backward compatibility, need to revisit
-            if snapshot_parameters.virt_snap_func and not array_connection_info.partition_name \
+            if virt_snap_func and not array_connection_info.partition_name \
```
shlomitn: this line can be replaced with if use_snap_object. But not a must, of course.

docs/content/configuration/creating_storageclass_vg.md
shlomitn, line 1: Very nice that you changed the documentation too!
I think, though, that we shouldn't use all this *only* in storageclass. Imagine someone coming to read the documentation. They may not know that we used to put this value in both places. So I think it should just say, for example in this file here: "The `virt_snap_func` parameter is configured in the StorageClass. When creating a snapshot, the driver automatically retrieves the value from the source volume's StorageClass" and that's it. And so on for all other places.
Maybe in the 'whats_new" file we can add that we removed it from the volumesnapshotclass.

shlomitn (commented):
Nice!

controllers/servers/csi/csi_controller_server.py
komal-56: Done! Simplified the condition to `if use_snap_object and space_efficiency:` since `use_snap_object` already incorporates the `virt_snap_func` check.

docs/content/configuration/creating_storageclass_vg.md
komal-56: Done! Simplified all documentation to not mention old behavior. Added breaking change notice in whats_new.md.

docs/content/configuration/creating_storageclass_vg.md
```diff
@@ -9,7 +9,7 @@ Create a storage class YAML file, similar to the following demo-storageclass.yam
 
 Volume groups can only be managed by **either** the associated VolumeGroup **or** the associated StorageClass (with the `volume_group` parameter). If a volume group is already associated with a VolumeGroup, then each volume of this StorageClass can be automatically deleted.{: restriction}
 
-When electing to set the optional "virt_snap_func" parameter, it **must** also be set with an identical value in the relevant VolumeSnapshotClass yamls.{: requirement}
+The `virt_snap_func` parameter is configured in the StorageClass. When creating a snapshot, the driver automatically retrieves the value from the source volume's StorageClass.{: tip}
```
shlomitn: I'm for removing this line altogether. It was there only to say that if you set it here, then you must also set it in volumesnapshotclass. So it's just not needed anymore. Let's drop it.

docs/content/configuration/creating_volumesnapshotclass.md
```diff
@@ -17,11 +17,9 @@ For IBM DS8000® family storage systems, the maximum prefix length is five chara
 
 For IBM Storage Virtualize® family storage systems, the `CSI` prefix is added as default if not specified by the user.{: tip}
 
-- The `virt_snap_func` parameter is optional but necessary in IBM Storage Virtualize® storage systems if using the Snapshot function. To enable the Snapshot function, set the value to _"true"_. The default value is _"false"_. If the value is `"false"` the snapshot will use the FlashCopy function.
+- For IBM Storage Virtualize® storage systems, the `virt_snap_func` parameter is configured in the StorageClass to control whether snapshots use the Snapshot function (`"true"`) or FlashCopy function (`"false"`, default). The driver automatically retrieves this value from the source volume's StorageClass. See [Creating a StorageClass](creating_storageclass.md) for details.{: tip}
```
shlomitn: Let's keep it just as it was. Is there a reason not to?

docs/content/configuration/creating_volumestorageclass.md
```diff
@@ -50,7 +50,7 @@ Volume groups can only be managed by **either** the associated VolumeGroup **or*
 
 - The `virt_snap_func` parameter is optional but necessary in IBM Storage Virtualize® storage systems if using the Snapshot function. To enable the Snapshot function, set the value to _"true"_. The default value is _"false"_. If the value is _"false"_ the snapshot will use the FlashCopy function.
 
-When electing to set the optional "virt_snap_func" parameter, it **must** also be set with an identical value in the relevant VolumeSnapshotClass yamls.{: requirement}
+The `virt_snap_func` parameter is configured in the StorageClass. When creating a snapshot, the driver automatically retrieves the value from the source volume's StorageClass.{: tip}
```
shlomitn: let's just remove this line

docs/content/release_notes/whats_new.md
```diff
@@ -7,6 +7,16 @@ IBM® block storage CSI driver 1.14.0 adds support for:
...
 
+## Breaking changes
+
+### virt_snap_func parameter configuration change
+
+The `virt_snap_func` parameter is now **only** read from the StorageClass and is **no longer** supported in VolumeSnapshotClass configurations. When creating a snapshot, the driver automatically retrieves the `virt_snap_func` value from the source volume's StorageClass.
+
+**Action required:** Remove any `virt_snap_func` parameter from your VolumeSnapshotClass YAML files. The parameter should only be configured in StorageClass YAML files. Any value specified in VolumeSnapshotClass will be ignored.
+
+For more information, see [Creating a StorageClass](../configuration/creating_volumestorageclass.md) and [Creating a VolumeSnapshotClass](../configuration/creating_volumesnapshotclass.md).
```
shlomitn: Ok, so the end of this line is... they won't see anything in volumesnapshotclass, will they? I think it can just be the first part of the sentence: "For more information, see [Creating a StorageClass](../configuration/creating_volumestorageclass.md)"

Also, above, I feel like this "action required" is a bit scary (or am I scared easily? :-) ) maybe we can keep the first line ("The `virt_snap_func` parameter is now **only**....") and add at the end "Any value specified in VolumeSnapshotClass will be ignored"

shlomitn (dismissed):

PR #893, opened 2026-08-24
fix(test): fix flaky test_updated_node_id_of_csi_node in full test suite

komal-56:
use MODIFIED_EVENT_TYPE instead of DELETED (wrong branch was entered)
replace side_effect=[list] with return_value to avoid stale mock state
use FAKE_DAEMON_SET_NAME instead of FAKE_NODE_PODS_NAME in daemon set manifest
set different FC initiator on stored HostDefinition to trigger define_host call

### DIFF
--- a/controllers/tests/controller_server/host_definer/csi_node_watcher_test.py
+++ b/controllers/tests/controller_server/host_definer/csi_node_watcher_test.py
@@ -95,21 +95,21 @@ class TestWatchCsiNodesResources(CsiNodeWatcherBase):
         self._prepare_mocks_for_updated_csi_node()
         host_definitions = self.ready_k8s_host_definitions
         host_definitions.items[0].spec.hostDefinition.nodeId = 'other_node_id'
+        host_definitions.items[0].metadata.annotations[common_settings.NODE_INITIATORS_FIELD] = '{"nvmeofc":[],"fc":["old_wwpn"],"iscsi":[]}'
         self.csi_node_watcher.host_definitions_api.get.return_value = self.ready_k8s_host_definitions
         self.csi_node_watcher.core_api.read_namespaced_secret.return_value = test_utils.get_fake_k8s_secret()
         test_utils.run_function_with_timeout(self.csi_node_watcher.watch_csi_nodes_resources, 0.5)
         self.assertEqual(1, len(self.nodes_on_csi_node_watcher))
-        # self.csi_node_watcher.storage_host_servicer.define_host.assert_called()
+        self.csi_node_watcher.storage_host_servicer.define_host.assert_called()
 
     def _prepare_mocks_for_updated_csi_node(self):
         self.nodes_on_watcher_helper[test_settings.FAKE_NODE_NAME] = test_utils.get_fake_managed_node()
         self.nodes_on_csi_node_watcher[test_settings.FAKE_NODE_NAME] = test_utils.get_fake_managed_node()
         self.managed_secrets_on_csi_node_watcher.append(test_utils.get_fake_secret_info())
         self.csi_node_watcher.csi_nodes_api.watch.return_value = iter(
-            [test_utils.get_fake_csi_node_watch_event(test_settings.DELETED_EVENT_TYPE)])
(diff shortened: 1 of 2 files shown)


controllers/tests/controller_server/host_definer/csi_node_watcher_test.py
```diff
@@ -95,21 +95,21 @@ def test_updated_node_id_of_csi_node(self):
...
-        # self.csi_node_watcher.storage_host_servicer.define_host.assert_called()
+        self.csi_node_watcher.storage_host_servicer.define_host.assert_called()
 
     def _prepare_mocks_for_updated_csi_node(self):
         self.nodes_on_watcher_helper[test_settings.FAKE_NODE_NAME] = test_utils.get_fake_managed_node()
         self.nodes_on_csi_node_watcher[test_settings.FAKE_NODE_NAME] = test_utils.get_fake_managed_node()
         self.managed_secrets_on_csi_node_watcher.append(test_utils.get_fake_secret_info())
         self.csi_node_watcher.csi_nodes_api.watch.return_value = iter(
-            [test_utils.get_fake_csi_node_watch_event(test_settings.DELETED_EVENT_TYPE)])
+            [test_utils.get_fake_csi_node_watch_event(test_settings.MODIFIED_EVENT_TYPE)])
```
shlomitn, line 110: Interesting! Why was it ever like this?? Why did it even work sometimes?
But it definitely looks more correct like this - since the function is called ..for_updated_csi_node...

shlomitn (approved):
Did you run all UT together and saw that they passed?

controllers/tests/controller_server/host_definer/csi_node_watcher_test.py
SushilSanjayBhile: MODIFIED looks like correct event — Kubernetes emits it (not DELETED) when a CSI node changes during a DaemonSet rollout, and the old DELETED path had a name in NODES guard that silently failed under stale mock state in the full suite. @komal-56, can you confirm this matches your debugging findings? Or Do you have any other reason as well for this modification?

komal-56:
> Did you run all UT together and saw that they passed?

Yes, ran the full test suite using ./scripts/run_unitests.sh and all tests passed.

controllers/tests/controller_server/host_definer/csi_node_watcher_test.py
komal-56: > Interesting! Why was it ever like this?? Why did it even work sometimes? But it definitely looks more correct like this - since the function is called ..for_updated_csi_node...

The assertion was commented out before this fix, so the test had nothing to fail on and always passed silently regardless of what happened internally. That's why no one noticed it was broken.

When I uncommented the assertion and ran the full test class, it failed because test_updated_csi_node_not_removed runs before this test and both share _prepare_mocks_for_updated_csi_node(). The side_effect=[list] gets consumed by the first test, leaving not_updated_daemon_set on every call for the second test — the while-loop never exits within the 0.5s timeout so define_host is never reached.

And yes, MODIFIED is the correct event — the function name _prepare_mocks_for_updated_csi_node makes that clear.

komal-56: > MODIFIED looks like correct event — Kubernetes emits it (not DELETED) when a CSI node changes during a DaemonSet rollout, and the old DELETED path had a name in NODES guard that silently failed under stale mock state in the full suite. @komal-56, can you confirm this matches your debugging findings? Or Do you have any other reason as well for this modification?

Yes, exactly. The failing logs from the ticket confirmed this — `In if DELETED_EVENT` showed that the code was entering the wrong branch, and `updated [0] desired [1]` showed that the while-loop was stuck. Together, these prevented the code from reaching `define_host`.

`MODIFIED` is the correct event for this test, as it is triggered when the CSI node changes during a DaemonSet rollout. `DELETED` represents a different scenario (node removal) and already has separate handling in the code.

There isn't any other reason for this modification; this change is based on the actual event flow observed during debugging.

SushilSanjayBhile:
LGTM

shlomitn:
Well then, merge! :)

